PR #2, opened 2021-09-01
upgrade hdwallet - NOTE: probably needs some more changes in the chain-adapters

mvpratt:
See PR #439 in the legacy chain-adapters repo for additional changes that we need to port over.


0xdef1cafe:
no longer valid

PR #9, opened 2021-09-02
Asset service part2

elmutt:
- service providing info from generated asset file. Has one function to get all assets for now but can be expanded upon as needed

- Tries to get generated asset file data from github repo first but falls back to local file if theres a problem.

- Moved around some existing logic to make a clear distinction between the asset service and code generating the data file

- Added a testnet token to demonstrate how we can distinguish between mainnet and testnets in our asset support


packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -0,0 +1,35 @@
+import fs from 'fs'
+import { baseAssets } from './baseAssets'
+import { getTokens } from './ethTokens'
+import { BaseAsset, NetworkTypes } from '../types'
+import { AssetService } from '../'
+
+const generateAssetData = async () => {
+  const generatedAssetData = await Promise.all(
```
majorhayes, line 8: Are we okay with this failing if any of the promises fail?

We may want to do a `Promise.allSettled` and then filter the results to the successful results.

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -0,0 +1,35 @@
...
+    `./src/service/generatedAssetData.json`,
+    JSON.stringify(generatedAssetData)
+  )
+
+  const assetService = new AssetService('blah')
+
+  await assetService.initialize()
+  const assets = assetService.byNetwork(NetworkTypes.ETH_ROPSTEN)
+
+  console.log('assets is', assets)
```
majorhayes, line 30: There is no error handling. `try/catch` needed around the various places things could fail.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Get asset data from assetFileUrl and flatten it for easier use
+   */
+  async initialize() {
+    try {
+      const { data } = await axios.get<BaseAsset[]>(this.assetFileUrl)
+      this.assetData = data
+    } catch (err) {
+      this.assetData = localAssetData as BaseAsset[]
```
majorhayes, line 27: Where is `localAssetData` defined?

packages/asset-service/src/service/AssetService.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (baseAsset.tokens) {
+        for (const tokenAsset of baseAsset.tokens) {
+          flatAssetData.push({
+            ...tokenAsset,
+            chain: baseAsset.chain,
+            network: baseAsset.network
+          })
+        }
+      }
+    }
```
majorhayes, line 45: This looks a lot like a `.flatMap`, which may be a simpler implementation.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -0,0 +1,60 @@
...
+  private assetData: BaseAsset[]
+  private flatAssetData: Asset[]
+
+  constructor(assetFileUrl: string) {
+    this.assetFileUrl = assetFileUrl
+  }
+
+  private checkInitialized() {
+    if (!this.assetData || !this.flatAssetData) throw new Error('Asset service not initialized')
+  }
```
majorhayes, line 17:
```suggestion
  get isInitialized(): boolean {
    return !(Array.isArray(this.assetData) && Array.isArray(this.flatAssetData))
  }

  private checkInitialized() {
    if (!this.isInitialized) throw new Error('Asset service not initialized')
  }
```

majorhayes (approved):

packages/asset-service/src/generateAssetData/generateAssetData.ts
cjthompson, line 8: Are we okay with this failing if any of the promises fail?

We may want to do a Promise.allSettled and then filter the results to the successful results.

packages/asset-service/src/generateAssetData/generateAssetData.ts
cjthompson, line 30: There is no error handling. try/catch needed around the various places things could fail.

packages/asset-service/src/service/AssetService.ts
cjthompson, line 17:
```suggestion
  get isInitialized(): boolean {
    return !(Array.isArray(this.assetData) && Array.isArray(this.flatAssetData))
  }

  private checkInitialized() {
    if (!this.isInitialized) throw new Error('Asset service not initialized')
  }
```

packages/asset-service/src/service/AssetService.ts
cjthompson, line 45: This looks a lot like a `.flatMap`.

packages/asset-service/src/generateAssetData/generateAssetData.ts
elmutt: @0xdef1cafe brought this up before on a previous PR fbefore I moved the file in this PR

In my opinion its not that beneficial to much error handling logic as this section is intended to be used as a locally run script  to generate the assets file and is not used dynamically in any other code. just a local script to be run by engineers before committing

packages/asset-service/src/generateAssetData/generateAssetData.ts
elmutt: In my opinion its not that beneficial to much error handling logic as this section is intended to be used as a locally run script to generate the assets file and is not used dynamically in any other code. just a local script to be run by engineers before committing

packages/asset-service/src/service/AssetService.ts
elmutt: As far as I can tell flatmap doesn't seem to apply to this situation.

The initial data structure is like this:

{
chain: 'eth',
symbol: 'eth'
slip44: 60,
etc...
tokens: [ { symbol: fox, etc...} ]
}

The idea is to flatten it so the tokens are no longer nested under eth but are a flat list of all the assets (base assets and tokens) instead of having the tokens nested under the base asset.

natven (approved):

unclerom (approved):

SterL (approved):

PR #25, opened 2021-09-08
add description function to asset service

elmutt:
add description function to asset service. make assetFileUrl optional


packages/asset-service/src/service/AssetService.ts
```diff
@@ -8,8 +8,8 @@ export class AssetService {
   private assetData: BaseAsset[]
   private flatAssetData: Asset[]
 
-  constructor(assetFileUrl: string) {
-    this.assetFileUrl = assetFileUrl
+  constructor(assetFileUrl?: string) {
+    this.assetFileUrl = assetFileUrl || ''
```
cjthompson, line 12: What happens if `assetFileUrl` is empty?

My preference would be to change the private variable type to `string | undefined` rather than filling it with an empty string. Then you can do a `if (typeof this.assetFileUrl === 'string)` check to see if it's been set.

cjthompson: `assetFileUrl` isn't used in the new `description` function and you didn't change anywhere that uses this, so I'm unsure why this change is here.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -64,4 +64,13 @@ export class AssetService {
       ? this.flatAssetData.filter((asset) => asset.network == network)
       : this.flatAssetData
   }
+
+  async description(name: string, tokenId?: string): Promise<string> {
+    const isToken = !!tokenId
```
cjthompson, line 69: if you have `lodash` as a dependency, just do `lodash.isString()`

packages/asset-service/src/service/AssetService.ts
```diff
@@ -64,4 +64,13 @@ export class AssetService {
...
+    const contractUrl = isToken ? `contract/${tokenId?.toLowerCase()}` : ''
+    const { data } = await axios.get(
+      `https://api.coingecko.com/api/v3/coins/${name.toLowerCase()}/${contractUrl}`
```
cjthompson, line 72: does this request work with a trailing `/` and no contract URL?

packages/asset-service/src/service/AssetService.ts
```diff
@@ -64,4 +64,13 @@ export class AssetService {
...
+    )
+    return data.description.en
```
cjthompson, line 74: This needs to be wrapped in a try/catch and return something like null if there's no description.

change the return type to `Promise<string | null>`, add a `try/catch` with `return null`.

You also aren't verifying that `data` actually contains a `description` object. It's a third party, we should never assume what data is coming back.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -64,4 +64,13 @@ export class AssetService {
...
+  }
```
cjthompson, line 75:
```suggestion
  async description(name: string, tokenId?: string): Promise<string | null> {
    if (typeof name !== 'string') throw new Error('Invalid asset name')
    const contractUrl = typeof tokenId === 'string' ? `/contract/${tokenId?.toLowerCase()}` : ''
    try {
      const { data } = await axios.get(
        `https://api.coingecko.com/api/v3/coins/${name.toLowerCase()}${contractUrl}`
      )
      return data?.description?.en || null
    } catch (e) {
      console.error('AssetService:description:error', e)
      return null
    }
  }
```

packages/asset-service/src/service/AssetService.ts
elmutt: I wanted to make it optional. I agree with you on letting it be undefined rather than empty string

millsmcilroy (approved):

packages/asset-service/src/service/AssetService.ts
elmutt: ya it works

packages/asset-service/src/service/AssetService.ts
elmutt: i pushed up a fix. lmk what you think

BitHighlander (approved):

packages/asset-service/src/service/AssetService.ts
```diff
@@ -25,6 +25,7 @@ export class AssetService {
    */
   async initialize() {
     try {
+      if (!this.assetFileUrl) throw new Error()
```
cjthompson, line 28: An error message would be nice, like `'Parameter AssetFileUrl is required'`

cjthompson (approved):

unclerom (approved):

PR #29, opened 2021-09-10
Seperate asset service from market service (final)

elmutt:
Final changes needed for separate use of asset service and market service.

Type changes, name changes, remove stuff from market service and new asset lookup function


DaoDev44 (approved):

packages/market-service/package.json
```diff
@@ -20,6 +20,7 @@
     "test": "jest test"
   },
   "dependencies": {
+    "@shapeshiftoss/asset-service": "^1.0.0",
```
0xdef1cafe, line 23: uhhh we can't be doing this or we end up with coupled packages

0xdef1cafe: @cjthompson ?

0xdef1cafe (changes requested):

theoboldfrazier (approved):
LGTM

packages/market-service/package.json
cjthompson: As long as there isn't a circular dependency it's fine.

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -20,6 +20,10 @@ const generateAssetData = async () => {
     `./src/service/generatedAssetData.json`,
     JSON.stringify(generatedAssetData)
   )
+
+  const assetService = new AssetService()
+
+  await assetService.initialize()
```
cjthompson, line 26: I don't see `assetService` being called anywhere else.

packages/market-service/package.json
cjthompson: However, it does look to me like it doesn't actually do anything with asset-service.

Also, we should be using our DI model which would take an AssetService in the constructor on not have a direct dependency on the package.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -1,19 +1,11 @@
 import axios from 'axios'
 import dayjs from 'dayjs'
-import { MarketService, AssetMarketData, HistoryData, HistoryTimeframe } from '../api'
+import { MarketService, MarketData, HistoryData, HistoryTimeframe } from '../api'
+import { ChainTypes } from '@shapeshiftoss/asset-service'
```
cjthompson, line 4: I think we should move shared types out into a `@shapeshiftoss/types` package

packages/asset-service/src/generateAssetData/generateAssetData.ts
elmutt: good call. was using this for testing. i can remove it

packages/market-service/src/coingecko/coingecko.ts
elmutt: Agreed. I want to do that in another PR so I can get this merged before people continue using the old market service for everything and make more changes

0xdef1cafe (approved):

PR #33, opened 2021-09-11
remove commitlint github action

### LINKED ISSUE #34: Remove commitlint
This was way more agro than I anticipated. I'd hoped it would only lint the merge commit message to `main` and `develop`, not every commit on a branch going into them.

It doesn't seem configurable to allow this behavior.

For discussion

pros

- extremely tidy commit messages
- enables automatic releases conforming to semantic versioning based on types of commits
- enables automatically generated changelogs
- can ignore merge commits, so we can merge develop into a PR to ensure it's up to date

cons

- requires all commits on a PR branch to confirm to the spec, or branches to be rebased/squashed to a single commit before PRing (not necessarily a bad thing)
- adds additional friction for new contributors
- may not be able to use github's super handy commit suggestion feature (TBD)

0xdef1cafe:
ci: commitlint was more aggressive than anticipated and is too onerous until we have agreement on it's usage and benefits.

closes #34


.github/workflows/develop.yml
```diff
@@ -7,15 +8,15 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Commit lint
-        uses: wagoid/commitlint-github-action@v4
+      # - name: Commit lint
+      #   uses: wagoid/commitlint-github-action@v4
```
cjthompson, line 19: Why not just delete it? Isn't that what git history is for?

cjthompson (approved):

.github/workflows/develop.yml
0xdef1cafe: done

elmutt (approved):

PR #38, opened 2021-09-13
update from network to chain

### LINKED ISSUE #37: Update lib getMarketData, getPriceHistory to handle asset chain

reallybeard:
Closes #37

### DIFF
--- a/packages/market-service/src/coingecko/coingecko.ts
+++ b/packages/market-service/src/coingecko/coingecko.ts
@@ -18,12 +18,11 @@ type CoinGeckoAssetData = {
   }
 }
 
-const getId = (chain: string) => {
-  if (chain === ChainTypes.ETH) return 'ethereum'
-  if (chain === ChainTypes.BTC) return 'bitcoin'
-  if (chain === ChainTypes.LTC) return 'litecoin'
-  throw new Error('Unsuppored chain type')
-}
+const coinGeckoIds = Object.freeze({
+  [ChainTypes.ETH]: 'ethereum',
+  [ChainTypes.BTC]: 'bitcoin',
+  [ChainTypes.LTC]: 'litecoin'
+})
 
 export class CoinGeckoMarketService implements MarketService {
   baseUrl = 'https://api.coingecko.com/api/v3'
--- a/packages/market-service/src/coingecko/coingecko.ts
+++ b/packages/market-service/src/coingecko/coingecko.ts
@@ -18,20 +18,25 @@ type CoinGeckoAssetData = {
   }
 }
 
+const getId = (chain: string) => {
+  if (chain === ChainTypes.ETH) return 'ethereum'
+  if (chain === ChainTypes.BTC) return 'bitcoin'
+  if (chain === ChainTypes.LTC) return 'litecoin'
+  throw new Error('Unsuppored chain type')
+}
+
 export class CoinGeckoMarketService implements MarketService {
   baseUrl = 'https://api.coingecko.com/api/v3'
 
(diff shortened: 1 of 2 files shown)


packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -18,20 +18,25 @@ type CoinGeckoAssetData = {
   }
 }
 
+const getId = (chain: string) => {
+  if (chain === ChainTypes.ETH) return 'ethereum'
```
technojak, line 22: We should have an enum for these hard coded values as well.

Maybe we have a `ChainTypesShort` and `ChainTypesLong` and read the values from those.

technojak (commented):
lgtm but I think we should use enums

bootsboogie (approved):

packages/market-service/src/coingecko/coingecko.ts
elmutt: The ChainTypesLong would be specific to coingecko. So maybe not needed?

elmutt: If we do add ChainTypesLong it should be specific to the coingecko service. Maybe call it coingeckAssetId or something since its only used in coingecko urls

technojak: yeah I agree with that. we can make an enum called `CoinGeckoIds` and define those

cjthompson: I'd really prefer not to have a bunch of information about chains spread around in different enums, etc. There needs to be single source of truth.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -18,20 +18,25 @@ type CoinGeckoAssetData = {
...
+  if (chain === ChainTypes.BTC) return 'bitcoin'
+  if (chain === ChainTypes.LTC) return 'litecoin'
+  throw new Error('Unsuppored chain type')
+}
```
cjthompson, line 26:
```suggestion
const coinGeckoIds = Object.freeze({
  [ChainTypes.ETH]: 'ethereum',
  [ChainTypes.BTC]: 'bitcoin',
  [ChainTypes.LTC]: 'litecoin'
})
```

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -18,20 +18,25 @@ type CoinGeckoAssetData = {
...
+
 export class CoinGeckoMarketService implements MarketService {
   baseUrl = 'https://api.coingecko.com/api/v3'
 
   getMarketData = async (chain: ChainTypes, tokenId?: string): Promise<MarketData | null> => {
-    let coingecko_id
-    if (chain === ChainTypes.ETH) coingecko_id = 'ethereum'
-    else throw new Error('Unsuppored chain type')
+    const id = getId(chain)
```
cjthompson, line 32:
```suggestion
    const id = coinGeckoIds[chain]
    
    if (id) {
```

cjthompson: then lines 51-54

```ts
      } catch (e) {
        console.warn(e)
      }
    }

    return null
  }
```

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -77,14 +82,15 @@ export class CoinGeckoMarketService implements MarketService {
     }
 
     try {
+      const id = getId(chain)
```
cjthompson, line 85:
```suggestion
    const id = coinGeckoChains[chain]
    if (id) {
      try {
```

PR #72, opened 2021-09-22
Import Zrx get quote

toshiSat:
I had to add some linter rules because catch blocks were forced to be unknown and it needed to be implicit any.
We will need to grab these types from the common package when it's created


packages/swapper/src/swappers/zrx/ZrxSwapper.test.ts
```diff
@@ -1,5 +1,178 @@
...
+    it('returns quote without fee data', async () => {
+      const { quoteInput } = setupQuote()
+      const swapper = new ZrxSwapper()
+      ;(zrxService.get as jest.Mock<unknown>).mockReturnValue(
+        Promise.resolve({
+          data: { success: true, price: '100', gasPrice: '1000' }
+        })
+      )
+      const quote = await swapper.getQuote(quoteInput)
+      console.log('quote', quote)
```
ryankeepkey, line 81: can remove this

DaoDev44, line 81: can remove this `console.log`

packages/swapper/src/swappers/zrx/ZrxSwapper.ts
```diff
@@ -1,7 +1,145 @@
...
+    if (!amount) return undefined
+    return new BigNumber(amount).toNumber().toLocaleString('fullwide', { useGrouping: false })
+  }
+
+  /**
+   * Get a basic quote (rate) for the pair
+   * @param input
+   */
+  async getQuote(input: GetQuoteInput): Promise<Quote | undefined> {
+    const {
```
bootsboogie, line 29: Im wondering if it would make sense to pull these methods out of the ZrxSwapper and have them in their own files.

My thought is that this file is going to get huge and it might be helpful to have them broken down into more digestible chucks and that way the test file wont get huge either.

thinking then all of our methods in this zrxSwapper would look more like this

```
async getQuote(input: GetQuoteInput): Promise<Quote | undefined> {
  return await importedGetQuote.bind(this, input)
}
```

DaoDev44 (approved):

packages/swapper/src/swappers/zrx/ZrxSwapper.ts
```diff
@@ -1,7 +1,21 @@
...
   getType() {
     return SwapperType.Zrx
   }
+
+  /**
+   * Get a basic quote (rate) for the pair
+   * @param input
+   */
+  async getQuote(input: GetQuoteInput): Promise<Quote | undefined> {
+    return getZrxQuote(input)
```
DaoDev44, line 19: Does this need to be `await`ed?

toshiSat: No return automatically awaits

packages/swapper/package.json
```diff
@@ -15,5 +15,11 @@
     "clean": "rm -rf dist",
     "test": "jest test",
     "type-check": "tsc --project ./tsconfig.json --noEmit"
+  },
+  "dependencies": {
+    "@shapeshiftoss/asset-service": "^1.0.0",
```
bootsboogie, line 20: Is this just a temporary dependency until we get a shared types package going?

bootsboogie (approved):
LGTM Just had one question?

packages/swapper/package.json
toshiSat: Yes, this will be the common package when we create it.  All I'm getting is types right now

packages/swapper/src/swappers/zrx/ZrxSwapper.ts
cjthompson: Well... That's not technically correct. This function simply returns a Promise. It doesn't wait for the Promise to be resolved before returning. The caller can await the Promise.

The async keyword tells the engine to wrap all return values in a Promise. In this case, the async keyword isn't necessary because the return value is already a Promise and await isn't called.

toshiSat: When it's an async function It automatically resolves the returned value.   It returns a Promise.resolve instead of a promise

toshiSat: wait or are you saying because the function is already resolving the promise it's unneeded as long as the return type is Promise<Quote> type?

PR #83, opened 2021-09-29
ci: add interim env var to avoid injection

0xdef1cafe:
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Commitlint PR title
         env:
           TITLE: ${{ github.event.pull_request.title }}
-        run: echo $TITLE | npx commitlint
+        run: printf '%s' "$TITLE" | npx commitlint
       - name: Build
         run: yarn build
       - name: Lint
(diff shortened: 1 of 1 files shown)


reidrankin (changes requested):
Sorry, but that's actually worse, because `echo $TITLE` will do word splitting -- you also need to do `printf '%s' "$TITLE" | npx commitlint`

.github/workflows/pr.yml
```diff
@@ -23,7 +23,9 @@ jobs:
         run: yarn install --frozen-lockfile
       # needs to run after install
       - name: Commitlint PR title
-        run: echo '${{ github.event.pull_request.title }}' | npx commitlint
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
+        run: echo $TITLE | npx commitlint
```
reidrankin, line 28:
```suggestion
        run: printf '%s' "$TITLE" | npx commitlint
```

reidrankin (approved):

cjthompson (approved):

bootsboogie (approved):

PR #89, opened 2021-09-30
chore: type txData in chain adapters

### LINKED ISSUE #92: Make chain adapters generic

0xdef1cafe:
this has flowing effects through web - @mvpratt we should talk more about if the abstraction should lie in unchained or chain adapters more before doing this, but it's a good example of where to capture the complexity.

closes #92


reidrankin (approved):

bootsboogie (approved):

natven:
Let's talk through this? I thought we were going to use generated types from the unchained api.

0xdef1cafe:
> Let's talk through this? I thought we were going to use generated types from the unchained api.

agreed, let's talk through it this afternoon before we merge this

0xdef1cafe:
as discussed @kaladinlight @natven @SterL closing this in favor of a "hybrid" of generic and chain specific info so clients consuming chain adapters don't have to switch at the top level everywhere

0xdef1cafe:
reopening this as we work on the hybrid approach

packages/types/src/types.ts
```diff
@@ -218,11 +235,48 @@ export type Transaction = {
...
+}
+
+export type Transaction<T extends ChainTypes = ChainTypes> = TransactionBase<T> & {
+  details: ChainFieldMap[T] & Record<string, undefined>
+}
+
+// unit tests for types
+
+// a generic transaction should have details.nonce of string | undefined
+ta.assert<ta.Equal<Transaction['details']['nonce'], string | undefined>>()
```
kaladinlight, line 258: Is this ta.assert stuff meant to live alongside types of would we want to move out into a .test file just to make this file a bit cleaner to look at?

reidrankin: I tried that a while ago on a different project; IIRC it doesn't actually work if we move it out into a test file because of magical jest reasons.

reidrankin: Works for me!

reidrankin:
ok, it passes tests now, though we probably need a little TLC on the commit history!

PR #91, opened 2021-10-01
feat: add getdefaultpair to zrxswapper

toshiSat:


packages/swapper/src/swappers/zrx/ZrxSwapper.test.ts
```diff
@@ -137,6 +137,11 @@ describe('ZrxSwapper', () => {
     await swapper.buildQuoteTx(args)
     expect(buildQuoteTx).toHaveBeenCalled()
   })
+  it('gets default pair', () => {
+    const swapper = new ZrxSwapper(zrxSwapperDeps)
+    const pair = swapper.getDefaultPair()
+    expect(pair).toHaveLength(2)
```
ryankeepkey, line 143: Do you think it would be helpful to expect on a specific value for the pair, ie `name = 'Ethereum'` or something like that?

toshiSat: Not really, it will just make it so when you change the values it will cause you to change it twice, but that does make me think.. we should probably at least check the chain type so that it is Ethereum there

elmutt (approved):

0xdef1cafe (approved):

PR #95, opened 2021-10-05
feat: add approval needed to lib

### LINKED ISSUE #60: Swapper in Lib: implement `approvalNeeded`
Implement method and tests

DaoDev44:
- Bring `approvalNeeded` over from platform-shared
- Add tests 
- Refactor `web3` logic inside of `getAllowanceRequired` to separate helper function

closes #60


elmutt (approved):

packages/swapper/src/swappers/zrx/approvalNeeded/approvalNeeded.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  it('returns false if sellAsset symbol is ETH', async () => {
+    const input = {
+      quote: { ...quoteInput, sellAsset: { ...sellAsset, symbol: 'ETH' } },
+      wallet
+    }
+
+    expect(await approvalNeeded(args, input)).toEqual({ approvalNeeded: false })
+  })
+
+  it('returns false if sellAsset chain is not ETH', async () => {
```
0xdef1cafe, line 61: not sure if this test is valid if 0x swapper only deals with eth?

packages/swapper/src/swappers/zrx/approvalNeeded/approvalNeeded.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getERC20Allowance } from '../utils/helpers/helpers'
+import { erc20AllowanceAbi } from '../utils/abi/erc20Allowance-abi'
+
+export async function approvalNeeded(
+  { adapterManager, web3 }: ZrxSwapperDeps,
+  { quote, wallet }: ApprovalNeededInput
+): Promise<ApprovalNeededOutput> {
+  const { sellAsset } = quote
+
+  if (sellAsset.symbol === 'ETH' || sellAsset.chain !== ChainTypes.Ethereum) {
```
0xdef1cafe, line 29: i would prob break this out, return false if symbol is eth, and throw if chain is not eth

0xdef1cafe (changes requested):

packages/swapper/src/swappers/zrx/approvalNeeded/approvalNeeded.test.ts
ryankeepkey: True, but once we implement other chain adapters there will be a chance where a different chain type can be passed in the `sellAsset`. This is just testing the guard clause so we don't send a receive address from the wrong chain.

0xdef1cafe: ok, ya so let's throw rather than return false if we pass a non eth chain asset to this swapper

0xdef1cafe (approved):

PR #97, opened 2021-10-06
ci: mark packages as public so they can be published

0xdef1cafe:
fixes this

https://github.com/shapeshift/lib/runs/3820210264

as per

https://github.com/lerna/lerna/issues/1821#issuecomment-510280765


phewitt (approved):
send it

PR #100, opened 2021-10-07
ci: bump package versions to allow semantic release to publish

0xdef1cafe:
we previously manually published `1.0.0` for the `asset-service` and semantic release tries to release the same version.

bump them all and it *should* work

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lib",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "private": true,
   "workspaces": {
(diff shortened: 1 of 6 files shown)


reidrankin (approved):
i like how you bumped the minor version instead of just the patch version to make it an order of magnitude more likely to just shut up and work

0xdef1cafe:
> i like how you bumped the minor version instead of just the patch version to make it an order of magnitude more likely to just shut up and work

you know whats up

SterL (approved):

PR #124, opened 2021-10-13
Btc unchained client

kaladinlight:


elmutt (approved):

BitHighlander:
> Nitpick: BIP32Params should be renamed BIP44Params before we make the wrong thing official -- I'm eating but will do the needful in a few

bip32 describes the path. the purpose can either be in reference to bip44/ or bip84 in that example case.  This lib project should also absolutely keep the same nomenclature as HDwallet that uses bip32path to describe to entire path string.

SterL (approved):

PR #140, opened 2021-10-20
fix: get pubkey params

SterL:


packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -130,25 +130,13 @@ export class BitcoinChainAdapter implements ChainAdapter<ChainTypes.Bitcoin> {
...
-          scriptType
-        }
-      ])
-
-      if (!(publicKeys ?? []).length) {
-        throw new Error('BitcoinChainAdapter: no public keys available from wallet')
-      }
-      const pubkey = publicKeys?.[0]?.xpub
+      const pubkey = await this.getPubKey(wallet, bip32Params)
       if (!pubkey) throw new Error('BitcoinChainAdapter: no pubkey available from wallet')
```
kaladinlight, line 134: We are already throwing this error essentially in `getPubKey` if we want to remove

elmutt (approved):

PR #146, opened 2021-10-21
fix: finish type namespacing

kaladinlight:


kaladinlight:
@natven @cjthompson Lower case types were just a stylistic decision that myself and parker thought made sense after looking at common practices around module aliasing and various example repos. However, I do agree this is purely stylistic and there is no right or wrong so if the consensus is pascal case, that change can be made.

@cjthompson I do not agree that these changes are just stylistic though. Providing namespacing to a shared types package to give context to what types are related to what lib packages, provides value imo. And simply breaking out one index.ts file of all types to organize things is useful from a continued maintenance pov as well.

PR #153, opened 2021-10-25
feat: market service market cap functionality

0xdef1cafe:
this PR

- exports things from caip i forgot to export
- generated market cap data keyed by caip19 ids
- adds `getByMarketCap` to the market service


elmutt (approved):

packages/market-service/src/coingecko/coingecko.test.ts
```diff
@@ -1,13 +1,129 @@
...
+      const perPage = 10
+      await getByMarketCap({ pages, perPage })
+      expect(mockedAxios.get).toHaveBeenCalledTimes(1)
+      const url =
+        'https://api.coingecko.com/api/v3/coins/markets?vs_currency=usd&order=market_cap_desc&per_page=10&page=1&sparkline=false'
+      expect(mockedAxios.get).toBeCalledWith(url)
+    })
+
+    it('can map coingecko to caip ids', async () => {
+      mockedAxios.get.mockResolvedValueOnce({ data: [btc] }).mockResolvedValue({ data: [eth] })
```
cjthompson, line 117: I wouldn't mind some "mockRejectedValue" tests to validate what is expected to happen in the API is down.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -38,6 +43,41 @@ const coingeckoIDMap: CoinGeckoIDMap = Object.freeze({
 export class CoinGeckoMarketService implements MarketService {
   baseUrl = 'https://api.coingecko.com/api/v3'
 
+  private readonly defaultGetByMarketCapArgs: GetByMarketCapArgs = {
+    pages: 10,
+    perPage: 250
+  }
+
+  async getByMarketCap(args?: GetByMarketCapArgs) {
+    const argsToUse = args ?? this.defaultGetByMarketCapArgs
```
cjthompson, line 52: you should merge the two instead so they can provide just "pages" or 'perPage" and still get the default for the other, unless that doesn't make sense.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -38,6 +43,41 @@ const coingeckoIDMap: CoinGeckoIDMap = Object.freeze({
...
+    const { pages, perPage } = argsToUse
+    const urlAtPage = (page: number) =>
+      `${this.baseUrl}/coins/markets?vs_currency=usd&order=market_cap_desc&per_page=${perPage}&page=${page}&sparkline=false`
+    const pageCount = Array(pages)
+      .fill(0)
+      .map((_v, i) => i + 1)
+    const combined = (
```
cjthompson, line 59: try/catch

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -38,6 +43,41 @@ const coingeckoIDMap: CoinGeckoIDMap = Object.freeze({
...
+      await Promise.all(
+        pageCount.map(async (page) => axios.get<CoinGeckoMarketCap>(urlAtPage(page)))
+      )
+    ).flat()
+    return combined
+      .map(({ data }) => data ?? []) // filter out rate limited results
+      .flat()
+      .sort((a, b) => (a.market_cap_rank > b.market_cap_rank ? 1 : -1))
+      .reduce((acc, cur) => {
```
cjthompson, line 68: except the sort, this could all be done in a single reduce.

cjthompson (approved):
Approved with comments

packages/market-service/src/coingecko/coingecko.ts
0xdef1cafe: done

packages/market-service/src/coingecko/coingecko.ts
0xdef1cafe: done

packages/market-service/src/coingecko/coingecko.ts
0xdef1cafe: agreed, i think this is a tad more reasonable. map is only operating on 10 items, flat is native, sort logic is very simple, and it keeps the biz logic to the reduce.

PR #176, opened 2021-10-29
style: bump prettier and enforce its formatting

reidrankin:
The `prettier` version we use is old and doesn't support `import type`. We've also been using eslint with the prettier plugin, but not prettier itself, so there's various formatting things that have slipped through that prettier doesn't like.

This bumps prettier (and unlocks its version), reformats everything (`yarn prettier -w .`), and adds an explicit prettier check to the lint step.


cjthompson:
Did you make any other changes besides the automatic linting? I can't look through 70k lines for changes.

cjthompson:
We should have talked about what prettier rules to use. Did you use the same ones from lib and web?

reidrankin:
I didn't add or remove any rules; this just just:
- bumps all the already-installed prettier package versions
- run `yarn prettier -w .`
- stuck explicit prettier runs into the lint commands in package.json

0xdef1cafe:
> I didn't add or remove any rules; this just:
> 
> * bumps all the already-installed prettier package versions
> * run `yarn prettier -w .`
> * stuck explicit prettier runs into the lint commands in package.json

this should be configured as a plugin for eslint it integrates with IDE's in one step and becomes a one step build failure.

0xdef1cafe:
also, you'll need to coordinate with anyone that has branches in progress that aren't already open as PRs, as this is the worlds nastiest merge conflict to fix.

PR #177, opened 2021-10-29
feat: move util functions from web

elmutt:
Move utils functions from web to lib


packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -197,7 +197,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
             addressNList: bip32ToAddressNList(
               `${path}/1/${String(account.chainSpecific.nextChangeAddressIndex)}`
             ),
-            scriptType: BTCOutputScriptType.PayToWitness,
+            scriptType: scriptType as unknown as BTCOutputScriptType,
```
reidrankin, line 200: this seems dangerous; what's going on here?

packages/chain-adapters/src/utils/utxoUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+}
+
+/**
+ * Returns coin purpose from a given script type
+ * @param scriptType
+ * @returns 44 | 49 | 84
+ */
+export const purposeFromScript = (
```
reidrankin, line 34: a purpose should not be derived from a script type; to do so assumes not only that every wallet follows BIP44, but that another BIP will always allocate a fresh purpose number to go with a new script type, and that's not guaranteed.

packages/chain-adapters/src/utils/utxoUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * Utility function to get params to get BIP32Params and scriptType for chain-adapter functions (getAddress, buildSendTransaction)
+ * Get return an object with utxo bip32Params derived from script type and asset
+ * @param scriptType
+ * @param asset
+ * @returns object with BIP32Params and scriptType or undefined
+ */
+export const bip32AndScript = (
```
reidrankin, line 55: I'm not sure what problem you're trying to solve here, but I don't think it should be solved this way.
- there's no need to return the script type if the caller knows it already
- you shouldn't get bip32params from an asset; you might be able to get a _default_ bip32 path from an asset, but an asset may be stored at any other path, and in the case of change addresses _definitely is_ in the normal use case.

reidrankin (dismissed):

packages/chain-adapters/src/utils/utxoUtils.ts
elmutt: what if I call it "bip44 purpose from script"?

packages/chain-adapters/src/utils/utxoUtils.ts
elmutt: I return the script thype so I can easily spread it and bip32Params into chain-adapter calls  in on easy line where its needed. namely getAddress and buildSendTx.

cjthompson (approved):

packages/types/src/base.ts
```diff
@@ -31,6 +31,12 @@ export enum NetworkTypes {
   ETH_RINKEBY = 'ETH_RINKEBY'
 }
 
+export enum UtxoAccountType {
+  SegwitNative = 'SegwitNative',
+  SegwitP2sh = 'SegwitP2sh',
+  P2pkh = 'P2pkh'
+}
```
reidrankin, line 38: I'm still fundamentally unhappy about the existence (and capitalization!) of this enum, but I can't think of anything better and l'll get over it.

reidrankin (approved):

PR #184, opened 2021-11-02
feat: asset service caip19s

0xdef1cafe:
this PR adds caip19 ids to generated assets in a non breaking way, and allows us to bring in market cap and asset data and join it in web


DaoDev44 (approved):

packages/swapper/package.json
```diff
@@ -21,6 +21,7 @@
     "swapcli": "yarn build && node ./dist/swappercli.js"
   },
   "dependencies": {
+    "@shapeshiftoss/asset-service": "^1.3.0",
```
cjthompson, line 24: Is this on purpose? It doesn't look like your make any other changes in `swapper`.

cjthompson (approved):

packages/swapper/package.json
0xdef1cafe: yeah - it doesn't build otherwise. i could move this to a devDependency as it's only used by swapper cli

PR #186, opened 2021-11-03
feat: dynamic btc fees

elmutt:
Dynamic btc fees from unchained

### DIFF
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.test.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.test.ts
@@ -122,6 +122,23 @@ const getTransactionMockResponse = {
   }
 }
 
+const getNetworkFeesMockedResponse = {
+  data: {
+    fast: {
+      blocksUntilConfirmation: 1,
+      satsPerKiloByte: 1024
+    },
+    average: {
+      blocksUntilConfirmation: 1,
+      satsPerKiloByte: 1024
+    },
+    slow: {
+      blocksUntilConfirmation: 1,
+      satsPerKiloByte: 1024
+    }
+  }
+}
+
 describe('BitcoinChainAdapter', () => {
   let args: bitcoin.ChainAdapterArgs = {} as any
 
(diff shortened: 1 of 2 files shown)


cjthompson:
Does this PR require an unchained PR to be merged and deployed? There are tests failing which I assume is because an updated unchained-client is required.

0xdef1cafe (approved):

elmutt:
> Does this PR require an unchained PR to be merged and deployed? There are tests failing which I assume is because an updated unchained-client is required.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.test.ts
```diff
@@ -122,6 +118,23 @@ const getTransactionMockResponse = {
   }
 }
 
+const getNetworkFeesMockedResponse = {
+  data: {
+    fast: {
+      blocksUntilConfirmation: 1,
+      satsPerKiloByte: 1024
```
natven, line 125: what's up with sats per kilo? txs can be significantly less than 1k

elmutt: thats how it comes back from unchained

natven (approved):

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.test.ts
kaladinlight: Value being returned from estimatesmartfee call: https://bitcoincore.org/en/doc/0.21.0/rpc/util/estimatesmartfee/

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -246,15 +246,29 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
...
+
+    if (
+      !feeData.data.fast?.satsPerKiloByte ||
+      !feeData.data.average?.satsPerKiloByte ||
+      !feeData.data.slow?.satsPerKiloByte
+    )
+      throw new Error('undefined fee')
+
+    // We have to round because coinselect library uses sats per byte which cant be decimals
+    const fast = String(Math.round(feeData.data.fast?.satsPerKiloByte / 1024))
```
pastaghost, line 259: Should we use Math.ceil() here instead to avoid underestimating fees?

PR #187, opened 2021-11-04
chore: bump and align package versions

0xdef1cafe:
needed to get web aligned on latest package versions

### DIFF
--- a/packages/chain-adapters/package.json
+++ b/packages/chain-adapters/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@shapeshiftoss/hdwallet-core": "^1.17.1",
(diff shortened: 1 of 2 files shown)


cjthompson:
There should be an updated yarn.lock also

0xdef1cafe:
> There should be an updated yarn.lock also

i've run yarn and it doesn't update the lock. it's because they're sub packages that don't have yarn locks, they're already in the root lockfile.

cjthompson (changes requested):
try using `peerDependencies`

cjthompson (approved):

kaladinlight (approved):

PR #190, opened 2021-11-04
fix: fees part 1 cleanup

elmutt:
- remove fees returning from buildSendTx
- remove recipients from buildSendTx
- add btc specific fee data and stub out return values

https://github.com/shapeshift/web/issues/215

We need to show fees on frontend and let them be selectable (it's all showing 0 right now) but the way we were expecting to do fees in frontend code was only applicable to eth and not the greatest (getting the fees from calling buildSendTx instead of getFees).

I will have a few additional changes to lib in upcoming prs but wanted to break it into smaller pieces.

### DIFF
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
@@ -5,7 +5,6 @@ import {
   BTCSignTx,
   BTCSignTxInput,
   BTCSignTxOutput,
-  BTCWallet,
   HDWallet,
   PublicKey,
   supportsBTC
@@ -162,7 +161,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
         scriptType
       })
 
-      if (!from) throw new Error('from undefined')
+      if (!from) throw new Error('BitcoinChainAdapter: from undefined')
 
       const account = await this.getAccount(pubkey.xpub)
       const estimatedFees = await this.getFeeData({ to, value, from })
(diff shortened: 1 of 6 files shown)


cjthompson:
Can you link to a ticket for this for context, or if there's no ticket, provide context. I don't understand the implications of the changes.

cjthompson (approved):
Approved but I don't have context around this change so I can't comment on correctness.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -125,34 +126,46 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
     tx: chainAdapters.BuildSendTxInput
   ): Promise<{
     txToSign: chainAdapters.ChainTxType<ChainTypes.Bitcoin>
-    estimatedFees: chainAdapters.FeeDataEstimate<ChainTypes.Bitcoin>
   }> {
     try {
       const {
         value,
         to,
-        recipients,
```
0xdef1cafe, line 134: remove this from the type too

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -125,34 +126,46 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
...
+          'BitcoinChainAdapter: signTransaction wallet does not support signing btc txs'
+        )
+
+      const from = await wallet.btcGetAddress({
+        addressNList,
+        coin: this.coinName,
+        scriptType
+      })
+
+      if (!from) throw new Error('from undefined')
```
0xdef1cafe, line 165:
```suggestion
      if (!from) throw new Error('BitcoinChainAdapter: from undefined')
```

0xdef1cafe (changes requested):

0xdef1cafe (approved):

PR #196, opened 2021-11-05
feat: add yearn erc20 assets to asset service

theoboldfrazier:


DaoDev44 (approved):

packages/asset-service/src/generateAssetData/exetendErc20.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+  endorsed: boolean
+  version: string
+  decimals: number
+  type: string
+  emergency_shutdown: boolean
+}
+
+export const extendErc20 = async (): Promise<TokenAsset[]> => {
+  const response: AxiosResponse = await axiosInstance.get(
```
0xdef1cafe, line 37: give this a type on the get request, like `get<YearnApiVault[]>(...` and you can destructure data straight off that.

also, we shouldn't try to cast and return `{}` as a TokenAsset, if we don't think it's a vault we should do a filter pass first.

lastly, don't cast `as TokenAsset`, do something like `const result: TokenAsset = {...` and let the compiler tell you what it needs to be

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -17,6 +19,15 @@ const generateAssetData = async () => {
     })
   )
 
+  const extendedERC20Tokens = await extendErc20()
+
+  const ethereumAsset = generatedAssetData.find(
+    (asset: Asset) => asset.caip19 === 'eip155:1/slip44:60'
+  )
+  if (ethereumAsset && extendedERC20Tokens?.length) {
+    ethereumAsset.tokens = flatten([ethereumAsset.tokens as TokenAsset[], extendedERC20Tokens])
```
0xdef1cafe, line 28: why are we casting here?

0xdef1cafe (changes requested):

packages/asset-service/src/generateAssetData/generateAssetData.ts
theoboldfrazier: My compiler was showing the tokens as any[]. I was trying to get it to recognize it as a TokenAsset. Wasn't necessary. Taken out.

packages/asset-service/src/generateAssetData/exetendErc20.ts
theoboldfrazier: Good call. I've updated it.

packages/asset-service/src/generateAssetData/generateAssetData.ts
0xdef1cafe: don't you want to do `flatten([ethereumAsset.tokens ?? [], extendedERC20Tokens]` because the compiler should already know it's an array of `TokenAsset` if it's defined?

theoboldfrazier: That makes sense. Done.

0xdef1cafe (approved):

PR #199, opened 2021-11-09
fix: fees

elmutt:
- Make fee types more logical and add chain specific data
- finish getFeeData for btc

https://github.com/shapeshift/web/issues/215

### DIFF
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import {
   bip32ToAddressNList,
   BTCInputScriptType,
@@ -18,6 +19,7 @@ import { ChainAdapter as IChainAdapter } from '../api'
 import { ErrorHandler } from '../error/ErrorHandler'
 import { toPath, toRootDerivationPath } from '../utils/bip32'
 import { toBtcOutputScriptType } from '../utils/utxoUtils'
+import BigNumber from 'bignumber.js'
 
 export interface ChainAdapterArgs {
   providers: {
@@ -139,8 +141,6 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
         throw new Error('BitcoinChainAdapter: (to and value) are required')
       }
 
-      const btcRecipients = [{ value: Number(value), address: to }]
-
       const path = toRootDerivationPath(bip32Params)
       const changeScriptType = toBtcOutputScriptType(scriptType)
       const pubkey = await this.getPubKey(wallet, bip32Params, scriptType)
@@ -169,7 +169,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
 
       const coinSelectResult = coinSelect<MappedUtxos, chainAdapters.bitcoin.Recipient>(
         mappedUtxos,
-        btcRecipients,
+        [{ value: Number(value), address: to }],
         Number(satoshiPerByte)
       )
       if (!coinSelectResult.inputs) {
@@ -177,6 +177,8 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
       }
 
       const { inputs, outputs, fee } = coinSelectResult
+
+      console.log('coinSelectResult fee is', fee)
 
       const signTxInputs: BTCSignTxInput[] = []
       for (const input of inputs) {
@@ -219,11 +221,10 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
         }
       })
 
-      const txToSign: BTCSignTx & { fee: number } = {
+      const txToSign: BTCSignTx = {
         coin: this.coinName,
         inputs: signTxInputs,
-        outputs: signTxOutputs,
-        fee
+        outputs: signTxOutputs
       }
 
       return { txToSign }
@@ -257,11 +258,15 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
   async getFeeData({
     to,
     from,
-    value
+    value,
+    wallet,
+    bip32Params,
+    scriptType
   }: chainAdapters.GetFeeDataInput): Promise<chainAdapters.FeeDataEstimate<ChainTypes.Bitcoin>> {
     const feeData = await this.providers.http.getNetworkFees()
 
-    if (!to || !from || !value) throw new Error('to, from and value are required')
+    if (!to || !from || !value || !wallet || !bip32Params || !scriptType)
+      throw new Error('to, from, wallet, bip32Params, scriptType and value are required')
 
     if (
       !feeData.data.fast?.satsPerKiloByte ||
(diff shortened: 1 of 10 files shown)


DaoDev44 (approved):

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -271,30 +269,50 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
       throw new Error('undefined fee')
 
     // We have to round because coinselect library uses sats per byte which cant be decimals
-    const fast = String(Math.round(feeData.data.fast?.satsPerKiloByte / 1024))
-    const average = String(Math.round(feeData.data.average?.satsPerKiloByte / 1024))
-    const slow = String(Math.round(feeData.data.slow?.satsPerKiloByte / 1024))
+    const fastPerByte = String(Math.round(feeData.data.fast?.satsPerKiloByte / 1024))
+    const averagePerByte = String(Math.round(feeData.data.average?.satsPerKiloByte / 1024))
+    const slowPerByte = String(Math.round(feeData.data.slow?.satsPerKiloByte / 1024))
```
cjthompson, line 274: ?. can result in undefined, so what does the output look like when that is the case? It should be easy to add test cases to cover that scenario

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -271,30 +269,50 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
...
+
+    const { data: utxos } = await this.providers.http.getUtxos({
+      pubkey
+    })
+
+    type MappedUtxos = Omit<bitcoin.api.Utxo, 'value'> & { value: number }
+    const mappedUtxos: MappedUtxos[] = utxos.map((x) => ({ ...x, value: Number(x.value) }))
```
cjthompson, line 281: Number is lossy. Is that ok here or should we use BigNumber to maintain accuracy?

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -271,30 +269,50 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
...
+    const { fee: slowFee } = coinSelect<MappedUtxos, chainAdapters.bitcoin.Recipient>(
+      mappedUtxos,
+      [{ value: Number(value), address: to }],
+      Number(slowPerByte)
+    )
 
     const confTimes: chainAdapters.FeeDataEstimate<ChainTypes.Bitcoin> = {
       [chainAdapters.FeeDataKey.Fast]: {
-        feePerUnit: fast,
+        txFee: String(fastFee),
```
cjthompson, line 301: Has this value been checked to be a valid numeric value first?

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -271,30 +269,50 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
...
+    const { fee: averageFee } = coinSelect<MappedUtxos, chainAdapters.bitcoin.Recipient>(
+      mappedUtxos,
+      [{ value: Number(value), address: to }],
+      Number(averagePerByte)
+    )
+    const { fee: slowFee } = coinSelect<MappedUtxos, chainAdapters.bitcoin.Recipient>(
+      mappedUtxos,
+      [{ value: Number(value), address: to }],
+      Number(slowPerByte)
+    )
```
cjthompson, line 297: I would recommend using the helper function "bnOrZero" which will return 0 if the input is not valid and returns a BigNumber.

This will guard against unexpected values and make sure we don't get weird math that ends up as NaN.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
kaladinlight: There is a check above which should narrow and not require the optional chaining here. This is a scenario that may be better moved into the unchained api as well where we throw if fees are not correctly returned from the node

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
kaladinlight: This number cast scenario is to conform with the `coinselect` interface, so not sure what we can do about that

packages/types/src/chain-adapters/ethereum.ts
```diff
@@ -46,7 +46,11 @@ export type QuoteFeeData = {
...
 
 export type BuildTxInput = {
-  fee: string
+  gasPrice: string
   gasLimit: string
   erc20ContractAddress?: string
 }
+export type GetFeeDataInput = {
+  contractAddress?: string
+  from: string
```
kaladinlight, line 55: since both chains and presumably more in the future have some sense of a from address/xpub. I wonder if this would make more sense as part of the base GetFeeDataInput type. We can just use the `pubkey` vernacular to imply "this is your from address or xpub"

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
kaladinlight: Numeric value being returned from `coinselect` package.

kaladinlight (approved):
One comment about "from" address/xpub for getFeeData function, otherwise looks good. Will add issue to do data checking on `FeeData` in unchained api to reduce load on client ingestion.

packages/types/src/chain-adapters/ethereum.ts
elmutt: ya thats a good point. Id prefer to leave it for now because everything is working as-is but its not hard to change and is a relevant point

Your comment also reminds me that a lot of this chainSpecific stuff seems kindof pointless

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: good point, got rid of the optional chaining

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: iI believe its fine because thats the format coinSelect expects it. We do it like this in buildSendTransactions for bitcoin and it hasnt been a problem.

they also use a number here in their example:

https://github.com/bitcoinjs/coinselect

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: should be good here from coinselect number value

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: Are you sure its needed here? Your saying fastFee/averageFee/slowFee should bet set in a bnOrZero?

In some cases I feel like setting to 0 instead of nan could be unintentionally hiding errors

0xdef1cafe (approved):

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
cjthompson: and coinselect has a strong guarantee on the return value?

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
cjthompson: I still think that we should consistently use the same strategy for holding and passing around numbers (like BigNumber) that avoids loss of precision.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
cjthompson: Why do you cast these as String here and then re-cast them as Number later?

If you need the value in both formats, putting in a BigNumber allows you to call `toNumber` and `toString` on the same value.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: So your saying `value` in `GetFeeDataInput` should be a BigNumber? 

I guess I see the point from a consistency standpoint but it needs toe be converted to a number for coinselect and its not being modified in any way before being passed to coinSelect.

Sorry I merged this before i saw this comment

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
elmutt: Sorry i didnt see this until after I merged. I can follow up with a small pr checking that it is a defined number

PR #203, opened 2021-11-09
chore: refactor bitcoin.getFeeData

cjthompson:
* DRY up the code
* allow flexibility to have `n` estimated fees instead of always being `3`
* if some fees aren't available, return undefined so we can return at least some fee data if it's available (instead of all or nothing)

### DIFF
--- a/packages/types/src/chain-adapters/bitcoin.ts
+++ b/packages/types/src/chain-adapters/bitcoin.ts
@@ -75,7 +75,7 @@ export type NodeTransaction = {
 }
 
 export type FeeData = {
-  satoshiPerByte: string
+  satoshiPerByte?: string
 }
 
 export type BuildTxInput = {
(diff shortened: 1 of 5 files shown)


packages/types/src/chain-adapters/bitcoin.ts
```diff
@@ -75,7 +75,7 @@ export type NodeTransaction = {
 }
 
 export type FeeData = {
-  satoshiPerByte: string
+  satoshiPerByte?: string
```
elmutt, line 78: should this ever be optional? seems like we would always need it

cjthompson:
I want to get some feedback on whether we should try to return partial fee results, if we have them, or only return data if we have all three estimated fees available. The former is more flexible, allows us to show something instead of nothing, but the latter is simpler (either handle an error or show all the fees).

PR #207, opened 2021-11-10
fix: asset service initialize

technojak:


elmutt (approved):

packages/asset-service/src/service/AssetService.ts
```diff
@@ -76,6 +76,7 @@ export class AssetService {
    */
   async initialize() {
     try {
+      if (!this.assetFileUrl) throw new Error('No assetFileUrl')
```
cjthompson, line 79: Is it worth try to parse it as a valid URL and throw an error if not? Probably not.

cjthompson (approved):

packages/asset-service/src/service/AssetService.ts
technojak: Yeah I was thinking about that as well. I'm not opposed to doing so but might be overkill in this case.

PR #225, opened 2021-11-13
fix: portis wallet not showing BTC balance

cjthompson:
# This PR contains a breaking change!

* [MAJOR] fix: replace references to "scriptType" with "accountType"
* feat: add "convertXpubVersion" function to convert any public key to an x-, y-, or z-pub

## Description
The purpose of this change is that `scriptType` can be assumed based on the `accountType`, but not the other way around. We should be passing around the `accountType` and then deriving the `scriptType` to use from that.

This fixes portis BTC balances not working because 
1. Portis always returns its public key in "xpub" format, even for Segwit public keys.
2. Blockbook generates addresses from a public key by INFERRING the account type from the public key prefix of `"xpub" | "ypub" | "zpub"`.
3. We now convert every public key we get into an `"xpub" | "ypub" | "zpub"` based on the `accountType` that's provided

### DIFF
--- a/packages/chain-adapters/package.json
+++ b/packages/chain-adapters/package.json
@@ -38,6 +38,10 @@
   },
   "devDependencies": {
     "@shapeshiftoss/hdwallet-native": "^1.17.1",
+    "@types/bs58check": "^2.1.0",
     "@types/multicoin-address-validator": "^0.5.0"
+  },
+  "peerDependencies": {
+    "bs58check": "^2.0.0"
   }
 }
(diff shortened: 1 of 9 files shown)


cjthompson:
The build has an error on a line that doesn't even exist in the repo:
```
src/bitcoin/BitcoinChainAdapter.ts:395:56 - error TS2304: Cannot find name 'scriptType'.

395     const id = `${toRootDerivationPath(bip32Params)}/${scriptType}`
                                                           ~~~~~~~~~~
```
I grepped the entire repo for "const id =" and there's only 2 in market-service that aren't anything like this.

natven (approved):

PR #228, opened 2021-11-15
fix: update deps

elmutt:
fix: update deps

### DIFF
--- a/packages/asset-service/package.json
+++ b/packages/asset-service/package.json
@@ -25,7 +25,7 @@
     "type-check": "tsc --project ./tsconfig.json --noEmit"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


DaoDev44 (approved):

cjthompson (commented):
The ironic thing here is that by bumping these version requirements, it'll increment the version numbers on those three packages, making them be out of date again in web.

0xdef1cafe:
> The ironic thing here is that by bumping these version requirements, it'll increment the version numbers on those three packages, making them be out of date again in web.

we're aware, and the fix is to make the packages dependency injected from the consumer side to remove this coupling in lib, but this will get things unblocked and web can use the new package versions

0xdef1cafe (approved):

PR #233, opened 2021-11-16
feat: send trade details over websocket from eth chain adapter

elmutt:
send trade details over websocket from eth chain adapter

### DIFF
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
@@ -301,6 +301,25 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
           }
         })
 
+        if (msg.trade) {
+          Object.entries(msg.trade).forEach(([symbol, { totalValue, token }]) => {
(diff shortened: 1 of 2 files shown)


DaoDev44 (approved):

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -301,6 +301,25 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
           }
         })
 
+        if (msg.trade) {
```
kaladinlight, line 304: This will send a transaction message for a trade, but there will also be a send and receive message sent for the same transaction based on the current logic.

There is also no `symbol`, `totalValue`, or `token` on the `Trade` payload for sending the `specificTx` function. There will only ever be a single trade payload per transaction as well (no use for Object.entries)

elmutt: ok

kaladinlight (approved):

0xdef1cafe (approved):

PR #256, opened 2021-11-27
fix: speed up invert of generated caip info

reidrankin:
This is, as far as I can tell, the biggest culprit in the Case of the Slow Page Load; the object allocation is high-overhead by itself, but the spread operator also gets transpiled down to use a babel shim that allocates extra temporary arrays of the intermediate objects' properties and loops over them.


elmutt (approved):

0xdef1cafe (approved):

0xdef1cafe:
my bad as i didn't intend this path to get hit at runtime, just generate, good catch

PR #266, opened 2021-12-03
feat: update subscribeTxs to use new tx parser payloads

kaladinlight:
- Update subscribeTxs to ingest new tx parser payloads
- Bump unchained package versions to v4.0.0
- Use caip2/caip19 in subscribeTxs (follow up PR to update getAccount)
- Helper functions to convert unchained types into chain adapter types
- Remove token data from tx message payloads (force use of asset service)
- Clean up types


packages/chain-adapters/package.json
```diff
@@ -25,6 +25,7 @@
     "type-check": "tsc --project ./tsconfig.json --noEmit"
   },
   "dependencies": {
+    "@shapeshiftoss/caip": "^1.5.0",
```
theoboldfrazier, line 28: I think you'll want to put this in the peerDependencies object. Helps with versioning.

kaladinlight: Yep, good call. I see those changes now, thanks!

kaladinlight: Is there a reason for things to be both devDependecies and peerDependencies? or is that leftover cruft?

theoboldfrazier: That's a good question, I think it needs to be in both. @majorhayes added it to both when he made these changes last week.

theoboldfrazier (approved):

packages/chain-adapters/package.json
kaladinlight: A little googling and it seems like peer deps are installed by default. I tried the --peer flag with yarn and that also didn't seem to work. Seems like there is a package that can do this for us, but not a huge deal right now so just adding the dupes

0xdef1cafe (approved):

PR #269, opened 2021-12-04
feat: caip10 account spec

0xdef1cafe:
* adds isCAIP2 function and tests for validation
* adds caip10 (account) support, with toCAIP10 and fromCAIP10


natven (approved):
🍭

PR #270, opened 2021-12-06
Feature osmosis

BitHighlander:


packages/types/src/base.ts
```diff
@@ -21,7 +21,14 @@ export enum ContractTypes {
 
 export enum ChainTypes {
   Ethereum = 'ethereum',
-  Bitcoin = 'bitcoin'
+  Bitcoin = 'bitcoin',
+  Osmosis = 'osmosis',
+  Cosmos = 'cosmos',
```
0xdef1cafe, line 26: i know this is preemptive and a draft PR, but let's not add these until we're ready to implement throughout the stack, including web, as we exhaust these ChainTypes in switch statements and throw in a lot of places on unimplemented.

PR #288, opened 2021-12-17
feat: add EIP-1559 parameters to getFeeData return object

pastaghost:


pastaghost:
@kaladinlight 12/17 - 03:53 Ignore the current build errors. This work depends on your corresponding unchained pr, which hasn't yet been merged.

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -308,14 +333,16 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
...
       { topic: 'txs', addresses: [address] },
-      (msg) => {
-        const transfers = msg.transfers.map<chainAdapters.TxTransfer>((transfer) => ({
-          caip19: transfer.caip19,
-          from: transfer.from,
-          to: transfer.to,
-          type: getType(transfer.type),
-          value: transfer.totalValue
-        }))
+      (msg: chainAdapters.SubscribeTxsMessage<ChainTypes.Ethereum>) => {
```
kaladinlight, line 336: These changes to subscribeTxs _shouldn"t_ be required I wouldn't think (both eth and btc)

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -253,26 +253,51 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
...
+          maxFeePerGas: String(
+            new BigNumber(feeData.maxFeePerGas).times(normalizationConstants.instant)
+          ),
+          maxPriorityFeePerGas: String(
+            new BigNumber(feeData.maxPriorityFeePerGas).times(normalizationConstants.instant)
+          )
         }
       },
       average: {
         txFee: new BigNumber(fees.fast).times(gasLimit).toPrecision(),
```
kaladinlight, line 278: If eip fees confirmed working, it seems like we would want that to be our default for `txFee`

packages/chain-adapters/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shapeshiftoss/chain-adapters",
-  "version": "1.22.1",
+  "version": "1.21.1",
```
kaladinlight, line 3: No manual change here. Leave at version from main and CI will bump and publish

packages/chain-adapters/package.json
```diff
@@ -37,19 +37,19 @@
     "@shapeshiftoss/caip": "^1.5.1",
     "@shapeshiftoss/hdwallet-core": "^1.18.2",
     "@shapeshiftoss/hdwallet-native": "^1.18.2",
-    "@shapeshiftoss/types": "^1.13.1",
-    "@shapeshiftoss/unchained-client": "^4.0.0",
-    "@shapeshiftoss/unchained-tx-parser": "^4.0.0",
+    "@shapeshiftoss/types": "^1.19.1",
```
kaladinlight, line 40: Looks like types package is at `1.20.0` on main if we want to use most recent. Don't think this version matters though by itself as it is a peer dependency. Only version that would affect chain adapters itself would be the `devDependencies` version number. (That is my understanding at least...)

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -458,7 +458,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
           txid: msg.txid
         })
       },
-      (err) => onError({ message: err.message })
+      (err: chainAdapters.SubscribeError) => onError({ message: err.message })
```
kaladinlight, line 461: Should not need this type and it is also the incorrect type being returned by `subscribeTxs` in unchained

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -332,7 +369,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
           txid: msg.txid
         })
       },
-      (err) => onError({ message: err.message })
+      (err: chainAdapters.SubscribeError) => onError({ message: err.message })
```
kaladinlight, line 372: Same type comment here

kaladinlight (approved):

0xean (approved):

asamere (approved):

PR #289, opened 2021-12-20
feat: add cosmos chain adapter

elmutt:


packages/caip/src/caip2/caip2.ts
```diff
@@ -19,6 +20,7 @@ export enum ChainReference {
   // caip2 uses max length of 32 chars of the genesis block
   BitcoinMainnet = '000000000019d6689c085ae165831e93',
   BitcoinTestnet = '000000000933ea01ad0ee984209779ba',
+  CosmosMainnet = 'cosmoshub-4',
   CosmosHubMainnet = 'cosmoshub-4',
```
kaladinlight, line 24: Seems like there was another caip2 merged through recently that now is kind of duplicated with what was in place here. that we should clean up

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -27,6 +27,10 @@ const unchainedUrls = {
   [ChainTypes.Ethereum]: {
     httpUrl: 'https://dev-api.ethereum.shapeshift.com',
     wsUrl: 'wss://dev-api.ethereum.shapeshift.com'
+  },
+  [ChainTypes.Cosmos]: {
+    httpUrl: 'http://api.cosmos.localhost',
+    wsUrl: ''
```
kaladinlight, line 33:
```suggestion
    wsUrl: 'ws://api.cosmos.localhost'
```

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -76,13 +80,12 @@ const main = async () => {
         txToSign: btcUnsignedTx.txToSign
       })
       console.log('btcSignedTx:', btcSignedTx)
+      const btcTxID = await btcChainAdapter.broadcastTransaction(btcSignedTx)
+      console.log('btcTxID: ', btcTxID)
```
kaladinlight, line 84: Do we want to leave commented so an accidental transaction isn't broadcasted?

packages/chain-adapters/src/api.ts
```diff
@@ -55,10 +49,4 @@ export type ChainAdapter<T extends ChainTypes> = {
   unsubscribeTxs(input?: chainAdapters.SubscribeTxsInput): void
 
   closeTxs(): void
-} & ChainSpecificAdapterProps<
```
kaladinlight, line 58: This stuff seems unrelated, not sure if there is reason behind this.

kaladinlight:
Need to bump unchained packages to v5.1.1 for cosmos api I think.

packages/chain-adapters/src/api.ts
0xdef1cafe: this will break validating ens address if we remove it i'm pretty sure?

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -0,0 +1,211 @@
+/* eslint-disable prettier/prettier */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
0xdef1cafe, line 3:
```suggestion
```

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -0,0 +1,211 @@
...
+      return {
+        balance: data.balance,
+        caip2: caip,
+        caip19: caip19.toCAIP19({ chain, network }),
+        chain: this.getType(),
+        chainSpecific: {
+          sequence: data.sequence
+        },
+        pubkey: data.pubkey
+      } as chainAdapters.Account<T>
```
0xdef1cafe, line 78: would prefer a `const result: chainAdapters.Account<T> = {...}; return result` here rather than a cast here if possible

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable prettier/prettier */
```
0xdef1cafe, line 1:
```suggestion
```

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -0,0 +1,127 @@
...
+} from '@shapeshiftoss/hdwallet-core'
+import { BIP44Params, chainAdapters, ChainTypes } from '@shapeshiftoss/types'
+import BigNumber from 'bignumber.js'
+
+import { ChainAdapter as IChainAdapter } from '../../api'
+import { ErrorHandler } from '../../error/ErrorHandler'
+import { toPath } from '../../utils'
+import { ChainAdapterArgs, CosmosSdkBaseAdapter } from '../CosmosSdkBaseAdapter'
+
+// import { cosmos } from '@shapeshiftoss/unchained-client'
```
0xdef1cafe, line 17:
```suggestion
```

packages/chain-adapters/src/cosmossdk/osmosis/OsmosisChainAdapter.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable prettier/prettier */
```
0xdef1cafe, line 1:
```suggestion
```

0xdef1cafe (changes requested):
at a first glance, we shouldn't be turning off the linter or prettier anywhere, and there's some casting we can avoid, but i haven't dug really deeply into any particular function impl.

packages/chain-adapters/src/api.ts
pastaghost: We're going to have to figure something out here, then. Changing the ChainAdapter definition from an `interface` to a `type` breaks the templated inheritance model for CosmosSdkBaseAdapter.

```
src/cosmossdk/CosmosSdkBaseAdapter.ts:25:83 - error TS2422: A class can only implement an object type or intersection of object types with statically known members.

25 export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implements IChainAdapter<T> {
```

packages/caip/src/caip19/caip19.ts
```diff
@@ -19,7 +19,8 @@ export enum AssetNamespace {
 export enum AssetReference {
   Bitcoin = 0,
   Ethereum = 60,
-  Cosmos = 118
+  Cosmos = 118,
+  Osmosis = 118
```
theoboldfrazier, line 23: Anyway we can close and re-open this PR to remove some of the github diff noise?

PR #292, opened 2021-12-21
feat: short circuit market data calls to providers that don't support that asset

### LINKED ISSUE #232: Don't make market data calls for assets not supported by market data provider
0x4bf5dc91e2555449293d7824028eb8fe5879b689

coingecko doesn't list as an asset, so we fail when looking up details for it.

need a concept of a blacklist as this will happen more often.

theoboldfrazier:
closes #232


0xdef1cafe (approved):

0xdef1cafe:
very noice

packages/caip/package.json
```diff
@@ -22,10 +22,14 @@
     "dev": "tsc --watch",
     "generate-coingecko": "yarn build && node dist/adapters/coingecko/generate.js",
     "generate-coincap": "yarn build && node dist/adapters/coincap/generate.js",
+    "generate-yearn": "yarn build && node dist/adapters/yearn/generate.js",
+    "generate": "yarn build && node dist/adapters/coingecko/generate.js && node dist/adatpters/coincap/generate.js && node dist/adapters/yearn/generate.js",
     "test": "jest test",
     "type-check": "tsc --project ./tsconfig.json --noEmit"
   },
-  "dependencies": {},
+  "dependencies": {
```
0xdef1cafe, line 30: actually, can we make this a dev dep please? it's only needed for generate script which is most def a dev task

0xdef1cafe (changes requested):

packages/caip/package.json
theoboldfrazier: @0xdef1cafe Good call, done.

cjthompson (approved):

0xdef1cafe (approved):

Issue #181, opened 2021-11-02
Bump prettier


mrnerdhair:
We need to bump `prettier` to get support for new language features, but that'll probably come with reformatting hell (a la #176).

PR #295, opened 2022-01-09
feat: support improved performance via dynamic imports

mrnerdhair:
- Loading web3-utils is a very expensive operation just to get numberToHex. I've copied over (and typed) the implementation more-or-less verbatim -- we might want to use another library or different impl though.
- Initializing AssetService is a CPU-heavy process, but in order to offload it to a worker the API needs to be full-async.
- Once again #181 prevents us from having some nice things.

### DIFF
--- a/packages/asset-service/src/service/AssetService.test.ts
+++ b/packages/asset-service/src/service/AssetService.test.ts
@@ -50,17 +50,19 @@ describe('AssetService', () => {
   })
 
   describe('byNetwork', () => {
-    it('should throw if not initialized', () => {
+    it('should throw if not initialized', async () => {
       const assetService = new AssetService(assetFileUrl)
       mockedAxios.get.mockResolvedValue({ data: mockBaseAssets })
-      expect(() => assetService.byNetwork(NetworkTypes.MAINNET)).toThrow(Error)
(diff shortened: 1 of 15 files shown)


0xdef1cafe:
@reidrankin how much performance do we empirically gain from this, vs the added complexity of workers, in terms of developer experience and maintenance? i'm ok with making the asset service async as it's only called from one place in the front end now, but have some concerns about the additional complexity added to swapper.

reidrankin:
This isn't worker-specific, but avoiding a blocking load of `web3` (and by extension `web3-utils`) saves ~500ms in dev mode. When that lib gets pulled in, it assumes it will be used for Many Important Things and not just string conversions, so it does some sort of elliptic-curve math upfront. Even ignoring the potential of offloading whatever it might want to do to a worker, just loading this stuff via dynamic `import()` speeds up the time that React can start rendering the page significantly.

0xdef1cafe:
as discussed via discord, can you please

* remove asset service changes
* remove new explicit `bn.js` dep
* investigate using `bignumber` for the `numberToHex` implementation
* speak to @theoboldfrazier about `investor-yearn` package async loading too

0xdef1cafe:
@mrnerdhair closing for housekeeping until this is ready

PR #296, opened 2022-01-09
feat: add ENS validation

gomesalexandre:
This adds `chain-adapters` support for ENS address in address validation.

### DIFF
--- a/packages/chain-adapters/src/api.ts
+++ b/packages/chain-adapters/src/api.ts
@@ -41,6 +41,9 @@ export interface ChainAdapter<T extends ChainTypes> {
   broadcastTransaction(hex: string): Promise<string>
 
   validateAddress(address: string): Promise<chainAdapters.ValidAddressResult>
+  validateEnsAddress?: T extends ChainTypes.Ethereum
+    ? (address: string) => Promise<chainAdapters.ValidAddressResult>
+    : never
 
   subscribeTxs(
     input: chainAdapters.SubscribeTxsInput,
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
@@ -291,7 +291,9 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
 
   async validateAddress(address: string): Promise<chainAdapters.ValidAddressResult> {
     const isValidAddress = WAValidator.validate(address, this.getType())
-    if (isValidAddress) return { valid: true, result: chainAdapters.ValidAddressResultType.Valid }
+    const isValidEnsAddress = new RegExp(/([A-Za-z]*\.)?([A-Za-z]{3,})\.eth/).test(address)
+    if (isValidAddress || isValidEnsAddress)
+      return { valid: true, result: chainAdapters.ValidAddressResultType.Valid }
     return { valid: false, result: chainAdapters.ValidAddressResultType.Invalid }
   }
 
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
@@ -291,7 +291,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
 
   async validateAddress(address: string): Promise<chainAdapters.ValidAddressResult> {
     const isValidAddress = WAValidator.validate(address, this.getType())
-    const isValidEnsAddress = new RegExp(/([A-Za-z]*\.)?([A-Za-z]{3,})\.eth/).test(address)
+    const isValidEnsAddress = /^([0-9A-Z]([-0-9A-Z]*[0-9A-Z])?\.)+eth$/i.test(address)
     if (isValidAddress || isValidEnsAddress)
       return { valid: true, result: chainAdapters.ValidAddressResultType.Valid }
     return { valid: false, result: chainAdapters.ValidAddressResultType.Invalid }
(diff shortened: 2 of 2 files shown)


packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -191,9 +191,11 @@ export class EthereumChainAdapter implements ChainAdapter {
     return ethAddress as string
   }
 
-  async validateAddress(address: string): Promise<ValidAddressResult> {
+  validateAddress(address: string): ValidAddressResult {
     const isValidAddress = WAValidator.validate(address, this.getType())
-    if (isValidAddress) return { valid: true, result: ValidAddressResultType.Valid }
+    const isValidEnsAddress = new RegExp(/([A-Za-z]*\.)?([A-Za-z]{3,})\.eth/).test(address)
```
gomesalexandre, line 196: Regex check might not be optimal here. As per the [EIP-137](https://eips.ethereum.org/EIPS/eip-137) spec, names can be more complex than this format, especially with internationalization. The spec recommends the [idna-uts46](https://www.npmjs.com/package/idna-uts46) lib, but [ens-normalize](`https://github.com/adraffy/ens-normalize.js`) seems to be a better fit API for these needs.

packages/chain-adapters/src/api.ts
```diff
@@ -153,5 +153,5 @@ export interface ChainAdapter {
 
   broadcastTransaction(hex: string): Promise<string>
 
-  validateAddress(address: string): Promise<ValidAddressResult>
+  validateAddress(address: string): ValidAddressResult
```
reidrankin, line 156: This intentionally returns a Promise even though it could be implemented synchronously at the moment in order to maintain implementation flexibility.

gomesalexandre: Makes sense! I'll revert it.

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
gomesalexandre: Since reverse lookup is done in https://github.com/shapeshift/web/pull/704/files it's probably enough to test it against regex. Support for Rinkeby's `.test` domains could be added when support for testnets is added.

gomesalexandre:
Rebase ✅

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -291,7 +291,9 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
 
   async validateAddress(address: string): Promise<chainAdapters.ValidAddressResult> {
     const isValidAddress = WAValidator.validate(address, this.getType())
-    if (isValidAddress) return { valid: true, result: chainAdapters.ValidAddressResultType.Valid }
+    const isValidEnsAddress = new RegExp(/([A-Za-z]*\.)?([A-Za-z]{3,})\.eth/).test(address)
```
reidrankin, line 294: - `/foobar/` is a regex literal, so no need to call `new RegExp()`
- need to match the whole string
- regex now respects RFC 1123
```suggestion
    const isValidEnsAddress = /^([0-9A-Z]([-0-9A-Z]*[0-9A-Z])?\.)+eth$/i.test(address)
```

reidrankin (changes requested):

packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
gomesalexandre: Nice, thanks! Before I commit these improvements, regarding RFC 1123, what do you think bringing back the `{3,}` repeater on the last group?
ENS has at least one gotcha compared to it
If you try to register a sub/domain name on ENS where the domain is less than 3 characters long, you will have an error:
https://app.ens.domains/name/test.vi.eth 🚫 
https://app.ens.domains/name/test.vit.eth/subdomains ✅

reidrankin: Did some looking at the ENS contracts to hunt this down -- I'm pretty sure that's an artificial restriction. There's nothing preventing names from being shorter except that the current registrar controller for the `.eth` domain won't register them for you. But if you had one, it would totally be resolvable, and even transferable and renewable. So IMO we should leave the regex open-structured since future controller implementations relax that restriction.

gomesalexandre: That's probably why I didn't find anything on the spec about it then. 
If it's something registar-specific and it's valid according to , I agree it should be returned as a valid address, because it is.

0xdef1cafe: do we want to consider supporting all TLDs as this is now supported by ENS, or is that too premature?

https://medium.com/the-ethereum-name-service/how-eth-names-compare-to-dns-names-imported-to-ens-ab855a6dbdb2

gomesalexandre: Ohhh that's a valid point. TBH I haven't seen a single DNS domain in the wild ever (that is, someone posting their regular DNS domain imported in ENS with a message like "guys this is my address).
But if it's valid it's valid, I'd say let's add it

reidrankin: I say it's premature. I'd want a curated whitelist of TLDs we're willing to resolve, given that they have different security properties than .eth, and the regex is a handy point for expanding those in the future.

reidrankin (approved):

reidrankin:
approved, but would be nice to have some tests to go with this

gomesalexandre:
> approved, but would be nice to have some tests to go with this

Yeah, since the whole module is fully untested I think it will make more sense to tackle all, or at least as many as possible for it, rather than adding just one! Created #313 for it

0xean:
This looks good to me, only question I have is how are we verifying that an ens address actually resolves?  
@gomesalexandre - thanks for creating the issue for additional tests!

0xean (approved):

gomesalexandre:
> This looks good to me, only question I have is how are we verifying that an ens address actually resolves? @gomesalexandre - thanks for creating the issue for additional tests!

We don't, with this implementation this is done by the client in https://github.com/shapeshift/web/pull/704.

An unresolvable address is still a valid address, from a spec point of view, but maybe it still makes sense for chain-adapters to handle such concerns

0xean:
Thanks @gomesalexandre - going to merge this!

PR #297, opened 2022-01-12
fix: return unconfirmed balances as part of account balance response

kaladinlight:


0xdef1cafe (approved):

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -115,8 +116,10 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
       const { chain, network } = caip2.fromCAIP2(caip)
       const { data } = await this.providers.http.getAccount({ pubkey: pubkey })
 
+      const balance = bnOrZero(data.balance).plus(bnOrZero(data.unconfirmedBalance))
+
       return {
-        balance: data.balance,
+        balance: balance.toString(),
```
asamere, line 122: is this always safe i.e. return as decimal? concern toString() will result in exponential notation with very large/precise numbers.

asamere: is balance an int (base units) here?

asamere: `balance.toFixed(n)`?

kaladinlight: Hmm... possible it defaults to exponential notation at large numbers. `toString()` being used heavily throughout whole stack to my knowledge, so we would probably need to address everywhere in another PR?

kaladinlight: looks like `toFixed()` is the trick. Think we need a chore PR to update throughout stack for sure.

asamere: Ok, thanks!

asamere (approved):

PR #314, opened 2022-01-18
Improve ChainAdapter typings

gomesalexandre:
This adds a `ChainSpecificAdapterProps<T, U, V>` type as proposed by @reidrankin in https://github.com/shapeshift/web/pull/704 and extends `ChainAdapter` with a single usage of it, for `ChainTypes.Ethereum`.

`ChainSpecificAdapterProps` is to be used as an intersection with `ChainAdapter` and allows to declare chain-specific properties for each chain.

Example usage: https://github.com/shapeshift/web/blob/65ea158bdba3623842eec4cb919bf3f650ae5bd7/src/components/Modals/Send/views/Address.tsx#L83


gomesalexandre:
Recreating directly from project branch now that I got authorization for it. CI will be able to properly ran this way.

PR #315, opened 2022-01-18
feat: improve ChainAdapter typings

gomesalexandre:
This adds a `ChainSpecificAdapterProps<T, U, V>` type as proposed by @reidrankin in shapeshift/web#704 and extends ChainAdapter with a single usage of it, for ChainTypes.Ethereum.

`ChainSpecificAdapterProps` is to be used as an intersection with ChainAdapter and allows to declare chain-specific properties for each chain.

Example usage: https://github.com/shapeshift/web/blob/65ea158bdba3623842eec4cb919bf3f650ae5bd7/src/components/Modals/Send/views/Address.tsx#L83

### DIFF
--- a/packages/chain-adapters/src/api.ts
+++ b/packages/chain-adapters/src/api.ts
@@ -1,7 +1,13 @@
 import { CAIP2 } from '@shapeshiftoss/caip'
 import { BIP44Params, chainAdapters, ChainTypes } from '@shapeshiftoss/types'
 
-export interface ChainAdapter<T extends ChainTypes> {
+type ChainSpecificAdapterProps<
+  T extends ChainTypes,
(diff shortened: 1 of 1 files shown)


0xdef1cafe:
as discussed, i think this is a much simpler/cleaner approach

https://canary.discord.com/channels/@me/929088734350168095/933783494818164846

@reidrankin i appreciate the aggressive typing here, but have concerns around maintainability. 

what do you think of the simpler `?.` approach?

0xdef1cafe:
@cjthompson thoughts?

cjthompson:
I agree with Reid. Stricter typing is preferred

0xdef1cafe (approved):

Issue #313, opened 2022-01-16
EthereumChainAdapter - Tests


gomesalexandre:
Tests should be added for all methods of `EthereumChainAdapter`, which is currently untested.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 3000.0 FOX (1241.75 USD @ $0.41/FOX)  attached to it as part of the ShapeShift fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/shapeshift/lib/313/100027486).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/shapeshift/lib/313/100027486).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $6,401,475.94 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gomesalexandre:
FYI @0xean this is current version of tests: https://github.com/shapeshift/lib/blob/main/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts. And the current run of tests for the module:

```shell
 PASS  packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
  EthereumChainAdapter
    getBalance
      ✓ is unimplemented (7 ms)
    getFeeData
      ✓ should return current ETH network fees (14 ms)

--------------------------|---------|----------|---------|---------|----------------------------------
File                      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
--------------------------|---------|----------|---------|---------|----------------------------------
All files                 |   33.08 |    17.43 |   17.39 |   32.69 |
 error                    |   44.44 |        0 |       0 |   44.44 |
  ErrorHandler.ts         |   44.44 |        0 |       0 |   44.44 | 6-13
 ethereum                 |   21.32 |    12.71 |   14.81 |   20.89 |
  EthereumChainAdapter.ts |   21.32 |    12.71 |   14.81 |   20.89 | 19-23,30-177,184,190-197,255-312
 utils                    |   45.16 |    26.02 |   22.22 |   45.29 |
  bignumber.ts            |      75 |       50 |   33.33 |   81.81 | 11-12
[...]

PR #319, opened 2022-01-20
feat(buildSendTransaction): throw on sendMax with 0 balance

gomesalexandre:
This was spotted while writing `EthereumChainAdapter` tests in #313.
When building a send tx and using the `sendMax` option, we currently throw on no ERC20 balance, but we don't in case of no native token balance.


packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
```diff
@@ -169,6 +169,8 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
           if (!erc20Balance) throw new Error('no balance')
           tx.value = erc20Balance
         } else {
+          if (new BigNumber(account.balance).isZero()) throw new Error('no balance')
```
pastaghost, line 172: Good thinking. We definitely should have been checking for this here.

0xean (approved):

PR #320, opened 2022-01-20
feat(EthereumChainAdapter): buildSendTransaction and broadcastTransaction tests

gomesalexandre:
#313

- add buildSendTransaction tests
- add broadcastTransaction tests

### DIFF
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
@@ -6,9 +6,15 @@
  */
 import { ETHWallet } from '@shapeshiftoss/hdwallet-core'
 import { NativeAdapterArgs, NativeHDWallet } from '@shapeshiftoss/hdwallet-native'
-import { chainAdapters } from '@shapeshiftoss/types'
+import { chainAdapters, ChainTypes } from '@shapeshiftoss/types'
+import { ethereum as unchainedEthereum } from '@shapeshiftoss/unchained-client'
+import { numberToHex } from 'web3-utils'
 
+import { bn } from '../utils/bignumber'
 import * as ethereum from './EthereumChainAdapter'
+
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
```diff
@@ -103,4 +144,298 @@ describe('EthereumChainAdapter', () => {
       )
     })
   })
+
+  describe('broadcastTransaction', () => {
```
gomesalexandre, line 148: Note to self: Might want to also test the unhappy path (https://github.com/shapeshift/unchained/blob/dae102cd30a99751a529ade6ad9ea249fe52cb5c/node/coinstacks/ethereum/api/src/controller.ts#L271).

pastaghost (approved):

0xean (approved):

0xean:
@gomesalexandre First off, thank you for the huge number of PRs - a tip is headed your way!

Second, are you able to resolve the conflict here and we can get this merged?

gomesalexandre:
@0xean Mergefix ✅

packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
```diff
@@ -55,11 +61,29 @@ jest.mock('axios', () => ({
...
+    gasLimit
+  }
+
+  const chainSpecificWithErc20ContractAddress = {
+    ...chainSpecificWithoutErc20ContractAddress,
+    erc20ContractAddress
+  }
+  const getInfoMockResponse = {
+    data: { network: 'mainnet' }
+  }
```
cjthompson, line 80: In order to avoid accidental mutations of these objects, these should be functions that return a copy of the object.

I pattern that I like to follow is:
```ts
const mockWhateverData = (obj: CorrectType) => _.merge({
  prop: 1,
  nested: {
    prop: 2
  }
}, obj)
```

this allows each test to call `mockWhateverData()` to get a clean copy of the mock or `mockWhateverData({ prop: false })` to change a specific property of the default mock to test something slightly differently.

PR #322, opened 2022-01-22
feat(EthereumChainAdapter): test throw on passed ENS 'to'

gomesalexandre:
#313
- buildSendTransaction: test throw on passed ENS 'to' in tx
- Immutably build `chainSpecific` mock with a `makeChainSpecific` builder which accepts an optional object of additional properties to be added to the base `chainSpecific` properties

### DIFF
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
@@ -8,6 +8,7 @@ import { ETHSignTx, ETHWallet } from '@shapeshiftoss/hdwallet-core'
 import { NativeAdapterArgs, NativeHDWallet } from '@shapeshiftoss/hdwallet-native'
 import { chainAdapters, ChainTypes } from '@shapeshiftoss/types'
 import { ethereum as unchainedEthereum } from '@shapeshiftoss/unchained-client'
+import { merge } from 'lodash'
 import { numberToHex } from 'web3-utils'
 
 import { bn } from '../utils/bignumber'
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
```diff
@@ -6,6 +6,10 @@
  */
 import * as ethereum from './EthereumChainAdapter'
 
+const EOA_ADDRESS = '0x05a1ff0a32bc24265bcb39499d0c5d9a6cb2011c'
+const CONTRACT_ADDRESS = '0xc770eefad204b5180df6a14ee197d99d808ee52d'
+const ENS_NAME = 'willywonka.eth'
```
pastaghost, line 11: The owner of this address has publicly associated his identity with this address and ENS name, but I think we should still try and use another ENS/address combination here if possible. How about vitalik.eth instead?

gomesalexandre: That's probably better yeah! I'll use `vitalik.eth` / `0xd8da6bf26964af9d7eed9e03e53415d37aa96045`

0xean (approved):

cjthompson (commented):
We shouldn't export a function just to be able to test it. Tests should test the public API.

In this case, the correct way to test this is to write a test for whatever internal function calls this function and verify the output of that function.

cjthompson (changes requested):
Please change the test to not require exporting an internal function

gomesalexandre:
> We shouldn't export a function just to be able to test it. Tests should test the public API.
> 
> In this case, the correct way to test this is to write a test for whatever internal function calls this function and verify the output of that function.

https://github.com/shapeshift/lib/commit/369ac4d675b178b498f893d8ecec46e034db8959 that has just been merged should test most of this then. I'll just add one more case for the ENS `to`.

cjthompson (approved):

gomesalexandre:
@0xean Mergefixed ✅

PR #335, opened 2022-01-28
feat: 821 extend yearn vaults

theoboldfrazier:


0xdef1cafe (approved):

0xdef1cafe:
@mrnerdhair can you please investigate security implications of `disableAllowList` as a priority? this PR (and corresponding web PR) is the only thing we need monday for juicy yearn vaults and DAO revenue.

mrnerdhair (approved):

mrnerdhair:
Looked into it this weekend; the allowlist option doesn't even do anything on mainnet yet, so this doesn't actually change anything. We'll need to revisit if/when they add that functionality in the SDK.

PR #341, opened 2022-02-05
Bug/fix yearn chart data

theoboldfrazier:


packages/market-service/src/yearn/yearn-vaults.ts
```diff
@@ -250,17 +250,29 @@ export class YearnVaultMarketCapService implements MarketService {
...
-            .div(`1e+${decimals}`)
-            .dp(6)
-            .toNumber()
+      return vaultDayData.reduce<HistoryData[]>((acc, current: VaultDayData) => {
+        const date = Number(current.timestamp)
+        const dateCheck = new Date(date).valueOf()
+        if (isNaN(dateCheck)) {
+          console.error('Yearn SDK vault has invalid date')
+          return acc
+        }
```
gomesalexandre, line 259: I was trying to come up with a better naming for this like `isValidDate` but actually that's a bit weird since the validity check happens on next line. What about something like this?
```suggestion
        if (!isValidDate(new Date(date).valueOf())) {
          console.error('Yearn SDK vault has invalid date')
          return acc
        }
```
where `isValidDate` would be a type guard like so:

`const isValidDate = (timestamp: number) => !isNan(timestamp)`

gomesalexandre (approved):

0xdef1cafe (approved):

0xdef1cafe:
looks good, haven't tested without merging lib but code wise it's fine

@theoboldfrazier does this actually fix https://github.com/shapeshift/web/issues/920

0xdef1cafe (approved):

theoboldfrazier:
@0xdef1cafe I think it fixes the underlying reason for that issue https://github.com/shapeshift/web/issues/920.

The issue wasn't the granularity, rather the bad use of the date type which caused issues because the chart was not able to interpret the strings I was sending back from yearn price history.

packages/market-service/src/yearn/yearn-vaults.ts
theoboldfrazier: I like that. I wrote a util function for it.

PR #344, opened 2022-02-05
feat: expired prop on yearn vaults

amitojsingh366:
adds the `expired` prop on yearn vaults

### DIFF
--- a/packages/investor-yearn/src/api/vaults.ts
+++ b/packages/investor-yearn/src/api/vaults.ts
@@ -28,7 +28,7 @@ export const getSupportedVaults = async (): Promise<SupportedYearnVault[]> => {
       chain: ChainTypes.Ethereum,
       provider: DefiProvider.Yearn,
       type: DefiType.Vault,
-      expired: !!vault.metadata.depositsDisabled || bnOrZero(vault.metadata.depositLimit).lte(0)
+      expired: vault.metadata.depositsDisabled || bnOrZero(vault.metadata.depositLimit).lte(0)
     }
   })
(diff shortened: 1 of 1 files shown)


packages/investor-yearn/src/api/vaults.ts
```diff
@@ -14,27 +14,21 @@ export type SupportedYearnVault = {
...
+  return vaults.map((vault: Vault) => {
+    return {
+      vaultAddress: toLower(vault.address),
+      name: `${vault.name} ${vault.version}`,
+      symbol: vault.symbol,
+      tokenAddress: toLower(vault.token),
+      chain: ChainTypes.Ethereum,
+      provider: DefiProvider.Yearn,
+      type: DefiType.Vault,
+      expired: !!vault.metadata.depositsDisabled || bnOrZero(vault.metadata.depositLimit).lte(0)
```
0xApotheosis, line 31: Do we need to coerce `vault.metadata.depositsDisabled`? It's already of type `boolean | undefined`.

0xApotheosis (changes requested):

packages/investor-yearn/src/api/vaults.ts
amitojsingh366: fixed, I was actually coercing it before as I was only using `vault.metadata.depositsDisabled` then alter on I also added the check for `vault.metadata.depositLimit`

0xApotheosis (approved):

0xdef1cafe (approved):

PR #346, opened 2022-02-06
feat: add transformVault util and change return types

amitojsingh366:

PR #347, opened 2022-02-06
refactor: fix windows builds

amitojsingh366:
this PR fixes @shapeshiftoss/caip builds on windows
this PR should be merged after #346 as I did an oopsie *(it contains changes from both PRs)*
closes https://github.com/shapeshift/web/issues/947


0xdef1cafe:
the fix here is not to change path naming conventions to something less correct, and update the docs to direct people to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/install). we shouldn't be changing code to work with backwards file systems.

amitojsingh366:
> the fix here is not to change path naming conventions to something less correct, and update the docs to direct people to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/install). we shouldn't be changing code to work with backwards file systems.

but we won't be able to build the electron app for keepkey over CI then

BitHighlander:
if it's the goal of CAIP to become a standard lib in the space using special characters in the pathing is problematic. 74 percent of the world is on windows. using WSL might help a developer by using ubuntu under the hood, but that does not fix this CAIP package from throwing errors on any software it's bundled with running under windows. I understand that the change has 0 benefits for webs developer flow. but this is purely an aesthetics change of a : vr _. this would greatly help us to not have to maintain a fork of the entire lib repo just to change a special character. and any other projects that are not web-based will be able to use this package in the space as well.

amitojsingh366:
> if it's the goal of CAIP to become a standard lib in the space using special characters in the pathing is problematic. 74 percent of the world is on windows. using WSL might help a developer by using ubuntu under the hood, but that does not fix this CAIP package from throwing errors on any software it's bundled with running under windows. I understand that the change has 0 benefits for webs developer flow. and this is purely an aesthetics change of a : vr _. however this would greatly help us to not have to maintain a fork of the entire lib repo just to change a special character. and any other projects that are not web-based will be able to use this package and not worry about throwing errors after this modification.

this was very properly phrased and I totally agree with this, @0xdef1cafe please give this another review

0xdef1cafe:
@BitHighlander if you can confirm this is working locally after a yarn clean build and link into web i'll approve

BitHighlander:
@0xdef1cafe testing locally. working good :shipit:

0xdef1cafe (approved):

pastaghost (approved):

Issue #353, opened 2022-02-10
Asset service blacklist


0xdef1cafe:
# Overview

Currently the app will display any asset with a balance in your portfolio if it is on the asset list, and soon even assets not on the asset will be displayed too. There are a handful of known scam tokens that have been airdropped to many users' wallets and display misleading fiat values (ie. KICK and Minereum). As app maintainers, we want to be able to prevent known scam tokens from appearing in users' wallets and portfolios. Contributors should be able to make pull requests to an open source repo to add new known scam tokens to the blacklist.

# Reference

1. The following assets no longer display in v2, even if users have a balance in their wallet:
* KICK (0x426CA1eA2406c07d75Db9585F22781c096e3d0E0)
* Minereum (0x426CA1eA2406c07d75Db9585F22781c096e3d0E0)
* UNIH (0x4bf5dc91e2555449293d7824028eb8fe5879b689) 
* 0x7761c09EA2DE3F7205Dba784Fd87Fd196E019D2A
* 0x6f0371247d21d7afad6e624f6e8dbb3bff0455c2
* 0xF9d25EB4C75ed744596392cf89074aFaA43614a8
* 0x0f8b6440a1f7be3354fe072638a5c0f500b044be
* 0x1412ECa9dc7daEf60451e3155bB8Dbf9DA349933
* 0x34278F6f40079eae344cbaC61a764Bcf85AfC949

# Acceptance criteria

* `blacklist.json` file added somewhere in the asset service with a list of blacklisted asset ids, [using our asset id implementation](https://github.com/shapeshift/lib/blob/main/packages/caip/src/caip19/caip19.test.ts#L43) in the format `["eip155:1/erc20:0x426ca1ea2406c07d75db9585f22781c096e3d0e0", "eip155:1/erc20:0x4bf5dc91e2555449293d7824028eb8fe5879b689"]`
* your implementation should filter out blacklisted tokens [somewhere in this function or another appropriate place](https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/generateAssetData/generateAssetData.ts#L20)
* your implementation should be cross-chain and not ethereum specific
* add all the tokens listed above to the blacklist
* asset ids in the blacklist should be lowercase and not checksum format
* run the `yarn generate` script to regenerate asset data, verifying blacklisted tokens do not appear in the `generatedAssetData.json` file, commit the updated file
* add documentation to the repo how to add blacklisted tokens to the list
[...]

PR #355, opened 2022-02-10
feat: asset blacklist functionality

NeOMakinG:
## How to test?

I tested with `0x426ca1ea2406c07d75db9585f22781c096e3d0e0` which should be inside `generatedAssetData.json`, it's not inside it anymore

Also tested with `ZRX` inside my own wallet, it doesn't show it anymore inside the `web` app if I blacklist it.

closes https://github.com/shapeshift/lib/issues/353


packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -29,25 +32,35 @@ const generateAssetData = async () => {
...
           ...yearnVaults,
           ...ironBankTokens,
           ...zapperTokens,
           ...underlyingTokens
         ]
-        const uniqueTokens = uniqBy(tokens, 'caip19') // Remove dups
+
+        let uniqueTokens = uniqBy(tokens, 'caip19') // Remove dups
+
+        // Remove blacklisted tokens
```
0xdef1cafe, line 46: good approach, but can you please do this outside the `if` block on line 26, such that this blacklist is not ethereum specific, and avoids the duplication on line 85?

somewhere on line 93 once we have the final data before we right it out would be a good place to do this filtering.

this will become a problem in the future with scam tokens on other chains, our asset ids are already chain-agnostic, and we want this solution to be chain agnostic.

0xdef1cafe (changes requested):
great start, loves the docs you added and covered the existing assets with manual descriptions too.

small requested change to make things more DRY and chain agnostic but otherwise spot on.

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -58,15 +71,22 @@ const generateAssetData = async () => {
...
         }
+
         return baseAssetWithTokens
       } else {
-        return baseAsset
+        if (!blacklistedAssets.includes(baseAsset.caip19)) {
+          return baseAsset
+        }
+
+        return
```
gomesalexandre, line 89: unhappy path first 🙏🏽 
```suggestion
        if (blacklistedAssets.includes(baseAsset.caip19)) { return }

        return baseAsset
```

NeOMakinG: Will be removed after the first review is completed anyway

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -17,8 +18,22 @@ import {
   getZapperTokens
 } from './ethTokens/extendErc20'
 
+const blacklistedAssets: Array<string> = blacklist
+
+const filterBlacklistedAssets = <T extends BaseAsset | TokenAsset>(assets: T[]) => {
```
0xdef1cafe, line 23: much much nicer, ty 🙏 as a bonus, you could add a unit test for this function if you like

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -58,19 +80,24 @@ const generateAssetData = async () => {
...
         }
+
         return baseAssetWithTokens
       } else {
         return baseAsset
       }
     })
   )
 
+  generatedAssetData = filterBlacklistedAssets(generatedAssetData)
```
0xdef1cafe, line 99: one last requested change, please assign to a new variable rather than mutating here. `filterBlacklistedAssets` should be pure, take all the assets and return the filtered subset, and not mutate it's inputs.

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -4,11 +4,12 @@ import { caip19 } from '@shapeshiftoss/caip'
 import { BaseAsset, ChainTypes, NetworkTypes, TokenAsset } from '@shapeshiftoss/types'
 import axios from 'axios'
 import fs from 'fs'
-import { chunk } from 'lodash'
+import { chunk, filter } from 'lodash'
```
0xdef1cafe, line 7:
```suggestion
import chunk from 'lodash/chunk'
import filter from 'lodash/filter'
```

0xdef1cafe (changes requested):

packages/asset-service/src/generateAssetData/generateAssetData.ts
NeOMakinG: Ok, I will do it tomorrow, beginning to be late here in France :)

packages/asset-service/src/generateAssetData/generateAssetData.ts
NeOMakinG: You're right!

packages/asset-service/src/generateAssetData/generateAssetData.ts
NeOMakinG: I think this unit test doesn't belong to AssetService, I'll create a new file introducing unit tests to utils functions, right?

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -17,6 +19,20 @@ import {
   getZapperTokens
 } from './ethTokens/extendErc20'
 
+const blacklistedAssets: Array<string> = blacklist
+
+export const filterBlacklistedAssets = <T extends BaseAsset | TokenAsset>(assets: T[]) => {
+  const filteredAssets = filter(assets, (token) => {
+    if ('tokens' in token && token.tokens) {
+      token.tokens = filterBlacklistedAssets(token.tokens)
+    }
```
gomesalexandre, line 28: A boolean isn't iterable, and the same goes for `undefined`, there would be a runtime error in trying to iterate over false (short-circuit) or `undefined` token.tokens.

Now I get it that this wouldn't happen at runtime but in this case, it means that the types are lying. We shouldn't rely on a `99.9% of the time, it works every time` to write logic, there's always that 0.01%.

The reason why this works is that the `operator in iterable` is making a an implicit discrimination depending on the operator, so why not make it explicit so we can avoid this suboptimal syntax? 

What about:

```suggestion
  const isBaseAsset = (asset: BaseAsset | TokenAsset): asset is BaseAsset =>
    Boolean((asset as BaseAsset)?.tokens)

  const filteredAssets = filter(assets, (token) => {
    if (isBaseAsset(token) && token.tokens) {
      token.tokens = filterBlacklistedAssets(token.tokens)
    }
```

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  explorerTxLink: 'https://live.blockcypher.com/btc/tx/',
+  sendSupport: false,
+  receiveSupport: false
+}
+
+const assetList = [EthAsset, BtcAsset]
+
+describe('Utils', () => {
+  describe('filterBlacklistedAssets', () => {
+    it('should filter eth from the asset list', () => {
```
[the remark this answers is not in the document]
gomesalexandre: `ETH` for the OCD peep around here though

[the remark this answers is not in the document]
NeOMakinG: Doesn't mater to me honestly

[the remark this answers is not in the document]
NeOMakinG: We can test with both if you want, adding 2-3 assets and see if they are all removed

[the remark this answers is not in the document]
NeOMakinG: Added MNE

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const assetList = [EthAsset, BtcAsset, MNEAsset]
+
+describe('Utils', () => {
+  describe('filterBlacklistedAssets', () => {
+    it('should filter eth from the asset list', () => {
+      const filteredAssetList = filterBlacklistedAssets(assetList)
+      expect(filteredAssetList).toHaveLength(1)
+      const ethFiltered = <BaseAsset>filteredAssetList[0]
```
gomesalexandre: Nitpick, `as` is usually used for casting

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  secondaryColor: '#FFFFFF',
+  icon: 'https://assets.coincap.io/assets/icons/256/btc.png',
+  explorer: 'https://live.blockcypher.com',
+  explorerAddressLink: 'https://live.blockcypher.com/btc/address/',
+  explorerTxLink: 'https://live.blockcypher.com/btc/tx/',
+  sendSupport: false,
+  receiveSupport: false
+}
+
+const MNEAsset: TokenAsset = {
```
gomesalexandre, line 41: It doesn't matter which token it is, that's an implementation detail, from a high-level perspective the asset is like any ERC20 and that's what's being tested
```suggestion
const ERC20Asset: TokenAsset = {
```

gomesalexandre (commented):
Since you're casting, tests can probably also be cleaned by removing most fields in your `BaseAsset` and `TokenAsset`s. I'd keep `caip19` (obviously) and possibly `name` and `symbol` for quickly identifying what the asset is, even though the variable name should be enough for that.

NeOMakinG:
@0xdef1cafe looks like there is a problem wth the [uniswap API](https://tokens.coingecko.com/uniswap/all.json), sometimes tokens doesn't have any address, it makes your toCAIP19 method crash, should I fix it inside this PR?

inside `ethTokens/index.ts` , `getTokens` should probably use a flatMap and return nothing if the `token.address` is undefined, or if you've a better solution...

0xdef1cafe:
> 

cool, other PR looks good and fixes coingecko rugging us, please just make the changes in this branch as they're rather related and we can review all at once, i closed the other PR.

NeOMakinG:
> > 
> 
> cool, other PR looks good and fixes coingecko rugging us, please just make the changes in this branch as they're rather related and we can review all at once, i closed the other PR.

Ok, pushed the fix + new version of generatedAssetData, please let me some minutes in order to verify it using the `web` app, because I don't know why generatedAssetData has a diff, probably because API updated or something like that

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,42 @@
...
+jest.mock(
+  './blacklist.json',
+  () => [
+    'bip122:000000000019d6689c085ae165831e93/slip44:0',
+    'eip155:1/erc20:0x426ca1ea2406c07d75db9585f22781c096e3d0e0'
+  ],
+  { virtual: true }
+)
+
+const BtcAsset = {
```
0xdef1cafe, line 14: please use mock asset data that is correctly typed from https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/service/AssetServiceTestData.ts#L140

separate them out into named assets, and recombine them into the same `mockAssets` array.

packages/asset-service/src/generateAssetData/baseAssets.ts
```diff
@@ -6,7 +6,7 @@ import {
   NetworkTypes
 } from '@shapeshiftoss/types'
 
-export const baseAssets: Array<BaseAsset> = [
+export const baseAssets: BaseAsset[] = [
```
0xdef1cafe, line 9: unnecessary change please revert

packages/asset-service/src/generateAssetData/ethTokens/index.ts
```diff
@@ -37,7 +37,7 @@ export async function getTokens(): Promise<TokenAsset[]> {
   const network = NetworkTypes.MAINNET
   const contractType = ContractTypes.ERC20
 
-  const tokens = uniswapTokenData.tokens.map((token) => {
+  const tokens = uniswapTokenData.tokens.flatMap((token) => {
```
0xdef1cafe, line 40:
```suggestion
  const tokens: TokenAsset[] = uniswapTokenData.tokens.flatMap((token) => {
```

to ensure this `flatMap` change doesn't have unintended consequences

packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -58,22 +84,27 @@ const generateAssetData = async () => {
...
         }
+
         return baseAssetWithTokens
       } else {
         return baseAsset
       }
     })
   )
 
+  const filteredAssetData = filterBlacklistedAssets(generatedAssetData)
```
0xdef1cafe, line 103:
```suggestion
  const generatedAssetData = filterBlacklistedAssets(unfilteredAssetData)
```

0xdef1cafe (changes requested):

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,33 @@
+import { BaseAsset } from '@shapeshiftoss/types'
+
+import { BTCMockedAsset, ETHMockedAsset } from '../service/AssetServiceTestData'
+import blacklist from './blacklist.json'
+import { filterBlacklistedAssets } from './utils'
+
+const blacklistedAssets: string[] = blacklist
+
+jest.mock('./blacklist.json', () => [ETHMockedAsset.tokens![0].caip19, BTCMockedAsset.caip19], {
```
NeOMakinG, line 9: @gomesalexandre @0xdef1cafe I'm really not sure about `ETHMockedAsset.tokens![0].caip19`

It's some mocked data, we know that `tokens` property exist, but TS throw an error

If it doesn't exist, jest throw an error, so I guess it should be fine, but if you have a better way...

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  describe('filterBlacklistedAssets', () => {
+    it('should filter BTC from the asset list', () => {
+      const filteredAssetList = filterBlacklistedAssets(blacklistedAssets, assetList)
+      expect(filteredAssetList[0]).toHaveProperty('caip19', ETHMockedAsset.caip19)
+    })
+
+    it('should filter ERC20 from the asset list', () => {
+      const filteredAssetList = filterBlacklistedAssets(blacklistedAssets, assetList)
+      const ethFiltered = filteredAssetList[0]
+      const blacklistedToken = ETHMockedAsset.tokens![1]
```
NeOMakinG, line 28: I have the same problem here, we know that tokens exist, and it's not related to the test goal at all

packages/asset-service/src/generateAssetData/Utils.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const blacklistedAssets: string[] = blacklist
+
+jest.mock('./blacklist.json', () => [ETHMockedAsset.tokens![0].caip19, BTCMockedAsset.caip19], {
+  virtual: true
+})
+
+const assetList: BaseAsset[] = [
+  Object.assign({}, ETHMockedAsset),
+  Object.assign({}, BTCMockedAsset)
```
NeOMakinG, line 15: Needed to clone these objects, overwise the test script mutate it and we're not able to test the final caip19 item

packages/asset-service/src/generateAssetData/Utils.test.ts
NeOMakinG: fixed using `as const` in mocks

packages/asset-service/src/generateAssetData/Utils.test.ts
NeOMakinG: fixed using as const in mocks

0xdef1cafe (changes requested):

packages/asset-service/src/service/AssetServiceTestData.ts
```diff
@@ -277,7 +281,7 @@ export const mockAssets: Asset[] = [
     sendSupport: false,
     receiveSupport: false
   }
-]
+] as BaseAsset[]
```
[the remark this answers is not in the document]
NeOMakinG: Ok, but the `as const` was here in order to avoir to cast all these `.tokens` access here: 
https://github.com/shapeshift/lib/pull/355/files#diff-0d5ad083ac650d73eb92d451dc27cf4e8cc54519da8394fd2016d6b2f7a170fbR7-R28

With these types, we can't anymore, so I'm forced to force this access, what's the good practice then? 😢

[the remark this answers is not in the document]
0xdef1cafe: `as const` is very different to `as SomeType`. `as const` says this won't change at runtime and the compiler can infer the type statically. `as SomeType` says "i know better than the compiler this is definitely `SomeType`". it needs to be use very judiciously and in general it's a code smell.

i'd rather do the `asset.tokens!` non-null assertion in a test under a specific code path, with a comment as to why, rather than asserting some test data matches a type that it may not, especially as tests change in the future.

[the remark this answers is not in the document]
NeOMakinG: Ok, understood, fixed it anyway so, it's reviewable now! Thanks for the feedback again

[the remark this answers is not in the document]
NeOMakinG: If you want me to comment it, I will do it later

0xean:
@0xdef1cafe - bump

packages/asset-service/README.md
```diff
@@ -10,3 +10,11 @@ It does not provide asset market data (price, volume, etc) or wallet balance inf
 2. Run `yarn generate` to build the `generatedAssetData.json` used by the the service to provide asset data.
 3. Commit and push changes to `generatedAssetData.json` to github
 4. Use assetService to lookup assets as needed.
+
+## Blacklist usage
+
+You can contribute to blacklist a flagged token by adding his [caip19](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md) to the [blacklist.json](https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/generateAssetData/blacklist.json) list.
```
gomesalexandre, line 16:
```suggestion
You can contribute to blacklist a flagged token by adding its [caip19](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md) to the [blacklist.json](https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/generateAssetData/blacklist.json) list.
```

gomesalexandre (approved):
🍬 looks good, typings are way cleaner now by not trying to get around the non-null assertion operator.

theoboldfrazier (approved):
This looks good!

0xdef1cafe:
hey @NeOMakinG - we're going to consider this work done from your side and pay out the bounty.

for security reasons we run the generate asset data script from our side now, and we merged another PR that will cause you conflicts, so i'm closing this PR and reopening here to get it merged

https://github.com/shapeshift/lib/pull/402

thanks for the great work and patience with this :)

PR #357, opened 2022-02-11
feat: add isTrusted to description of asset service

DaoDev44:
Add `isTrusted` boolean to asset service description function in order to determine the trusted (hard coded in the `descriptions.json` file) and non trusted (3rd party) descriptions.

### DIFF
--- a/packages/asset-service/src/service/AssetService.test.ts
+++ b/packages/asset-service/src/service/AssetService.test.ts
@@ -145,9 +145,10 @@ describe('AssetService', () => {
     it('should return the overridden description if it exists', async () => {
       const assetService = new AssetService(assetFileUrl)
 
-      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual(
-        'overriden description'
-      )
+      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual({
+        description: 'overriden description',
+        isTrusted: true
+      })
     })
 
     it('should return a string if found', async () => {
@@ -157,7 +158,10 @@ describe('AssetService', () => {
       const assetService = new AssetService(assetFileUrl)
       const description = { en: 'a blue fox' }
       mockedAxios.get.mockResolvedValue({ data: { description } })
-      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual(description.en)
+      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual({
+        description: description.en,
+        isTrusted: false
+      })
     })
 
     it('should throw if not found', async () => {
--- a/packages/asset-service/src/service/AssetService.test.ts
+++ b/packages/asset-service/src/service/AssetService.test.ts
@@ -38,7 +38,7 @@ const EthAsset: Asset = {
 jest.mock(
   './descriptions.json',
   () => ({
-    'eip155:3/slip44:60': 'overriden description'
+    'eip155:3/slip44:60': 'overridden description'
   }),
   { virtual: true }
 )
@@ -146,7 +146,7 @@ describe('AssetService', () => {
       const assetService = new AssetService(assetFileUrl)
 
       await expect(assetService.description({ asset: EthAsset })).resolves.toEqual({
-        description: 'overriden description',
+        description: 'overridden description',
         isTrusted: true
       })
     })
@@ -159,8 +159,7 @@ describe('AssetService', () => {
       const description = { en: 'a blue fox' }
       mockedAxios.get.mockResolvedValue({ data: { description } })
       await expect(assetService.description({ asset: EthAsset })).resolves.toEqual({
-        description: description.en,
-        isTrusted: false
+        description: description.en
       })
     })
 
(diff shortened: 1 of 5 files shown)


packages/asset-service/src/service/AssetService.ts
```diff
@@ -114,13 +119,24 @@ export class AssetService {
...
-    if (descriptions[asset.caip19]) return descriptions[asset.caip19]
+    // Return overriden asset description if it exists and add isTrusted for description links
+    if (descriptions[asset.caip19]) {
+      return {
+        description: descriptions[asset.caip19],
+        isTrusted: true
+      }
+    }
+
+    if (asset.dataSource !== AssetDataSource.CoinGecko) {
```
0xdef1cafe, line 133: we *only* want to trust the `descriptions.json` file, everything else is still untrusted

0xdef1cafe (changes requested):

packages/asset-service/src/service/AssetService.ts
DaoDev44: Totally agree. The only description object that is returned with `isTrusted` as true is when we find a description for a specific caip19 on line 126.

packages/asset-service/src/service/AssetService.test.ts
```diff
@@ -145,9 +145,10 @@ describe('AssetService', () => {
     it('should return the overridden description if it exists', async () => {
       const assetService = new AssetService(assetFileUrl)
 
-      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual(
-        'overriden description'
-      )
+      await expect(assetService.description({ asset: EthAsset })).resolves.toEqual({
+        description: 'overriden description',
```
mrnerdhair, line 149: also needs to be fixed on line 41: https://github.com/shapeshift/lib/blob/3c88bd562a96b29409673bf3a7e760ec4582f725/packages/asset-service/src/service/AssetService.test.ts#L41
```suggestion
        description: 'overridden description',
```

packages/asset-service/src/service/AssetService.ts
```diff
@@ -114,13 +119,24 @@ export class AssetService {
     return result
   }
 
-  async description({ asset }: { asset: Asset }): Promise<string> {
+  async description({ asset }: { asset: Asset }): Promise<DescriptionData> {
     const descriptions: Record<string, string> = assetsDescriptions
 
-    // Return overriden asset description if it exists
-    if (descriptions[asset.caip19]) return descriptions[asset.caip19]
+    // Return overriden asset description if it exists and add isTrusted for description links
```
mrnerdhair, line 125:
```suggestion
    // Return overridden asset description if it exists and add isTrusted for description links
```

packages/asset-service/src/service/AssetService.ts
```diff
@@ -47,6 +47,11 @@ type ByTokenIdArgs = {
   tokenId?: string
 }
 
+type DescriptionData = {
+  description: string
+  isTrusted: boolean
```
mrnerdhair, line 52: seems like it might be convenient to be able to omit this field in literals; that said, if so, allowing both `false` and `undefined` would be denormalized.

```suggestion
  isTrusted?: true
```

packages/asset-service/src/service/AssetService.ts
```diff
@@ -47,6 +47,11 @@ type ByTokenIdArgs = {
...
+}
```
mrnerdhair, line 53: I know this is super-flipping-paranoid, but maybe it'd be nice to only return frozen objects from `description()`. IDK what kind of exploit code might try setting `isTrusted = true` on an existing description but I'd sleep better at night if it wasn't even something to worry about.

```suggestion
type DescriptionData = Readonly<{
  description: string
  isTrusted: boolean
}>
```

0xdef1cafe: this is a really good point and should be included

packages/asset-service/src/service/AssetService.ts
0xdef1cafe: i like this too

DaoDev44: Done

packages/asset-service/src/service/AssetService.ts
DaoDev44: Done

0xdef1cafe (approved):

packages/types/src/base.ts
```diff
@@ -51,6 +51,7 @@ type AbstractAsset = {
   caip2: string
   chain: ChainTypes
   description?: string
+  isTrustedDescription?: boolean | undefined
```
0xdef1cafe, line 54:
```suggestion
  isTrustedDescription?: boolean
```

`boolean | undefined` means if it is defined, it can also be undefined

0xdef1cafe (changes requested):

0xdef1cafe (approved):

theoboldfrazier (approved):

PR #362, opened 2022-02-11
fix: verify than token address is not empty inside ethTokens

NeOMakinG:
Currently, the CI crash because there is a token with an empty address on coingecko

# How to test?

Verify that the CI is ✅


0xdef1cafe:
hey @NeOMakinG can you please just make this change in your blacklist PR?

PR #370, opened 2022-02-12
feat: add Cosmos SDK ChainType and NetworkType

gomesalexandre:
- add Cosmos SDK ChainType in enum
- add CosmosHub (Gaia) Vega testnet NetworkType in enum

### DIFF
--- a/packages/types/src/base.ts
+++ b/packages/types/src/base.ts
@@ -21,14 +21,16 @@ export enum ContractTypes {
 
 export enum ChainTypes {
   Ethereum = 'ethereum',
-  Bitcoin = 'bitcoin'
+  Bitcoin = 'bitcoin',
+  CosmosSDK = 'cosmosSdk'
 }
 
 export enum NetworkTypes {
   MAINNET = 'MAINNET',
   TESTNET = 'TESTNET', // BTC, LTC, etc...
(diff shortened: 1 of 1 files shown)


0xdef1cafe (approved):

packages/types/src/base.ts
```diff
@@ -21,14 +21,19 @@ export enum ContractTypes {
 
 export enum ChainTypes {
   Ethereum = 'ethereum',
-  Bitcoin = 'bitcoin'
+  Bitcoin = 'bitcoin',
+  CosmosSDK = 'cosmosSdk'
```
cjthompson, line 25:
```suggestion
  Cosmos = 'cosmos'
```

Why did you choose to put `SDK` in the chain name?

cjthompson:
Can you add a reference to the issue in the description?

cjthompson:
I do not like the idea of merging PRs that result in broken builds. Even though these are separate tickets, I'd prefer a single PR that builds correctly over 2 PRs are do not.

gomesalexandre:
> I do not like the idea of merging PRs that result in broken builds. Even though these are separate tickets, I'd prefer a single PR that builds correctly over 2 PRs are do not.

Will move it over to #371 in this case.

PR #379, opened 2022-02-15
feat: add "byChainId" function to ChainAdapter Manager

cjthompson:
This is a step toward not using the ChainTypes enum by using a CAIP2 string to get the ChainAdapter instance.

`byChainId` takes a `CAIP2` value and returns a ChainAdapter instance or throws an error.

This PR is not associated with an existing issue.


0xApotheosis (approved):
LGTM!

theoboldfrazier (approved):

Issue #354, opened 2022-02-10
Asset Service: Generate icons for assets without an icon.


theoboldfrazier:
## Background

We want to display icons for assets that do not have a remote icon resource. In the asset-service package, we go out to third parties (coingecko, coincap, etc.) to collect a list of assets and store those assets in a json file within the package. During this asset generation, we store a URL for an asset's icon, then on display, we reach out to that remote resource to download the icon into the app.

For assets without a remote resource, we want to be able to generate an icon for display with that asset. We can use identicon, or gravatar or a library to generate the icon.

## Acceptance Criteria:

1) Add a module to asset-service to deterministically generate an icon based on the caip19 of an asset.

2) During rendering in web, if the asset does not have a remote resource icon (property icon on the Asset type), use the new module to generate the icon on the fly and display that icon.

3) The display of the icon should include up to 5 characters of the asset's symbol.

### Example assets currently in the app that do not have an icon.

https://etherscan.io/token/0x30ceCB5461A449A90081F5a5F55db4e048397BAB
https://etherscan.io/token/0xd51e852630debc24e9e1041a03d80a0107f8ef0c
https://etherscan.io/token/0x90528aeb3a2b736b780fd1b6c478bb7e1d643170

### Bounty Hunters

- [Join our Discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info

NeOMakinG:
Do we really need a module for this?

It could be a React component with a simple SVG, and an url props, if the url is set => display a simple image, else => use 5 first letter of the asset as a prop inside the SVG

If you feel I'm right, this issue can be moved to the `web` repository

NeOMakinG:
And of course, I would love to work on this issue :D

amitojsingh366:
I would like to work on this, the original issue for this *(https://github.com/shapeshift/web/issues/959)* was also assigned to me

theoboldfrazier:
We would like to have the icon generation module in lib/asset-service so that it can be reused outside of web. There will be web code needed to implement the module.

amitojsingh366:
> We would like to have the icon generation module in lib/asset-service so that it can be reused outside of web. There will be web code needed to implement the module.

@theoboldfrazier we can do this using something like canvas then, I can implement that

[...]

PR #385, opened 2022-02-17
feat: generate identicon for assets without logo

Soptq:
This PR generates deterministic identicon for assets that does not have a proper icon.

Related issue: #354


packages/asset-service/src/service/AssetService.ts
```diff
@@ -12,15 +18,45 @@ export const flattenAssetData = (assetData: BaseAsset[]): Asset[] => {
...
           ...tokenAsset,
           chain: baseAsset.chain,
           network: baseAsset.network,
           slip44: baseAsset.slip44,
           explorer: baseAsset.explorer,
           explorerAddressLink: baseAsset.explorerAddressLink,
           explorerTxLink: baseAsset.explorerTxLink
-        })
+        }
+        if (typeof flatAssetDatum.icon === 'undefined') {
```
gomesalexandre, line 30: Is there any reason this doesn't work?
```suggestion
        if (!flatAssetDatum.icon) {
```

packages/asset-service/src/service/AssetService.ts
```diff
@@ -1,6 +1,12 @@
+import { caip19 } from '@shapeshiftoss/caip'
 import { Asset, AssetDataSource, BaseAsset, ChainTypes, NetworkTypes } from '@shapeshiftoss/types'
 import axios from 'axios'
 
+/* eslint-disable */
+const Identicon = require('identicon.js')
+const { fonts, renderPixels } = require('js-pixel-fonts')
```
gomesalexandre, line 7: If possible, it would be better to use ESM. I've checked and none of these libraries support it so we're effectively force to use require statements.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -12,15 +18,45 @@ export const flattenAssetData = (assetData: BaseAsset[]): Asset[] => {
...
+          const identiconImage = new Identicon(tokenId?.substring(2), options).render()
+          // render pixel symbol
+          for (let x = 0; x < imageSize; x++) {
+            const offsettedX = Math.floor((x - offsetX) / symbolScale)
+            if (offsettedX < 0 || offsettedX >= 5) continue
+            for (let y = 0; y < imageSize; y++) {
+              const offsettedY = Math.floor((y - offsetY) / symbolScale)
+              if (offsettedY < 0 || offsettedY >= 7) continue
+              if (symbolPixels[offsettedY][offsettedX]) {
+                identiconImage.buffer[identiconImage.index(x, y)] = identiconImage.color.apply(
```
gomesalexandre, line 50: Isn't there any way to make this work without fiddling with the internals of identicon's `pngLib`?
This is not documented and will make it harder to maintain.

packages/asset-service/src/service/AssetService.ts
Soptq: Thanks for your suggestion!

packages/asset-service/src/service/AssetService.ts
Soptq: Yes, both of these libraries are not supporting ESM so we currently can only require them.

Soptq:
Hi @gomesalexandre, the latest commit adds extensive comments to the identicon rendering process for future maintenance ;)

gomesalexandre:
Thanks for the PR @Soptq 🎉 ! Appreciate the work that went into researching how to generate identicons with text.
I'll have a more in-depth look at this later today and test it then.

cjthompson:
We did this in our legacy code by just using a URL for the image like this:

see `fallbackSrc`.

This solution doesn't require any additional code or adding dependencies.

```
const AssetIcon = ({ iconUrl, symbol = 'BTC', size = '24px', ...props }) => {
  return (
    <Box pos='relative' display='flex' alignItems='center'>
      <Image
        key={symbol}
        src={iconUrl}
        fallbackSrc={`https://identicon-api.herokuapp.com/${symbol}/100?format=png`}
        {...props}
      />
    </Box>
  )
}
```

Soptq:
Hi @cjthompson , that is true and actually in my first implementation I do similarly things, but there are 3 reasons that I think I need to change to current approach:
1. According to the motivation of the related issue, the generating of identicon is required to be reused in outside of web.
2. The fallback approach can't render some additional letters above the icon, which is also a requirement of the related issue.
3. The fallback approach might cause additional HTTP connection, while local rendering does not.

packages/asset-service/src/service/AssetService.ts
```diff
@@ -1,9 +1,64 @@
...
+  // In order to additionally render text, we acquire PNGLib object here which has some convenient png operator.
+  // https://github.com/stewartlord/identicon.js/blob/master/pnglib.js
+  const identiconImage = new Identicon(tokenId?.substring(2), options).render()
+
+  // we just return if no symbol
+  if (symbol.length === 0) return `data:image/png;base64,${identiconImage.getBase64()}`
+
+  // here we start to render the text
+  // we only take the first letter from the symbol to render
+  const symbolPixels = renderPixels(symbol[0], fonts.sevenPlus)
```
NeOMakinG, line 33: In the issue, theobold says that `The display of the icon should include up to 5 characters of the asset's symbol.`

Did I missed something then?

packages/asset-service/src/service/AssetService.ts
```diff
@@ -1,9 +1,64 @@
...
+const Identicon = require('identicon.js')
+const { fonts, renderPixels } = require('js-pixel-fonts')
+/* eslint-enable */
+
 import assetsDescriptions from './descriptions.json'
 import localAssetData from './generatedAssetData.json'
 
+const getRenderedIdenticonB64 = (symbol: string, tokenId: string): string => {
+  // `imageSize` defines the size (px) of generated identicon. It can be modified according to needs.
+  const imageSize = 128
```
NeOMakinG, line 15: What do you guys think about making this configurable? Might be too big in some cases

packages/asset-service/src/service/AssetService.ts
```diff
@@ -1,9 +1,64 @@
...
 
+/* eslint-disable */
+const Identicon = require('identicon.js')
+const { fonts, renderPixels } = require('js-pixel-fonts')
+/* eslint-enable */
+
 import assetsDescriptions from './descriptions.json'
 import localAssetData from './generatedAssetData.json'
 
+const getRenderedIdenticonB64 = (symbol: string, tokenId: string): string => {
```
NeOMakinG, line 13: I think the original demand was to have a module, in case you want to generate the image:

`Add a module to asset-service to deterministically generate an icon based on the caip19 of an asset.`

It means that this method should be exported from `asset-service` so any app using it can generate the icon on the fly

`During rendering in web, if the asset does not have a remote resource icon (property icon on the Asset type), use the new module to generate the icon on the fly and display that icon.`

I love the fact that finally, we provide directly the generated icon into the asset, instead of making every users generate it but it can lead to some places where you want to have a 64x64 icon but you only have 128x128, wdyt?

packages/asset-service/src/service/AssetService.ts
Soptq: Yes, that is indeed a requirement from the issue. However, during my testing I find the text might be too small if we need to render 5 characters. Here is the example:

Render 5 characters, scale = 4:
Render 5 characters:

Render 3 characters, scale = 7:
Render 3 characters:

Render 1 characters, scale = 12:
Render 1 characters:

So I concluded that rendering 1-3 characters might be most suitable for better UX. Anyway, in the latest commit we will render 3 characters instead of 1.

packages/asset-service/src/service/AssetService.ts
Soptq: The latest commit makes the size configurable

packages/asset-service/src/service/AssetService.ts
Soptq: You are right, I have updated the PR to allow more configurable options.

Soptq:
Hi @NeOMakinG , thanks for your constructive feedbacks and I have updated the PR to address each of your concerns ;)

I also added a slight shadow to the rendered identicon without increasing computational complexity, now the rendering is like:
final

packages/asset-service/src/service/AssetService.ts
Soptq: And the text now is also configurable so that whether to render 1 character or 5 characters can be decided by the downstream applications.

NeOMakinG:
> Hi @NeOMakinG , thanks for your constructive feedbacks and I have updated the PR to address each of your concerns ;)
> 
> I also added a slight shadow to the rendered identicon without increasing computational complexity, now the rendering is like: final

Looks like you read in my mind, yesterday I was thinking "The text is white, do we have some cases where the text could be unreadable?" 👍

NeOMakinG:
I will let maintainers review the code style and implementation, because I myself struggle while contributing but I found a little point:
- `Identicon` shouldn't be used as a file name for `getRenderedIdenticonBase64`, you are exporting a function generating an image with text, which is using Identicon, but you are not exporting Identicon, so the naming looks wrong to me

I'm just a bit curious, what's inside the SVG, an image tag with the base64?

Soptq:
> I will let maintainers review the code style and implementation, because I myself struggle while contributing but I found a little point:
> 
> * `Identicon` shouldn't be used as a file name for `getRenderedIdenticonBase64`, you are exporting a function generating an image with text, which is using Identicon, but you are not exporting Identicon, so the naming looks wrong to me
> 
> I'm just a bit curious, what's inside the SVG, an image tag with the base64?

1. You are right! I guess I will rename it to `IdenticonPNGBase64` or something like that.
2. Yes, it should be something like ``. Actually in current PR this module only supports PNG so `format` should not be an option in `IdenticonOptions`. Will update it as well very soon.

NeOMakinG:
> > I will let maintainers review the code style and implementation, because I myself struggle while contributing but I found a little point:
> > 
> > * `Identicon` shouldn't be used as a file name for `getRenderedIdenticonBase64`, you are exporting a function generating an image with text, which is using Identicon, but you are not exporting Identicon, so the naming looks wrong to me
> > 
> > I'm just a bit curious, what's inside the SVG, an image tag with the base64?
> 
> 1. You are right! I guess I will rename it to `IdenticonPNGBase64` or something like that.
> 2. Yes, it should be something like ``. Actually in current PR this module only supports PNG so `format` should not be an option in `IdenticonOptions`. Will update it as well very soon.

I would myself just call it `generateAssetIcon` I think so, but as I said, I'm not a maintainer here, my review is not mandatory :p

Soptq:
> > > I will let maintainers review the code style and implementation, because I myself struggle while contributing but I found a little point:
> > > 
> > > * `Identicon` shouldn't be used as a file name for `getRenderedIdenticonBase64`, you are exporting a function generating an image with text, which is using Identicon, but you are not exporting Identicon, so the naming looks wrong to me
> > > 
> > > I'm just a bit curious, what's inside the SVG, an image tag with the base64?
> > 
> > 
> > 
> > 1. You are right! I guess I will rename it to `IdenticonPNGBase64` or something like that.
> > 2. Yes, it should be something like ``. Actually in current PR this module only supports PNG so `format` should not be an option in `IdenticonOptions`. Will update it as well very soon.
> 
> I would myself just call it `generateAssetIcon` so you don't stick with a dependency if you want to change Identicon later I think so, but as I said, I'm not a maintainer here, my review is not mandatory :p
> 
> Also, be careful, if you're still working on it but wants to receive some reviews, just set the `draft` mode :)

Thanks for your advice! I think this module is ready for reviewing, sometimes some ideas just pop out in my head, making me want to do some refactoring or add some additional features before merging, but the PR itself is already functional and meet the requirements of the related issue at the first place. ;)

packages/asset-service/src/service/AssetService.ts
```diff
@@ -12,15 +14,35 @@ export const flattenAssetData = (assetData: BaseAsset[]): Asset[] => {
...
+              size: 128,
+              background: [45, 55, 72, 255]
+            },
+            identiconText: {
+              symbolScale: 7,
+              enableShadow: true
+            }
+          }
+          if (tokenId)
+            flatAssetDatum.icon = getRenderedIdenticonBase64(
```
theoboldfrazier, line 39: We'll want to generate icons for base assets as well as tokens. Currently, there aren't a lot of base assets, but in the future there may be some that may not have an icon. So I think this same logic applied to the base assets as well should work.

cjthompson, line 39: This is being called during app launch. We measured the time to generate an icon at around 30ms. So, if there are a substantial number of assets without icons, this loop could cause a significant delay in app startup time.

I'm discussing with some other core devs about whether we should put this in the generate script so it happens at build time, or have the `AssetIcon` component leverage the function to generate the icon only at display time.

I'm open to thoughts on these or other approaches to reduce the time to first load of the app.

theoboldfrazier: @Soptq Major Hayes makes a good point. I pegged the icon generation at 30ms per icon. If we have a very large set of assets without an icon, say 1000, then that would add 30sec to the initial app startup.

Can you move the logic implementation from here to the asset generation files in asset-service/generateAssetData/ ethTokens and cosmos?

0xdef1cafe: adding another +1 to this, especially if we can just base64 them into the static data. 

the other option is to expose this `getRenderedIdenticonBase64` function that web can call at runtime to generate them as components mount, rather than at app load - i don't expect generating these is going to be any slower than a network request to load an icon, and it keeps the initial payload smaller.

Soptq:
@theoboldfrazier @cjthompson @0xdef1cafe how about asynchronously lazy loading the asset icon in the web end?

For now I think I can make this PR generate asset icons at compile time, but I am a little concerned that the size of generated json file will increase, resulting in a longer startup time as well (however, it will be surely much quicker than current implementation especially when the assets are abundant)

cjthompson:
> @theoboldfrazier @cjthompson @0xdef1cafe how about asynchronously lazy loading the asset icon in the web end?
> 
> For now I think I can make this PR generate asset icons at compile time, but I am a little concerned that the size of generated json file will increase, resulting in a longer startup time as well (however, it will be surely much quicker than current implementation especially when the assets are abundant)

Can you do a hacky experiment to see how the size of the json file will change?

Soptq:
> > @theoboldfrazier @cjthompson @0xdef1cafe how about asynchronously lazy loading the asset icon in the web end?
> > For now I think I can make this PR generate asset icons at compile time, but I am a little concerned that the size of generated json file will increase, resulting in a longer startup time as well (however, it will be surely much quicker than current implementation especially when the assets are abundant)
> 
> Can you do a hacky experiment to see how the size of the json file will change?

Sure.

Before: 3.49 MB
After: 3.58 MB

The resolution of generated icon is 128 * 128

cjthompson:
I'm okay with that. If you can update the PR with that change I'll approve it.

Thanks for your work on this. Internally we should have spent a little more time talking about this before creating the bounty so I appreciate your flexibility.

theoboldfrazier (approved):
Looks good!

0xdef1cafe:
@Soptq we're accepting this work, but closing this PR and running the generate script from a core team member for security reasons here https://github.com/shapeshift/lib/pull/400 - we'll get the bounty paid out, thanks for the great work :)

0xean:
@Soptq - payment plus tip sent! thanks!

PR #390, opened 2022-02-18
WIP CosmosChainAdapter

cjthompson:
# DRAFT

Starting work on a CosmosChainAdapter. There's a lot of unchained work that's not ready yet but it would be nice to have a ChainAdapter that is partially functional that we can use for the web code that needs to get done. 

Many of the functions will probably just return mock data or nothing until the unchained work is ready.


cjthompson:
it turns out that @pastaghost is already working on this.

PR #391, opened 2022-02-20
refactor(EthereumChainAdapter): use maker functions for tests

gomesalexandre:
- move `const getGasFeesMockedResponse = {}` to a `makeGetGasFeesMockedResponse` function
- move `const estimateGasMockedResponse = {}` to a `makeEstimateGasMockedResponse` function
- move `const getInfoMockResponse = {}` to a `makeGetInfoMockResponse` function
- add a `makeGetAccountMockResponse` function instead of passing a new `getAccount` mock object every test, since only the balance/erc20 balance can change
- add a `makeChainAdapterArgs` function instead of mutating the `args` defined in `beforeEach`


theoboldfrazier (approved):

cjthompson (approved):
This looks great.

Just as something to think about for future testing, when needed, I often use `lodash.merge` as part of my mock data functions so you can have a set of default data but override with a deep merge as needed.

DaoDev44 (approved):

PR #394, opened 2022-02-23
feat: osmosis market service

### LINKED ISSUE #364: Cosmos: Add MarketDataProvider for cosmos
AC:
* Find a data provider for cosmos market data
* Create osmosisToCaip19 adapter
* Create a MarketDataProvider class to retrieve market data from said provider

See existing Coincap data provider as a reference.

DaoDev44:
- Add Osmosis Market Data Provider for assets that are only on Osmosis
- Add Osmosis market adapter

closes #364


packages/market-service/src/osmosis/osmosis.ts
```diff
@@ -0,0 +1,145 @@
...
+      )
+      const marketData = data[0]
+
+      if (!marketData) {
+        return {
+          price: '',
+          marketCap: '',
+          volume: '',
+          changePercent24Hr: 0
+        }
```
theoboldfrazier, line 63: Would we want to return null here?

DaoDev44: Done

0xdef1cafe (approved):

PR #396, opened 2022-02-25
feat: add metadata to transaction WebSocket messages

0xApotheosis:
**Description**
Add metadata to transaction WebSocket messages to propagate the addition of metadata in https://github.com/shapeshift/unchained/pull/278.

### DIFF
--- a/packages/types/src/chain-adapters/index.ts
+++ b/packages/types/src/chain-adapters/index.ts
@@ -137,7 +137,7 @@ export type SubscribeTxsMessage<T extends ChainTypes> = {
   status: TxStatus
   tradeDetails?: TradeDetails
   transfers: Array<TxTransfer>
-  data?: TxMetadata | undefined
+  data?: TxMetadata
 }
 
 export enum TradeType {
--- a/packages/types/src/chain-adapters/index.ts
+++ b/packages/types/src/chain-adapters/index.ts
@@ -153,7 +153,7 @@ export type TradeDetails = {
 
 export interface TxMetadata {
   method?: string
-  parser?: string
+  parser: string
 }
 
 export type TxTransfer = {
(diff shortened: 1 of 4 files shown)


packages/types/src/chain-adapters/index.ts
```diff
@@ -137,6 +137,7 @@ export type SubscribeTxsMessage<T extends ChainTypes> = {
   status: TxStatus
   tradeDetails?: TradeDetails
   transfers: Array<TxTransfer>
+  data?: TxMetadata | undefined
```
kaladinlight, line 140:
```suggestion
  data?: TxMetadata
```

packages/types/src/chain-adapters/index.ts
```diff
@@ -150,6 +151,11 @@ export type TradeDetails = {
   type: TradeType
 }
 
+export interface TxMetadata {
+  method?: string
+  parser?: string
```
kaladinlight, line 156: If parser metadata is included, should we at least require these two fields? I know this currently matches the payload coming from unchained, but just got me thinking.

kaladinlight (approved):
Two small comments, but looks good!

packages/types/src/chain-adapters/index.ts
0xApotheosis: That's a fair point, though at the moment `method` is always undefined for zrx parsing, as we haven't yet sourced and added an ABI for it.

packages/types/src/chain-adapters/index.ts
0xApotheosis: https://github.com/shapeshift/lib/pull/396/commits/501b22431f8b51ae1febae7c0635e324518a5381

packages/types/src/chain-adapters/index.ts
kaladinlight: True. Let's keep it in mind as a required base metadata type that we can extend if needed if it makes sense.

0xApotheosis: Agree - I've made `parser` required for now in https://github.com/shapeshift/lib/pull/396/commits/b768e20811faf34b50379a38a189637b77e2336e.

0xdef1cafe (approved):

PR #397, opened 2022-02-25
feat: add Cosmos based chain adapters

cjthompson:
* add `CosmosSdkBaseAdapter` as a common implementation for Cosmos-based chain adapters
* add `CosmosChainAdapter` and `OsmosisChainAdapter` as chain-specific implementations

This PR will be a little smaller if #398 is merged first

PR #398, opened 2022-02-25
fix: caip support for Osmosis

cjthompson:
Osmosis needs it own ChainType but the caip namespace is `cosmos`

### DIFF
--- a/packages/caip/src/caip19/caip19.test.ts
+++ b/packages/caip/src/caip19/caip19.test.ts
@@ -33,8 +33,8 @@ describe('caip19', () => {
       expect(result).toEqual('cosmos:cosmoshub-4/slip44:118')
     })
 
-    it('can return ibc asset caip 19 for cosmos', () => {
-      const chain = ChainTypes.Cosmos
+    it('can return ibc asset caip 19 for osmosis', () => {
+      const chain = ChainTypes.Osmosis
       const network = NetworkTypes.OSMOSIS_MAINNET
       const assetNamespace = AssetNamespace.IBC
       const assetReference = '346786EA82F41FE55FAD14BF69AD8BA9B36985406E43F3CB23E6C45A285A9593'
@@ -44,8 +44,8 @@ describe('caip19', () => {
       )
     })
 
-    it('can return native asset caip 19 for cosmos', () => {
-      const chain = ChainTypes.Cosmos
+    it('can return native asset caip 19 for osmosis', () => {
+      const chain = ChainTypes.Osmosis
       const network = NetworkTypes.OSMOSIS_MAINNET
       const assetNamespace = AssetNamespace.NATIVE
       const assetReference = 'uion'
@@ -53,8 +53,8 @@ describe('caip19', () => {
       expect(result).toEqual('cosmos:osmosis-1/native:uion')
     })
 
-    it('can return cw20 asset caip 19 for cosmos', () => {
-      const chain = ChainTypes.Cosmos
+    it('can return cw20 asset caip 19 for osmosis', () => {
+      const chain = ChainTypes.Osmosis
       const network = NetworkTypes.OSMOSIS_MAINNET
       const assetNamespace = AssetNamespace.CW20
       const assetReference = 'canlab'
@@ -62,8 +62,8 @@ describe('caip19', () => {
       expect(result).toEqual('cosmos:osmosis-1/cw20:canlab')
     })
 
-    it('can return cw721 asset caip 19 for cosmos', () => {
-      const chain = ChainTypes.Cosmos
+    it('can return cw721 asset caip 19 for osmosis', () => {
+      const chain = ChainTypes.Osmosis
       const network = NetworkTypes.OSMOSIS_MAINNET
       const assetNamespace = AssetNamespace.CW721
       const assetReference = 'osmosiskitty'
@@ -233,7 +233,7 @@ describe('caip19', () => {
     it('can parse a osmosis native token', () => {
       const caip19 = 'cosmos:osmosis-1/slip44:118'
       const { chain, network, assetNamespace, assetReference } = fromCAIP19(caip19)
-      expect(chain).toEqual(ChainTypes.Cosmos)
+      expect(chain).toEqual(ChainTypes.Osmosis)
       expect(network).toEqual(NetworkTypes.OSMOSIS_MAINNET)
       expect(assetNamespace).toBeUndefined()
       expect(assetReference).toBeUndefined()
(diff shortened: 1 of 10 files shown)


0xApotheosis (approved):
LGTM

packages/caip/src/caip2/caip2.ts
```diff
@@ -32,72 +32,84 @@ type ToCAIP2Args = {
...
+        case NetworkTypes.COSMOSHUB_MAINNET:
+        case NetworkTypes.COSMOSHUB_VEGA: {
+          const reference: ChainReference = referenceMap[network]
+          const caip2 = `${namespace}:${reference}`
+          return caip2
+        }
+      }
+      break
+    }
+    case ChainTypes.Osmosis: {
```
gomesalexandre, line 108: Can't we reuse the same logic for both swith cases?

gomesalexandre (approved):
LGTM, checked against the `chain-adapters` implementation in https://github.com/shapeshift/lib/pull/397/files

packages/caip/src/caip2/caip2.ts
0xdef1cafe: yeah this is valid

packages/types/src/base.ts
```diff
@@ -22,7 +22,8 @@ export enum ContractTypes {
 export enum ChainTypes {
   Ethereum = 'ethereum',
   Bitcoin = 'bitcoin',
-  Cosmos = 'cosmos'
+  Cosmos = 'cosmos',
+  Osmosis = 'osmosis'
```
0xdef1cafe, line 26: this will need a corresponding

0xdef1cafe (changes requested):

packages/caip/src/caip19/caip19.test.ts
```diff
@@ -33,8 +33,8 @@ describe('caip19', () => {
       expect(result).toEqual('cosmos:cosmoshub-4/slip44:118')
     })
 
-    it('can return ibc asset caip 19 for cosmos', () => {
-      const chain = ChainTypes.Cosmos
+    it('can return ibc asset caip 19 for osmosis', () => {
```
kaladinlight, line 36: We should add the applicable corresponding cosmos caip19 tests as well. There is no ibc cosmos test case or osmosis mainnet test case

I am not sure if the cw variants are applicable to cosmos or osmosis specific

packages/caip/src/caip2/caip2.test.ts
```diff
@@ -91,14 +91,14 @@ describe('caip2', () => {
     it('can turn Osmosis mainnet to chain and network', () => {
       const osmosisCaip2 = 'cosmos:osmosis-1'
       const { chain, network } = fromCAIP2(osmosisCaip2)
-      expect(chain).toEqual(ChainTypes.Cosmos)
+      expect(chain).toEqual(ChainTypes.Osmosis)
```
kaladinlight, line 94: No tests for invalid namespace or reference added.

packages/caip/src/caip2/caip2.ts
```diff
@@ -126,20 +138,16 @@ export const fromCAIP2: FromCAIP2 = (caip2) => {
       const chain = ChainTypes.Cosmos
```
kaladinlight, line 138: Stylistic pref to just have ChainTypes.Cosmos inline inside switch now that there are two types for the same namespace.

packages/caip/src/caip2/caip2.ts
```diff
@@ -32,72 +32,84 @@ type ToCAIP2Args = {
...
       break
     }
-
-    case ChainTypes.Bitcoin: {
+    case ChainTypes.Cosmos: {
       const referenceMap = shapeShiftToCAIP2[chain].reference
       switch (network) {
-        case NetworkTypes.MAINNET:
-        case NetworkTypes.TESTNET: {
+        case NetworkTypes.COSMOSHUB_MAINNET:
```
kaladinlight, line 99: Has there been discussion on how mainnet will be managed with cosmos upgrades? Will `COSMOSHUB_MAINNET` just keep changing the underlying value once version is upgraded from v4? Would we want to be explicit and just name it `COSMOSHUB_MAINNET_4`? Will there be any need to reference a previous mainnet version where this would be useful (previous tx history)?

packages/caip/src/caip2/caip2.ts
```diff
@@ -32,72 +32,84 @@ type ToCAIP2Args = {
...
+        case NetworkTypes.COSMOSHUB_VEGA: {
```
kaladinlight, line 100: Not sure how testnet upgrades go, but the same naming comment could be applied here as well. Mainnet currently is generic whereas testnet is explicit in this case. I think explicit is the way to go similar to `ETH_ROPSTEN` vs `ETH_RINKEBY`

packages/caip/src/caip2/caip2.ts
cjthompson: I get type errors when I try to consolidate the Cosmos and Osmosis switch statement into one due to how it looks up the reference map.

packages/caip/src/caip2/caip2.ts
cjthompson: Are you making a suggested change? If so can you be more clear about what change you want to see? I don't understand what you're trying to say I should change.

packages/caip/src/caip2/caip2.ts
cjthompson: No, but I didn't add this enum. It's already in `main`. If there should be a change to this I feel that's outside the scope of the intended changes for this PR specifically.

kaladinlight: Well maybe quick poll on follow up discussion?

packages/caip/src/caip2/caip2.ts
kaladinlight: This is just an additional comment in line with the previous question surrounding naming convention for our network types. I think it deserves a quick discussion over

packages/caip/src/caip2/caip2.ts
cjthompson: I don't understand the question.

cjthompson:
I understand the desire to have a discussion about naming of `NetworkTypes` enum. However, this PR doesn't make any changes to that enum so I don't think it's fair to block this PR because that discussion is out of scope of these changes.

packages/caip/src/caip2/caip2.ts
gomesalexandre: > Will there be any need to reference a previous mainnet version where this would be useful (previous tx history)?

I'm not sure that's something we would need. If you try to access previous mainnets on Figment's [Hubble](https://hubble.figment.io/cosmos/chains/cosmoshub-3), you will have an info message stating that "Chain appears to be stopped. No new blocks for over 1 year!".
My understanding is we should be able to always use the newest chain, but happy to continue this in Discussions.

PR #405, opened 2022-02-28
feat: add "Contract" TxType

0xApotheosis:
**Description**

Adds `TxType.Contract`, in support of current tx parsing work.

**Context**

Supports https://github.com/shapeshift/web/issues/1068


gomesalexandre (approved):
LGTM according to https://github.com/shapeshift/web/pull/1076

What about calling this type `ContractCall` however? That would be reflecting the tx type better IMO.

PR #411, opened 2022-03-03
feat: utxo chain adapter refactor

### LINKED ISSUE #125: Refactor BitcoinChainAdapter to UTXOChainAdapter
# Background
Chain adapters are an abstraction layer that our applications use to hide chain specific functionality.  Currently we have implemented a  [BitcoinChainAdapter](https://github.com/shapeshift/lib/blob/main/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts)  that should be able to be made generic enough to be turned into a UTXOChainAdapter that supports all UTXO base chains (LTC, DOGE, etc)

This will require some modifications to inject chain specific dependencies via the constructor, and it's internal functionality. This will allow us to support many utxo chains in the future by instantiating a new instance of the adapter with appropriate constructor arguments.

# Acceptance Criteria
- refactor the BitcoinChainAdapter to `AbstractUTXOChainAdapter`
- create concrete classes for `BitcoinChainAdapter` and `DogecoinChainAdapter`
- getType() should return the UTXO ChainType that is based on the arguments passed into the constructor. 
- getCaip2() should return correct caip2 based on the UTXO passed into the constructor
- getAccount() returns a more generic promise type with a correct chain in the return object
- getTxHistory() returns a more generic promise type with a correct chain in the return object
- use slip44 coin types from the spec https://github.com/satoshilabs/slips/blob/master/slip-0044.md
- Add / modify unit tests for the new generic adapter
- Search for anywhere we are creating this chain adapter in a hard coded fashion would need to be updated. 

See https://github.com/shapeshift/lib/pull/289 as a reference to how this is being done for Cosmos.

# Bounty Hunters
- This issue may take a bit of exploring to understand how this can be refactored in a generic fashion and the bounty amount hopefully reflects some time spent on that. 
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Join our discord [here](https://discord.gg/shapeshift)
- Please include an expected timeline for you to complete work in the work plan when you apply for this bounty!

0xean:
This PR moves much of the BitcoinChainAdapter logic in a UTXOBaseAdapter class that can be re-used for future UTXO's.

It also includes some minor updates to the README

Have a few questions to run down before we merge and will add those as comments momentarily.

Currently, all existing tests pass, but have not done any additional testing past that. 

Closes #125

### DIFF
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
@@ -42,20 +42,14 @@ export class ChainAdapter
 
   constructor(args: ChainAdapterArgs) {
     super(args)
-    if (args.chainId) {
-      try {
-        const { chain } = caip2.fromCAIP2(args.chainId)
-        if (chain !== ChainTypes.Bitcoin) {
-          throw new Error()
-        }
-        this.chainId = args.chainId
-      } catch (e) {
-        throw new Error(`The ChainID ${args.chainId} is not supported`)
-      }
-    } else {
-      throw new Error('The ChainID must be supplied in constructor args')
+    if (!args.chainId) {
+      throw new Error('chainId required')
     }
-
+    const { chain } = caip2.fromCAIP2(args.chainId)
+    if (chain !== ChainTypes.Bitcoin) {
+      throw new Error('chainId must be a bitcoin chain type')
+    }
+    this.chainId = args.chainId
     this.coinName = args.coinName
   }
 
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
@@ -14,7 +14,7 @@ import {
   NetworkTypes,
   UtxoAccountType
 } from '@shapeshiftoss/types'
-import { bitcoin } from '@shapeshiftoss/unchained-client'
+import * as unchained from '@shapeshiftoss/unchained-client'
 import coinSelect from 'coinselect'
 import split from 'coinselect/split'
 
--- a/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
+++ b/packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
@@ -1,66 +1,35 @@
-import { AssetNamespace, AssetReference, CAIP2, caip2, caip19 } from '@shapeshiftoss/caip'
+import { caip2 } from '@shapeshiftoss/caip'
 import {
   bip32ToAddressNList,
   BTCOutputAddressType,
   BTCSignTx,
   BTCSignTxInput,
   BTCSignTxOutput,
-  HDWallet,
-  PublicKey,
   supportsBTC
 } from '@shapeshiftoss/hdwallet-core'
-import {
-  BIP44Params,
-  chainAdapters,
-  ChainTypes,
-  NetworkTypes,
-  UtxoAccountType
-} from '@shapeshiftoss/types'
-import * as unchained from '@shapeshiftoss/unchained-client'
+import { chainAdapters, ChainTypes, NetworkTypes, UtxoAccountType } from '@shapeshiftoss/types'
+import { bitcoin } from '@shapeshiftoss/unchained-client'
 import coinSelect from 'coinselect'
 import split from 'coinselect/split'
-import WAValidator from 'multicoin-address-validator'
 
 import { ChainAdapter as IChainAdapter } from '../api'
 import { ErrorHandler } from '../error/ErrorHandler'
 import {
   accountTypeToOutputScriptType,
   accountTypeToScriptType,
-  bnOrZero,
-  convertXpubVersion,
   getStatus,
   getType,
   toPath,
   toRootDerivationPath
 } from '../utils'
+import { ChainAdapterArgs, UTXOBaseAdapter } from '../utxo/UTXOBaseAdapter'
 
-export interface ChainAdapterArgs {
-  providers: {
-    http: unchained.bitcoin.V1Api
-    ws: unchained.ws.Client<unchained.SequencedTx>
-  }
-  coinName: string
-  chainId?: CAIP2
-}
-
-export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
-  private readonly providers: {
-    http: unchained.bitcoin.V1Api
-    ws: unchained.ws.Client<unchained.SequencedTx>
-  }
-
-  public static readonly defaultBIP44Params: BIP44Params = {
-    purpose: 84, // segwit native
-    coinType: 0,
-    accountNumber: 0
-  }
-
-  private readonly chainId: CAIP2 = 'bip122:000000000019d6689c085ae165831e93'
-
-  // TODO(0xdef1cafe): constraint this to utxo coins and refactor this to be a UTXOChainAdapter
-  coinName: string
-
+export class ChainAdapter
+  extends UTXOBaseAdapter<ChainTypes.Bitcoin>
+  implements IChainAdapter<ChainTypes.Bitcoin>
+{
   constructor(args: ChainAdapterArgs) {
+    super(args)
     if (args.chainId) {
       try {
         const { chain } = caip2.fromCAIP2(args.chainId)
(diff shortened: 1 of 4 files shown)


packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,159 @@
...
+  bnOrZero,
+  convertXpubVersion,
+  toRootDerivationPath
+} from '../utils'
+
+export type UTXOChainTypes = ChainTypes.Bitcoin // to be extended in the future to include other UTXOs
+
+export interface ChainAdapterArgs {
+  providers: {
+    http: bitcoin.api.V1Api //?
```
0xean, line 20: I am assuming we want to make this more generic, but from what I can tell there is not an interface that would allow this for a generic UTXO unchained provider.  In the absence of that, should we leave this as-is for now or move it up to the implementing `BitcoinChainAdapter `class

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,159 @@
...
+ * Base chain adapter for all UTXO chains. When extending please add your ChainType to the
+ * UTXOChainTypes. For example:
+ *
+ * `export type UTXOChainTypes = ChainTypes.Bitcoin | ChainTypes.Litecoin`
+ */
+export abstract class UTXOBaseAdapter<T extends UTXOChainTypes> implements IChainAdapter<T> {
+  protected chainId: CAIP2
+  protected coinName: string
+  protected readonly providers: {
+    http: bitcoin.api.V1Api
```
0xean, line 37: same question

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,159 @@
...
+    ws: bitcoin.ws.Client
+  }
+
+  public static readonly defaultBIP44Params: BIP44Params = {
+    purpose: 84, // segwit native
```
0xean, line 42: also wanted to check on this, due to lack of experience with UTXO's and the `purpose` here.  Not sure if this should be moved up a level to the implementing class

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -71,79 +40,18 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
       } catch (e) {
         throw new Error(`The ChainID ${args.chainId} is not supported`)
       }
+    } else {
```
0xdef1cafe, line 43: i'd probably throw here rather than opaquely defaulting to bitcoin

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xdef1cafe: i think this is ok for now - but will let @kaladinlight chime in here, as these types come from unchained

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xdef1cafe: yeah - these BIP44Params should be moved into the implementing class imo. @cjthompson can weigh in here too

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  getCaip2(): CAIP2 {
+    return this.chainId
+  }
+
+  getChainId(): CAIP2 {
+    return this.chainId
+  }
+
+  async getAccount(pubkey: string): Promise<chainAdapters.Account<T>> {
```
0xdef1cafe, line 90: this may also need to move into the implementing class. `this.providers.http.getAccount` for BCH for example will end up calling a different coinstack at like `api.doge.shapeshift.com` - @kaladinlight can confirm

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,159 @@
...
+        },
+        pubkey: data.pubkey
+      } as chainAdapters.Account<T>
+    } catch (err) {
+      return ErrorHandler(err)
+    }
+  }
+
+  async broadcastTransaction(hex: string): Promise<string> {
+    const { data } = await this.providers.http.sendTx({
```
0xdef1cafe, line 120: this prob needs to to move to concrete class too

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
kaladinlight: This is matching how cosmos base adapter is currently, but you are correct that each coinstack will have their own api client that would need to be controlled by the implementing class. As we only support btc right now, we can start here and continue to iterate

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
kaladinlight: Yeah it seems like we need to think through this abstraction (both cosmos and utxo) a bit more to ensure we are setting this up correctly.

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: okay, going to resolve this conversation and leave as is for now.  I will add a comment to the codebase.

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
0xean: this was the behavior of the previous adapter, but will modify now

0xean: @0xdef1cafe - lets chat there is specifically a test for "should return chainAdapter with default Bitcoin chainId if called with no chainId" so am concerned there was a reason for this.

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: if we had a generic provider interface, this would not matter which coinstack it ended up calling, right?

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: same comment as above

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: @kaladinlight - can you confirm that if we had a proper abstraction on the provider this would be fine to leave in the base class?

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
0xean: per discussion with @0xdef1cafe , killing this behavior and the associated test

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
kaladinlight: Each http provider is generated directly from the swagger.json being served by the specific coinstack. There is BaseAPI interface that each coinstack api must implement, but the base return payloads are very minimal and intended to be extended by each coinstack as they see fit. We also do not expose this interface anywhere at this time.

We could try and create this interface as a part of the client package to use within chain adapters. I'm also wondering if we could utilize a type union and that would just result in the shared result set. We do need to think about what it would look like to extend the base logic with any possible unique data for a specific child chain adapter.

0xean: Okay, for this PR is your recommendation then to move this to the implementing class until we have a more generic solution?

kaladinlight: I think you can just leave it for now and make a comment that when additional utxo chains are added, we need to ensure the generic base implementation works against all imported api types. We only support BTC anyways right now and this is the same setup that cosmos base adapter currently has.

0xean: awesome, thanks will do and move this to a "non-draft" pr so we can get it merged.

0xean:
alright, this should be good to go now, all conflicts have been resolved.

0xean:
argh, one sec and then should be ready for review.

0xean:
alright actually ready now.

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
```diff
@@ -0,0 +1,161 @@
...
+      const { chain, network } = caip2.fromCAIP2(caip)
+      const { data } = await this.providers.http.getAccount({ pubkey: pubkey })
+
+      const balance = bnOrZero(data.balance).plus(bnOrZero(data.unconfirmedBalance))
+
+      return {
+        balance: balance.toString(),
+        chain: this.getType(),
+        caip2: caip,
+        caip19: caip19.toCAIP19({
```
kaladinlight, line 107: Another TODO is we should probably set `assetId` similarly to keeping track of `chainId` that we would be able to use for the native asset caip19 as determined by the implementing class. This would remove the Bitcoin reference in the base adapter

kaladinlight (approved):
A couple follow up remarks, but generally looks good to go as a starting place to start iterating on. Thanks!

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -19,48 +17,31 @@ import {
...
-  chainId?: CAIP2
-}
-
-export class ChainAdapter implements IChainAdapter<ChainTypes.Bitcoin> {
-  private readonly providers: {
-    http: bitcoin.api.V1Api
-    ws: bitcoin.ws.Client
-  }
-
+export class ChainAdapter
```
asamere, line 33: Consider naming this BitcoinChainAdapter or similar to avoid name conflict with api.ChainAdapter interface?

0xean: I assume this would introduce breaking changes upstream since this was the name already in place.  But yea, I agree it would be better for a variety of reasons.

kaladinlight: Base interface is already being imported as `IChainAdapter`. I have personally always found our prefixing unnecessary and overly verbose, ie. `bitcoin/BitcoinChainAdapter.ts`, `export class BitcoinChainAdapter`. Everything is already structured and scoped under the appropriate chain name and where we are importing multiple chains, the importing follows: 
```
import * as bitcoin from './bitcoin'
import * as cosmos from './cosmossdk/cosmos'
```

This reads really nicely as `new bitcoin.ChainAdapter` imo.

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: I am not sure I am following.  The bitcoin reference in the base adapter is just for the provider. Can you clarify what you are asking to change here?

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
0xean: Okay, so are going with no change for the time being?

kaladinlight: Regardless of official decision, I don't think it belongs in this PR regardless

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
kaladinlight: I probably should have made the comment on a better line. See here: https://github.com/shapeshift/lib/pull/411/files/784d591ee848ebe7444da8bf528a595c0727bde7#diff-91f7561fb5e5062d3ca7bc6c42e96c285c398a4518d8485bdcb9f672cb4148d7R111

We are using `AssetReference.Bitcoin` in the `UTXOBaseAdapter` to create the caip19 string, whereas I am suggesting that if we add an `assetId` (bitcoin caip19) class variable that is set from the bitcoin `ChainAdapter` class, we can use `this.assetId` which will be of the proper type as determined by the implementing class.

kaladinlight (approved):

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
asamere: yep all good with me. thank @kaladinlight

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: Got it, make sense. please take a look at the latest commit

kaladinlight: That looks great!

TBD if we want to keep `getCaip19()`. We have both for caip2 since the `getCaip2()` was added to the base `ChainAdapter` interface, but if we are trying to move naming to `chainId` and `assetId` we may just want the `getAssetId()` function. I haven't been following super closely with that naming discussion, so may be good for @0xdef1cafe or whoever else to weigh in.

You have my stamp for everything as is and we can iterate as needed from here.

asamere (approved):
Regarding "providers" we should keep in mind eventually/ultimately backend providers will be dynamic, with multiple potential backends for each chain. These could come from an ETH contract, IPFS, etc but will change as operators and infra availability fluctuates.

packages/chain-adapters/src/utxo/UTXOBaseAdapter.ts
0xean: Thanks a ton for all the help and reviews. @0xdef1cafe - lmk your preference here and we can merge this so I don't have to fix yet another conflict ;)

PR #415, opened 2022-03-04
feat: cosmos `subscribeTxs`

kaladinlight:
- Add websocket client
- Implement `subscribeTxs`
- Use `unchained-tx-parser` client side


packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -27,6 +28,8 @@ export class ChainAdapter
 
   constructor(args: ChainAdapterArgs) {
     super(args)
+
+    this.parser = new parser.cosmos.TransactionParser({ chainId: this.chainId })
```
cjthompson, line 32: Why does a transaction parser need to be a class instance? I would think simple functions would be enough.

However, a discussion on this topic should have been had before the implementation was written rather than after.

kaladinlight: We instantiate rpc providers, potentially api clients (midgard?), and adds the possibility of registering custom parsers.

elmutt (approved):

0xdef1cafe (approved):

PR #426, opened 2022-03-10
fix: removed error from required chainId and set on class

theoboldfrazier:


0xean:
For reference: https://github.com/shapeshift/lib/pull/411#discussion_r819681654

packages/chain-adapters/src/bitcoin/BitcoinChainAdapter.ts
```diff
@@ -33,17 +33,15 @@ export class ChainAdapter
     coinType: 0,
     accountNumber: 0
   }
+  protected readonly supportedChainIds: CAIP2[] = ['bip122:000000000019d6689c085ae165831e93']
+  protected readonly chainId: CAIP2 = this.supportedChainIds[0]
```
0xean, line 37: will this not override the base class `chainId` and not allow us to pass in a chainId in the parameters if we want to?

my TS knowledge may be failing me here, but maybe instead just assigning 

`this.chainId =  this.supportedChainIds[0]`
In the constructor if chainId is not defined would be better.

theoboldfrazier: This makes sense to me and I've made this update.

0xdef1cafe (approved):

PR #430, opened 2022-03-11
feat: Write Asset Descriptions for LP tokens - crvDOLA

joaophelip:


willyogo (changes requested):
Please use all caps for token symbols `DOLA`

https://curve.fi/factory/27 is the right pool, but it's important to link straight to the deposit page and not the swap page `https://curve.fi/factory/27/deposit`

PR #435, opened 2022-03-14
feat: cosmos sign and broadcast tx

### LINKED ISSUE #306: ChainAdapters: Cosmos - implement signAndBroadcastTransaction

DaoDev44:
- Add `signAndBroadcastTransaction` to cosmos and osmosis chain adapters
- Update `buildSendTransaction` to have dynamic `accountNumber` and `sequence` number
- Update chainAdapter CLI 

- closes #306

### DIFF
--- a/packages/chain-adapters/src/ChainAdapterCLI.ts
+++ b/packages/chain-adapters/src/ChainAdapterCLI.ts
@@ -211,7 +211,6 @@ const main = async () => {
         txToSign: cosmosUnsignedTx.txToSign
       })
       console.log('broadcastedTx:', broadcastedTx)
-
     } catch (err) {
       console.log('cosmosTx error:', err.message)
     }
--- a/packages/chain-adapters/src/ChainAdapterCLI.ts
+++ b/packages/chain-adapters/src/ChainAdapterCLI.ts
@@ -188,21 +188,30 @@ const main = async () => {
 
     // send cosmos example
     try {
+      const value = '100'
+
+      const feeData = await cosmosChainAdapter.getFeeData({ sendMax: false })
+      const fee = feeData.slow.txFee
+      const gas = feeData.slow.chainSpecific.gasLimit
+
       const cosmosUnsignedTx = await cosmosChainAdapter.buildSendTransaction({
-        to: `0x47CB53752e5dc0A972440dA127DCA9FBA6C2Ab6F`,
-        value: '1',
+        to: 'cosmos1j26n3mjpwx4f7zz65tzq3mygcr74wp7kcwcner',
+        value,
         wallet,
         bip44Params: cosmosBip44Params,
-        chainSpecific: { gas: '0' }
+        chainSpecific: { gas, fee }
       })
-      const cosmosSignedTx = await cosmosChainAdapter.signTransaction({
+
+      if (!cosmosChainAdapter.signAndBroadcastTransaction) return
+
+      console.log('comsos unsigned tx', cosmosUnsignedTx)
+
+      const broadcastedTx = await cosmosChainAdapter.signAndBroadcastTransaction({
         wallet,
         txToSign: cosmosUnsignedTx.txToSign
       })
-      console.log('cosmosSignedTx:', cosmosSignedTx)
+      console.log('broadcastedTx:', broadcastedTx)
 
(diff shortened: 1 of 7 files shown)


packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -134,26 +134,53 @@ export class ChainAdapter
...
       return ErrorHandler(err)
     }
   }
 
-  async getFeeData(
-    /* eslint-disable-next-line @typescript-eslint/no-unused-vars -- Disable no-unused-vars lint rule for unimplemented methods */
-    input: Partial<chainAdapters.GetFeeDataInput<ChainTypes.Cosmos>>
-  ): Promise<chainAdapters.FeeDataEstimate<ChainTypes.Cosmos>> {
-    throw new Error('Method not implemented.')
+  async getFeeData({
```
DaoDev44, line 146: Hard coding the fees here because we don't have a way to dynamically get fee data without sending a singed transaction first. This would break the current send flow by making the user sign a transaction first, then get the fee data, then sign again with the correct fee data. The fee values are arbitrary, but we wanted to keep the fees low and similar to other wallet fees. Any feedback is welcome.

0xdef1cafe: @DaoDev44 @willyogo @DiggyDiggy2 how do other cosmos apps approach this? is there really no way to get fee data without a signed message?

DaoDev44: @0xdef1cafe After discussing with @kaladinlight, this is probably the best path in the sort term, however, unchained can be modified to take similar `getFeeData` arguments as we do with ETH and serialize it to a format that the gaia node can accept, but that might be a bigger lift and can be done in the second round of cosmos improvements.

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -188,21 +188,29 @@ const main = async () => {
...
+        to: 'cosmos1j26n3mjpwx4f7zz65tzq3mygcr74wp7kcwcner',
+        value,
         wallet,
         bip44Params: cosmosBip44Params,
-        chainSpecific: { gas: '0' }
+        chainSpecific: { gas, fee }
       })
-      const cosmosSignedTx = await cosmosChainAdapter.signTransaction({
+
+      if (!cosmosChainAdapter.signAndBroadcastTransaction) return
```
asamere, line 205: is there ever a case where this function isn't defined? extra check doesn't hurt anything but wondering if I'm missing something

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -72,7 +72,8 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implement
         }),
         chain: this.getType(),
         chainSpecific: {
-          sequence: data.sequence
+          accountNumber: data.accountNumber.toString(),
```
asamere, line 75: have you tested this with a cosmos account that has never transacted? watch out for undefined accountNumber for new accounts. possible it's an empty string in api so no harm

packages/chain-adapters/src/ChainAdapterCLI.ts
DaoDev44: This function is only used for wallets that cannot sign offline (ie metamask), so technically there could be a time where it isn't defined.

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
DaoDev44: No I haven't tested that use case. I'll test that.

0xdef1cafe (approved):
i'm ok to get this in, so we can get the whole stack tested, and follow up with bugfix PRs if necessary if you are @asamere

asamere (approved):

gomesalexandre (approved):
LGTM 🍬 ! Tested the payload in https://github.com/shapeshift/web/pull/1100 and the TXs are sent.

Issue #444, opened 2022-03-16
Remove `ChainTypes`/`NetworkType` in favor of consistent use of CAIP2s/CAIP19s


mrnerdhair:
(I've done some research into this; see #445.)

PR #445, opened 2022-03-16
research: ChainTypes/NetworkType/CAIP2/CAIP19 refactor & cleanup

mrnerdhair:
This is the result of a research spike I did trying to determine what a solution to #444 might look like. It builds and passes tests, but the commit history needs some work and it would require matching changes in `web`. (I did enough of those to have some confidence that the changes here can indeed be ported over to web successfully, but didn't finish them.)


mrnerdhair:
(Whoops! Sorry, meant to close this once it was linked to #444.)

PR #452, opened 2022-03-17
fix: make accountType optional in BitcoinChainAdapter getAddress

gomesalexandre:
This follows https://github.com/shapeshift/lib/pull/447 - we have a default Segwit Native `accountType`, so it shouldn't be required to explicitly pass it.


0xdef1cafe (approved):
in general i prefer explicit over implicit but this is better than before

PR #455, opened 2022-03-18
feat: implement Foxy asset & market data

toshiSat:


packages/asset-service/src/generateAssetData/ethTokens/foxy.ts
```diff
@@ -0,0 +1,32 @@
...
+    name: 'FOXy',
+    precision: 18,
+    tokenId: assetReference,
+    contractType: assetNamespace,
+    color: '#CE3885',
+    secondaryColor: '#CE3885',
+    icon: 'https://raw.githubusercontent.com/shapeshift/lib/main/packages/asset-service/src/icons/foxy-icon.png',
+    sendSupport: true,
+    receiveSupport: true,
+    symbol: 'FOXY'
```
0xdef1cafe, line 28:
```suggestion
    symbol: 'FOXy'
```

0xdef1cafe: not sure what we want the symbol here? check w/ erik?

packages/market-service/src/foxy/foxy.ts
```diff
@@ -0,0 +1,133 @@
...
+import { CoinCapMarketCap } from '../coincap/coincap-types'
+import { RATE_LIMIT_THRESHOLDS_PER_MINUTE } from '../config'
+import { bn } from '../utils/bignumber'
+import { isValidDate } from '../utils/isValidDate'
+import { rateLimitedAxios } from '../utils/rateLimiters'
+
+export const FOXY_CAIP19 = 'eip155:1/erc20:0x61fcabb591d63d00e897a67c64658d376fead816'
+const FOX_COINCAP_ID = 'fox-token'
+
+const axios = rateLimitedAxios(RATE_LIMIT_THRESHOLDS_PER_MINUTE.COINGECKO)
```
0xdef1cafe, line 21: wrong rate limit - coincap not coingecko

packages/market-service/src/foxy/foxy.ts
```diff
@@ -0,0 +1,133 @@
...
+import dayjs from 'dayjs'
+
+import { MarketService } from '../api'
+import { CoinCapMarketCap } from '../coincap/coincap-types'
+import { RATE_LIMIT_THRESHOLDS_PER_MINUTE } from '../config'
+import { bn } from '../utils/bignumber'
+import { isValidDate } from '../utils/isValidDate'
+import { rateLimitedAxios } from '../utils/rateLimiters'
+
+export const FOXY_CAIP19 = 'eip155:1/erc20:0x61fcabb591d63d00e897a67c64658d376fead816'
```
0xdef1cafe, line 18: would be nice to get this from a single source of truth, e.g. where we generate assets, rather than risk this contract getting redeployed and missing this

packages/market-service/src/foxy/foxy.ts
```diff
@@ -0,0 +1,133 @@
...
+      return { [caip19]: marketData } as MarketCapResult
+    } catch (e) {
+      console.warn(e)
+      return {}
+    }
+  }
+
+  findByCaip19 = async ({ caip19 }: MarketDataArgs): Promise<MarketData | null> => {
+    try {
+      if (caip19.toLowerCase() !== FOXY_CAIP19.toLowerCase()) return null
```
0xdef1cafe, line 40: log here before returning null

packages/market-service/src/foxy/foxy.ts
```diff
@@ -0,0 +1,133 @@
...
+      console.warn(e)
+      throw new Error('FoxyMarketService(findByCaip19): error fetching market data')
+    }
+  }
+
+  findPriceHistoryByCaip19 = async ({
+    caip19,
+    timeframe
+  }: PriceHistoryArgs): Promise<HistoryData[]> => {
+    if (caip19.toLowerCase() !== FOXY_CAIP19.toLowerCase()) return []
```
0xdef1cafe, line 61: log here before returning []

packages/market-service/src/foxy/foxy.ts
```diff
@@ -0,0 +1,133 @@
...
+import { bn } from '../utils/bignumber'
+import { isValidDate } from '../utils/isValidDate'
+import { rateLimitedAxios } from '../utils/rateLimiters'
+
+export const FOXY_CAIP19 = 'eip155:1/erc20:0x61fcabb591d63d00e897a67c64658d376fead816'
+const FOX_COINCAP_ID = 'fox-token'
+
+const axios = rateLimitedAxios(RATE_LIMIT_THRESHOLDS_PER_MINUTE.COINGECKO)
+
+export class FoxyMarketService implements MarketService {
```
0xdef1cafe, line 23: wait, why do we even need this, if coincap is the adapter backing this?

just update the coincap caip adapters to point the foxy caip19 at the asset id for the underlying token?

0xdef1cafe: as discussed:

this is valid over just the caip adapter idea, please comment heavily that this is only a stop gap until we get coingecko/coincap upstream sources, at which point this is to be deleted.

packages/market-service/src/foxy/foxy.ts
0xdef1cafe: use `getFoxyToken`

0xdef1cafe (changes requested):
do the things

packages/market-service/src/foxy/foxy.ts
toshiSat: should we have the market-service use asset-service as a peer dependency just for this?

0xdef1cafe (approved):

PR #468, opened 2022-03-22
chore: add error handling

DaoDev44:
- Add error handling to `buildSendTransaction`

### DIFF
--- a/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
@@ -102,7 +102,7 @@ export class ChainAdapter
         try {
           const val = bnOrZero(account.balance).minus(gas)
           if (val.isFinite() || val.lte(0)) {
-            throw new Error('CosmosChainAdapter: transaction value is invalid')
+            throw new Error(`CosmosChainAdapter: transaction value is invalid: ${val.toString()}`)
           }
           tx.value = val.toString()
         } catch (error) {
--- a/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
@@ -99,7 +99,15 @@ export class ChainAdapter
       const account = await this.getAccount(from)
 
       if (sendMax) {
-        tx.value = bnOrZero(account.balance).minus(gas).toString()
+        try {
+          const val = bnOrZero(balance).minus(gas)
+          if (!isFinite(val.toNumber()) || val.toNumber() < 0) {
+            throw new Error('CosmosChainAdapter: transaction value is invalid')
+          }
+          tx.value = val.toString()
+        } catch (error) {
+          return ErrorHandler(error)
+        }
       }
 
       const utx: CosmosTx = {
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -98,14 +98,22 @@ export class ChainAdapter
       const account = await this.getAccount(from)
 
       if (sendMax) {
-        tx.value = bnOrZero(account.balance).minus(gas).toString()
+        try {
+          const val = bnOrZero(account.balance).minus(gas)
+          if (!isFinite(val.toNumber()) || val.toNumber() < 0) {
```
0xdef1cafe, line 103: you can use val.isFinite() and val.lte(0) here

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -98,14 +98,22 @@ export class ChainAdapter
...
+          return ErrorHandler(error)
+        }
       }
 
       const utx: CosmosTx = {
         fee: {
           amount: [
             {
-              amount: bnOrZero(fee).toString(),
+              amount: String(bnOrZero(fee)),
```
0xdef1cafe, line 116: shouldn't we be calling .toString() with some precision stuff instead?

DaoDev44: The precision is handled on the frontend I believe, so I don't think we need to specify that. The `String` vs `toString` can go either way for me.

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -98,14 +98,22 @@ export class ChainAdapter
       const account = await this.getAccount(from)
 
       if (sendMax) {
-        tx.value = bnOrZero(account.balance).minus(gas).toString()
+        try {
+          const val = bnOrZero(account.balance).minus(gas)
+          if (val.isFinite() || val.lte(0)) {
+            throw new Error('CosmosChainAdapter: transaction value is invalid')
```
gomesalexandre, line 104: Shouldn't we log the actual wrong tx value?
```suggestion
            throw new Error(`CosmosChainAdapter: transaction value is invalid: ${val.toString()}`)
```

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -134,8 +142,8 @@ export class ChainAdapter
         addressNList,
         tx: utx,
         chain_id: caip2.ChainReference.CosmosHubMainnet,
-        account_number: account.chainSpecific.accountNumber,
-        sequence: account.chainSpecific.sequence
+        account_number: String(account.chainSpecific.accountNumber),
+        sequence: String(account.chainSpecific.sequence)
```
gomesalexandre, line 146: According to typings, these should already come as a string so no need to cast them

gomesalexandre (approved):
One small comment regarding what looks like a casting we don't need, otherwise looks good

gomesalexandre (approved):

0xdef1cafe (approved):

PR #470, opened 2022-03-22
fix: foxy icon url

toshiSat:
Update icon url.  This was unable to be tested due to needing to be in main before github hosted it.
This is the correct url


NeOMakinG (approved):
Seems legit

0xdef1cafe (approved):

PR #477, opened 2022-03-23
feat: label ibc voucher assets as Asset on Osmosis

0xdef1cafe:


0xdef1cafe:
yolo merging as we've demoed this working locally

PR #479, opened 2022-03-24
feat: return staking details for cosmos type accounts

kaladinlight:


packages/types/src/chain-adapters/cosmos.ts
```diff
@@ -3,9 +3,51 @@ export type BuildTxInput = {
...
+  amount: string
+}
+
 export type Account = {
   sequence: string
   accountNumber: string
+  delegations?: Array<Delegation>
+  redelegations?: Array<Redelegation>
+  undelegations?: Array<Undelegation>
+  rewards?: Array<Reward>
```
theoboldfrazier, line 50: Can we make these required and if there isn't any just return an empty array?

Makes it easier down the line to only deal with an array type instead of `[] | undefined`

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -62,22 +63,65 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implement
   async getAccount(pubkey: string): Promise<chainAdapters.Account<T>> {
     try {
       const caip = this.getCaip2()
-      const { chain, network } = caip2.fromCAIP2(caip)
       const { data } = await this.providers.http.getAccount({ pubkey })
 
+      const delegations = (
+        data.delegations.length ? data.delegations : undefined
```
theoboldfrazier, line 69: Rather than dealing with a ternary here, if we require the type on the cosmos account, we can just do the map, and if there is no length, then it'll set it to an empty array.

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -62,22 +63,57 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implement
   async getAccount(pubkey: string): Promise<chainAdapters.Account<T>> {
     try {
       const caip = this.getCaip2()
-      const { chain, network } = caip2.fromCAIP2(caip)
       const { data } = await this.providers.http.getAccount({ pubkey })
 
+      const delegations = data.delegations.map<chainAdapters.cosmos.Delegation>((v) => ({
```
gomesalexandre, line 68: oh no 
```suggestion
      const delegations = data.delegations.map<chainAdapters.cosmos.Delegation>(delegation => ({
```

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -62,22 +63,57 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implement
...
+          address: v.sourceValidator
+        },
+        entries: v.entries.map<chainAdapters.cosmos.RedelegationEntry>((e) => ({
+          assetId: this.assetId,
+          completionTime: Number(e.completionTime),
+          amount: e.balance
+        }))
+      }))
+
+      const undelegations = data.unbondings.map<chainAdapters.cosmos.Undelegation>((v) => ({
```
gomesalexandre, line 90: ditto

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -62,22 +63,57 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosChainTypes> implement
...
-      const { chain, network } = caip2.fromCAIP2(caip)
       const { data } = await this.providers.http.getAccount({ pubkey })
 
+      const delegations = data.delegations.map<chainAdapters.cosmos.Delegation>((v) => ({
+        assetId: this.assetId,
+        amount: v.balance.amount,
+        validator: {
+          address: v.validator
+        }
+      }))
```
cjthompson, line 74: I would prefer these transform functions be moved into separate functions so make them more testable and reusable.

```ts
const transformDelegations = (delegations: chainAdapters.cosmos.Delegation) => ({
assetId: this.assetId,
        amount: v.balance.amount,
        validator: {
          address: v.validator
        }
      }))

const delegations = data.delegations.map<chainAdapters.cosmos.Delegation>(transformDelegations)
```

kaladinlight: Not sure it provides a ton of value, but I can. It is already reusable in the sense it exists in the base adapter, and testing `getAccount` seems sufficient to me once we actually add tests.

0xdef1cafe (approved):

PR #481, opened 2022-03-24
fix: fix yearn sdk instantiation

0xdef1cafe:
`disableAllowList` was removed from the `Yearn()` constructor and is now defaulted - see upstream issue here https://github.com/yearn/yearn-sdk/issues/242

this fixes the build for CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "version": "independent"
   },
   "scripts": {
+    "local-ci": "yarn clean && yarn && yarn lint && yarn build && yarn type-check && yarn test",
     "build": "lerna run build --include-dependencies",
     "clean": "rimraf node_modules dist packages/**/dist packages/**/node_modules",
     "dev": "lerna run dev --stream --parallel",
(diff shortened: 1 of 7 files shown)


theoboldfrazier (approved):
Nice, the only reason we added that in was because they required it in the constructor.

package.json
```diff
@@ -9,6 +9,7 @@
     "version": "independent"
   },
   "scripts": {
+    "local-ci": "yarn clean && yarn && yarn lint && yarn build && yarn type-check && yarn test",
```
theoboldfrazier, line 12: Do we want to link-packages here too, since PRs may be working across packages?

Could be `yarn clean-link && yarn lint && yarn type-check && yarn test`

0xdef1cafe: good thought but not quite - `yarn link-packages` in `lib` is just setting the packages up so they can be consumed by `web`.

`yarn clean` breaks all the links, and reflects the state the CI is in locally, so we don't need to `clean-link` here

Issue #473, opened 2022-03-22
Fiat to fiat market service spike


0xdef1cafe:
### Overview

This is a spike/research ticket for @stackedq as a prerequisite to implementing a currency switcher in web.

This is necessary because we can't get good historical data for crypto assets priced in other currencies. We can however, get reliable forex data for the USD -> arbitrary fiat currencies.

We can't use use the spot price historically because the forex markets move over time.

We also need to consider a free reliable source of data that can be consumed from the client without an API key, and consider rate limits.

### Reference

- Top 10 fiat currencies by market cap? https://fiatmarketcap.com/ product to confirm
- see current fiat market service calls here https://github.com/shapeshift/web/blob/develop/src/state/slices/marketDataSlice/marketDataSlice.tsx#L80 and https://github.com/shapeshift/web/blob/develop/src/state/slices/marketDataSlice/marketDataSlice.tsx#L96
- maybe look at google finance?

### Acceptance Criteria

- confirm with product that top 10 currencies is acceptable
- find a reliable source of spot and historical market data for these currencies that doesn't require an API key and realistically won't get users rate limited on the client
- ensure we can get similar historical data granularity as the crypto market services can provide
- `findByFiatSymbol` - takes a string union of `CNY | EUR | JPY | ...` for the top 10 fiat assets and returns a `MarketData` type
- `findPriceHistoryByFiatSymbol` - takes the same input as `findByFiatSymbol` but returns `HistoryData` - just a price and timestamp
- stub out the service with some basic tests to prove this will work

### Need By Date

n/a

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 2800.0 FOX (775.54 USD @ $0.28/FOX)  attached to it as part of the ShapeShift fund.__

[...]

PR #495, opened 2022-03-27
feat: fiat to fiat

stackedq:
Fiat to fiat implementation, 
gets fiat latest and historical rates for `web`.
closes: #473


packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
```diff
@@ -0,0 +1,103 @@
...
+}: FiatMarketDataArgs): Promise<MarketData | null> => {
+  try {
+    const { data } = await axios.get<ExchangeRateHostRate>(
+      `${baseUrl}/latest?base=${baseCurrency}&symbols=${symbol}`
+    )
+    const result = data as ExchangeRateHostRate
+    // we only need the price key in the `web`
+    return {
+      price: result.rates[symbol].toString(),
+      marketCap: '',
```
DaoDev44, line 31: Wonder if marketCap and volume should be `'0'` instead of empty string.

packages/types/src/market.ts
```diff
@@ -40,3 +40,35 @@ export type FindAllMarketArgs = {
 export type MarketCapResult = {
   [k: string]: MarketData
 }
+
+export enum SupportedFiatCurrencies {
```
DaoDev44, line 44: Based on the requirements, `SupportedFiatCurrencies` should be in string union form ie, `'USD' | 'EUR'`

packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
```diff
@@ -0,0 +1,103 @@
...
+const baseUrl = 'https://api.exchangerate.host'
+
+export const findByFiatSymbol = async ({
+  symbol
+}: FiatMarketDataArgs): Promise<MarketData | null> => {
+  try {
+    const { data } = await axios.get<ExchangeRateHostRate>(
+      `${baseUrl}/latest?base=${baseCurrency}&symbols=${symbol}`
+    )
+    const result = data as ExchangeRateHostRate
```
gomesalexandre, line 27: Wouldn't the generic in `axios.get()` assume the `data` type if you check for the truthiness of `data` to narrow it down?

packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+export const findPriceHistoryByFiatSymbol = async ({
+  symbol,
+  timeframe
+}: FiatPriceHistoryArgs): Promise<HistoryData[]> => {
+  const end = dayjs().startOf('day')
+  let start
+  switch (timeframe) {
```
gomesalexandre, line 46: This is a bit boilerplaty but after playing with TSPlayground trying to get the enum keys and a default, I couldn't come up with a better way to achieve this.

packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
stackedq: @gomesalexandre You're absolutely right

packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
```diff
@@ -0,0 +1,103 @@
...
+    console.warn(e)
+    throw new Error('FiatMarketService(findByFiatSymbol): error fetching market data')
+  }
+}
+
+export const findPriceHistoryByFiatSymbol = async ({
+  symbol,
+  timeframe
+}: FiatPriceHistoryArgs): Promise<HistoryData[]> => {
+  const end = dayjs().startOf('day')
```
gomesalexandre, line 44: `By default, Day.js parses and displays in local time.` - can you confirm that the UTC timestamp we pass to`exchangerate.host` will return correct data on different browser/system times? Got burned with UTC offsets in the past so just double checking

gomesalexandre (commented):
Added one concern regarding the UTC timestamp that we send to `exchangerate.host` but other than that this LGTM!

packages/market-service/src/exchange-rates-host/exchange-rates-host.ts
stackedq: this is using the local time. dayjs supports timezones with a plugin though, should that be installed and integrated?

stackedq: IMO though local time is the way to go, because we're using these information for the client.

stackedq: well, actually it may return wrong information because it's using the YYYY-MM-DD format instead of a timestamp...

stackedq: there's like 0.1% difference between `exchangerate.host` and [Bitterex](https://www.tradingview.com/chart/?symbol=BITTREX%3AUSDEUR) candles daily close results.

DaoDev44 (approved):

elmutt (approved):

packages/types/src/market.ts
```diff
@@ -40,3 +40,43 @@ export type FindAllMarketArgs = {
...
   [k: string]: MarketData
 }
+
+/**
+ * stackedQ: We need the list in the frontend,
+ * and there's no way to convert an union type to an array,
+ * so the list is defined as an array of strings,
+ * but SupportedFiatCurrencies is an union of strings
+ */
+export const SupportedFiatCurrenciesList = [
```
0xdef1cafe, line 50: we don't actually need this list - if we define a string union, the compiler is smart enough to autocomplete which strings are valid types

packages/market-service/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './market-service-manager'
+export * from './exchange-rates-host/exchange-rates-host'
```
0xdef1cafe, line 2: i do wonder if this should be part of the market service api, much like the crypto services are, so we can have multiple backing sources to avoid rate limiting/downtime. 

what do you think?

0xdef1cafe (changes requested):
don't need string array, and question about abstraction here.

packages/types/src/market.ts
stackedq: This list is used to render the supported fiat currencies in the "currency" section of the settings modal. I couldn't find a way to convert a string union type to an array, that's why this is defined this way.

packages/market-service/src/index.ts
stackedq: Actually, in my first approach, I wrote a similar market-service-manager, but only for the fiat-rate providers, it seemed a little bit too much because we only got one provider right now, but I'm ok if you want it to be implemented that way for easier future provider integrations.

packages/types/src/market.ts
0xdef1cafe: ah ok, i see now, and defining the types from the string array as const is also the way to go :)

packages/market-service/src/index.ts
0xdef1cafe: @stackedq after discussing with @cjthompson lets please go ahead and create this abstraction, in the same way the market service currently does it to source crypto prices, such that we can add redundancy/extra capability transparently in future.

stackedq: done

packages/market-service/src/fiat-market-service.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      expect(result).toBeNull()
+    })
+  })
+
+  describe('findPriceHistoryByFiatSymbol', () => {
+    const args = {
+      symbol: <const>'EUR',
+      timeframe: HistoryTimeframe.HOUR
+    }
+    it('can return from fist market service and skip the next', async () => {
```
0xdef1cafe, line 38:
```suggestion
    it('can return from first market service and skip the next', async () => {
```
```suggestion
    it('can return from fist market service and skip the next', async () => {
```

0xdef1cafe, line 38:
```suggestion
    it('can return from first market service and skip the next', async () => {
```
```suggestion
    it('can return from fist market service and skip the next', async () => {
```

0xdef1cafe, line 38:
```suggestion
    it('can return from first market service and skip the next', async () => {
```

packages/market-service/src/fiat-market-service.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      const result = await findPriceHistoryByFiatSymbol(args)
+      expect(result).toEqual(mockERHPriceHistoryData)
+    })
+    it('can return null if no data found', async () => {
```
0xdef1cafe, line 42:
```suggestion
    it('can return empty array if no data found', async () => {
```

0xdef1cafe, line 42:
```suggestion
    it('can return empty array if no data found', async () => {
```

packages/market-service/src/exchange-rates-host/exchange-rates-host.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    it('should return fiat market data for EUR', async () => {
+      mockedAxios.get.mockResolvedValue({ data: eurRate })
+      expect(await exchangeRateHostService.findByFiatSymbol(args)).toEqual(mockERHFindByFiatSymbol)
+    })
+
+    it('should return null on network error', async () => {
+      mockedAxios.get.mockRejectedValue(Error)
+      jest.spyOn(console, 'warn').mockImplementation(() => void 0)
```
0xdef1cafe, line 32: make sure to restore this

0xdef1cafe, line 32: make sure to restore this

packages/market-service/src/exchange-rates-host/exchange-rates-host.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      mockedAxios.get.mockResolvedValue({ data: mockHistoryData })
+      expect(await exchangeRateHostService.findPriceHistoryByFiatSymbol(args)).toEqual(
+        mockERHPriceHistoryData
+      )
+    })
+
+    it('should return null on network error', async () => {
+      mockedAxios.get.mockRejectedValue(Error)
+      jest.spyOn(console, 'warn').mockImplementation(() => void 0)
```
0xdef1cafe, line 63: make sure to restore this

0xdef1cafe, line 63: make sure to restore this

0xdef1cafe (changes requested):
few tiny ones but otherwise looking good now - @cjthompson would appreciate another review here too please

packages/types/src/market.ts
```diff
@@ -40,3 +40,47 @@ export type FindAllMarketArgs = {
...
```
cjthompson, line 50:
```suggestion
export const SupportedFiatCurrenciesList = Object.freeze([
```

packages/types/src/market.ts
```diff
@@ -40,3 +40,47 @@ export type FindAllMarketArgs = {
...
+}
+
+export type FiatPriceHistoryArgs = {
+  symbol: SupportedFiatCurrencies
+  timeframe: HistoryTimeframe
+}
+
+export type FindByFiatMarketType = (args: FiatMarketDataArgs) => Promise<MarketData | null>
+
+export type FiatPriceHistoryType = (args: FiatPriceHistoryArgs) => Promise<HistoryData[]>
```
cjthompson, line 86: I don't think any of these need to be in the `types` package.  The only types that should be in `types` are ones that are used in multiple lib packages to avoid circular dependencies.

These types look to be very `market-service` specific and aren't imported into any other packages.

Move these to the `market-service` package.

packages/market-service/src/fiat-market-service.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  ExchangeRateHostService: jest.fn().mockImplementation(() => {
+    return {
+      findByFiatSymbol: jest.fn(() => mockERHFindByFiatSymbol),
+      findPriceHistoryByFiatSymbol: jest.fn(() => mockERHPriceHistoryData)
+    }
+  })
+}))
+
+describe('fiat market service', () => {
+  describe('findByFiatSymbol', () => {
```
cjthompson, line 17: I'd like to see more tests, especially covering error conditions.

cjthompson (approved):
approved with comments

packages/market-service/src/fiat-market-service.test.ts
stackedq: added a test case for handling `NaN` price from history API. useless validate date check removed.

PR #497, opened 2022-03-27
fix: foxy rebase logic and types

0xdef1cafe:
this is required to unblock https://github.com/shapeshift/web/pull/1321

* **wip - @toshiSat to complete - handle stake events that trigger a rebase**
* include assetId in rebase history return type
* adjust the return type to be called blockTime in alignment with what web consumes
* adjust rebase balance diffs that are initiated by unstake events to exclude the transfer value


0xdef1cafe:
reviewed with @theoboldfrazier and @pastaghost

theoboldfrazier (approved):

PR #502, opened 2022-03-28
feat: scrape tokemak for apr

toshiSat:
Just a way to temporarily get a more accurate apr for Tokemak

From the Tokemak discord when asked about an API for getting this information

Today at 3:26 PM
we're going to stand up an API for it -- aiming to push this in the next few weeks

### DIFF
--- a/packages/investor-foxy/package.json
+++ b/packages/investor-foxy/package.json
@@ -29,6 +29,7 @@
     "@ethersproject/providers": "^5.5.3",
     "bignumber.js": "^9.0.2",
     "lodash": "^4.17.21",
+    "puppeteer": "^13.5.1",
     "web3": "^1.7.1",
     "web3-core": "^1.7.1",
     "web3-utils": "^1.7.1"
--- a/packages/investor-foxy/src/api/api.ts
+++ b/packages/investor-foxy/src/api/api.ts
@@ -4,6 +4,7 @@ import { ChainAdapter } from '@shapeshiftoss/chain-adapters'
 import { ChainTypes } from '@shapeshiftoss/types'
 import { BigNumber } from 'bignumber.js'
 import { toLower } from 'lodash'
+import puppeteer from 'puppeteer'
 import Web3 from 'web3'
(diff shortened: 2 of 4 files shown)


NeOMakinG:
Interesting! is it worth implementing that puppeteer trick if they plan to release an API in the next few weeks?

mrnerdhair:
I don't think this will work; `puppeteer` drives a headless copy of Chrome, which we definitely can't launch from inside a webpage. (It might work in `jest` because that's a node environment, but I'll eat my hat if it works in the browser.)

toshiSat:
This is basically just a proof of concept, but mrnerdhair is right, it won't work. 
I've confirmed with product that it's ok if we keep estimating the apr for now and either 1 of three things will happen:

1.  Tokemak releases apy's on an api and we just call that or..
2.  Within 2-3 weeks we will start getting reward info.  I can then grab the circulatingSupply + what the avg rewards per day along with the price of TOKE and the price of FOX and calculate that way.  
3.  We are hosting a bot to batch withdrawal requests in case someone doesn't call a correct function in time.  
I could store the api scraping there info there, but in general, web scraping is gross and having a bot return on single piece of info for the system seems gross too

Either way, we've decided to not implement this just yet and we should have a follow up solution sometime in April.

toshiSat:
This is basically just a proof of concept, but mrnerdhair is right, it won't work. 
I've confirmed with product that it's ok if we keep estimating the apr for now and either 1 of three things will happen:

1.  Tokemak releases apy's on an api and we just call that or..
2.  Within 2-3 weeks we will start getting reward info.  I can then grab the circulatingSupply + what the avg rewards per day along with the price of TOKE and the price of FOX and calculate that way.  
3.  We are hosting a bot to batch withdrawal requests in case someone doesn't call a correct function in time.  
I could store the api scraping there info there, but in general, web scraping is gross and having a bot return on single piece of info for the system seems gross too

Either way, we've decided to not implement this just yet and we should have a follow up solution sometime in April.

PR #503, opened 2022-03-28
chore: update all instances of foxy addresses

toshiSat:


DaoDev44 (approved):

packages/market-service/src/foxy/foxy.ts
```diff
@@ -15,7 +15,7 @@ import { bn } from '../utils/bignumber'
 import { isValidDate } from '../utils/isValidDate'
 import { rateLimitedAxios } from '../utils/rateLimiters'
 
-export const FOXY_CAIP19 = 'eip155:1/erc20:0x61fcabb591d63d00e897a67c64658d376fead816'
+export const FOXY_CAIP19 = 'eip155:1/erc20:0xDc49108ce5C57bc3408c3A5E95F3d864eC386Ed3'
```
0xdef1cafe, line 18: let's use this as the single source of truth and import it in the tests

0xdef1cafe (changes requested):

0xdef1cafe (approved):

PR #507, opened 2022-04-01
fix: bignumber error

toshiSat:
Bignumber.js and web3's BN don't play nice when in a production optimized build

Also, send date with getTimeUntilClaimable


0xdef1cafe (approved):
Get in, nice catch

reallybeard (approved):

PR #511, opened 2022-04-05
Fix time to claim

toshiSat:


DaoDev44 (approved):

packages/investor-foxy/src/api/api.ts
```diff
@@ -903,12 +903,14 @@ export class FoxyApi {
     } catch (e) {
       throw new Error(`Failed to get block number: ${e}`)
     }
-    const epochsLeft = coolDownInfo.endEpoch - epoch.number - 1 // epochs left after the current one
+    const epochsLeft = bnOrZero(coolDownInfo.endEpoch).minus(epoch.number) // epochs left until can claim
```
asamere, line 906: curious why endEpoch can be undefined?

toshiSat: It can't, it can only be zero, I figured it was fine using that though

asamere: oh probably because they've reached the epoch making them eligible to claim?

asamere (approved):
Looks good 👍

PR #513, opened 2022-04-06
fix: add KICK to the blacklist

NeOMakinG:
## Context

The Kick token is showing on the `web` app:
image

We need to add the token to the blacklist, because it was missing, it's a scam token.

## Result
image

## Warning

It seems that just rebuilding the `generatedAssetData.json` file made the job, probably because `yearnfinance` removed the asset, but I would prefer to blacklist it in case the asset is inside another datasource in the future.


0xdef1cafe:
@NeOMakinG can you please open this PR from a branch on this repo so one of the core team members can run the asset generation script on our side?

NeOMakinG:
https://github.com/shapeshift/lib/pull/524

PR #517, opened 2022-04-11
fix(chain-adapters): improperly declared async validators

GMSteuart:


gomesalexandre:
@GMSteuart It can be tempting at first glance to make these methods synchronous functions as they effectively don't do anything asynchronous, but it was actually an engineering decision to try and use async methods for consistency and flexibility: https://github.com/shapeshift/lib/pull/296#discussion_r781399182

gomesalexandre:
Thanks for the great comments, definitely agree with all these points! Without any context, there is no reason why anyone seeing this code wouldn't jump on it to make it synchronous. Comments would be a great way to convey it.

> @gomesalexandre Making these synchronous methods asynchronous has added additional complexity to the frontend that would not have been implemented if the functions were marked as synchronous

Is there any bottlenecks you've encountered from these being asynchronous? With current usages, we use these methods in send forms [`validate` method](https://github.com/shapeshift/web/blob/60ffe3ffaa2e2198c72ad475dabbdcaa88d65819/src/components/Modals/Send/views/Address.tsx#L86) which can - or not - be made async so it should be just about async/awaiting it.

> Additionally, when development decisions like this are made, commenting the code acknowledging this intention is greatly beneficial; adding a link to the issue or comment thread helps too.

Agreed! I think we could add it on top of the typings to make sure these functions stay synchronous, what do you think? Would they also be helpful in actual usages of these methods?
> 
> Will a validator here ever actually be asynchronous?

They currently don't, but they could be in the future, e.g offloading the validation task to a webworker, using an external service if bech32/WAValidator validation doesn't cut it for future supported chains for any reason

GMSteuart:
@gomesalexandre Thank you for the reply. I always enjoy a good technical discussion.

> They currently don't, but they could be in the future, e.g offloading the validation task to a webworker, using an external service if bech32/WAValidator validation doesn't cut it for future supported chains for any reason

I understand the desire to implement code that is both extensible and future proof but the results of this decision are anything but that. There can exist both asynchronous validators and synchronous validators. I'd ask how one effectively argues being more flexible when making a decision that is clearly pigeon holing? Additionally, async validation at this point is purely speculation.

Please take a look at lines 88-90 in [Address.tsx](https://github.com/shapeshift/web/blob/0417b57520ee2cc492e4c0bc5033dd38d5cebffa/src/components/Modals/Send/views/Address.tsx#L88) file in the web repo. How much time passes from the boolean flag swapping state? How many other parts of this component are written because of this state change? Is this considered flexible and consistent?

> Agreed! I think we could add it on top of the typings to make sure these functions stay synchronous, what do you think? Would they also be helpful in actual usages of these methods?

I don't see any reason for the async versions of these methods to exist in the code base at this time. Make a comment saying the async methods exist in the file history and if needed those methods will be like a phoenix when the time comes.

elmutt:
I think it makes sense to remove the promises as well. 

IMO making them async so they all "look" the same doesnt really make sense and adds to overall code compleity while potentially misleading an engineer about the inner workings of the functions

elmutt (approved):

0xdef1cafe:
relevant discussion https://github.com/shapeshift/web/discussions/1474

@GMSteuart if you make a corresponding draft PR for web to update the consumption of these validators, we can approve and merge this.

cjthompson:
> Additionally, async validation at this point is purely speculation.

Hit the nail on the head.

cjthompson (approved):

0xdef1cafe:
@GMSteuart converting this to draft until we have a corresponding web PR ready to go, as this is a breaking change.

mrnerdhair:
Note that this change will lock us into loading `multicoin-address-validator` and all its deps synchronously on page load.

cjthompson:
> Note that this change will lock us into loading `multicoin-address-validator` and all its deps synchronously on page load. (Don't know what the performance impact of that is, but as you all know I'm wary of including synchronous methods in the API, that's one reason.)

`async` functions add complexity to React code. For example, `useEffect` doesn't support async functions so IIFEs are required and that can lead to errors like calling `setState` after a component is unmounted (since `setState` call is in a Promise).

There are practical tradeoffs here that need to be considered.

gomesalexandre:
@0xdef1cafe do we want to stamp this one after your latest comment in https://github.com/shapeshift/web/discussions/1474 ?

0xdef1cafe:
@GMSteuart do you want to bring this out of draft and make an associated web PR?

PR #531, opened 2022-04-18
test(market-service): spy on console methods to fix log spew in tests

### LINKED ISSUE #530: Fix log spew
### Overview

As a follow-up to [web issue #1293](https://github.com/shapeshift/web/issues/1293), and to quote @0xdef1cafe, "We have a bunch of log spew in CI from jest and it's heaps gross aye."

Check and existing CI run for an example

https://github.com/shapeshift/lib/runs/6050265393?check_suite_focus=true

### Acceptance Criteria

- `yarn test` runs without any rank log spew

### Need By Date

N/A

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xf11nix:
Closes #530 

`yarn test` output (with `--coverage=false` to squeeze relevant data into a readable screenshot)

Screen Shot 2022-04-18 at 10 50 25 pm

### DIFF
--- a/packages/market-service/src/coincap/coincap.test.ts
+++ b/packages/market-service/src/coincap/coincap.test.ts
@@ -143,6 +143,7 @@ describe('coincap market service', () => {
       await expect(coinMarketService.findByCaip19(args)).rejects.toEqual(
         new Error('MarketService(findByCaip19): error fetching market data')
       )
+      expect(consoleSpy).toHaveBeenCalledTimes(1)
       consoleSpy.mockRestore()
     })
   })
@@ -177,6 +178,7 @@ describe('coincap market service', () => {
       await expect(coinMarketService.findPriceHistoryByCaip19(args)).rejects.toEqual(
         new Error('MarketService(findPriceHistoryByCaip19): error fetching price history')
       )
+      expect(consoleSpy).toHaveBeenCalledTimes(1)
       consoleSpy.mockRestore()
     })
   })
--- a/packages/market-service/src/coincap/coincap.test.ts
+++ b/packages/market-service/src/coincap/coincap.test.ts
@@ -139,10 +139,11 @@ describe('coincap market service', () => {
 
     it('should return null on network error', async () => {
       mockedAxios.get.mockRejectedValue(Error)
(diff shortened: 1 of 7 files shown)


0xdef1cafe:
nice - did you consider using a `beforeEach` and `afterEach` in the `describe` blocks, and just have the `expect...` in each test?

0xf11nix:
Thanks for the review @0xdef1cafe. 
My apologies, I definitely should have made note of this in my initial submission.

It appears that any spy that is registered in the before hook, is done so in the "global " scope of the file. And as a result, we end up with some very odd behavior. 

For example, if we were to set `consoleSpy = jest.spyOn(console, 'warn').mockImplementation(() => void 0)` in the before hook, but only call `mockRestore` it in a single test (say, the first test), all of the following tests would still behave as if `mockRestore` had never been called. This struck me as odd and suggested that the spy was not _actually_ block scoped, as it appeared (and documented, albeit, poorly).

From what I have observed, setting the spy in the the before* hook will cause that override to persist for all tests in that one file, regardless of whether `mockRestore` is called in the afterEach hook, or anywhere else.

I had noticed that this pattern had already existed in the `logger` package, so I had just assumed that this "global state pollution" was a known issue/constraint of Jest.

From my existing observations, it would seem that if we weren't fussed about making sure we were unloading the spy override, we could just add `consoleSpy = jest.spyOn(console, 'warn').mockImplementation(() => void 0)` to the beforeEach hook, and forget the `mockRestore` all together (since it doesn't seem to actually do anything in this context). Though, I don't think that's actually what we are trying to achieve.

Given this additional context (that I absolutely should have included previously), how would we like to proceed? I'm happy to keep digging to try and find and more "min diff" solution if that's what we'd prefer.

cjthompson:
https://github.com/bpedersen/jest-mock-console

This PR solves the problem for those specific tests but doesn't prevent it from happening in any future tests added. I would suggest looking at this npm module and just mocking out console for the entire test suite.

There's no need for us to call `expect.toHaveBeenCalled` on console logs because it's not an important part of the function and we may change loggers in the future.

PR #537, opened 2022-04-21
feat: add `code` to errors

cjthompson:
This feature will allow us to support translations for error messages.

* The `code` property of an error should be a short string that could be used to lookup a translation for what error message should be shown to a user.
* Error codes are normalized to be UPPER_CASE+SNAKE_CASE so we can have some consistency. For example, `ERR_RATE_LIMIT`.
* All errors have a default code that matches the error's name.

### DIFF
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -29,6 +29,17 @@ try {
 
 ```
 
+### Error.code
+
+All errors support a `code` property. This property is designed to be used for internationalization
+so that translated text can be displayed based on the `code` rather than on the `message`.
+
+```ts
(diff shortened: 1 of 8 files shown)


0xApotheosis (approved):
I like it.

I'd prefer to see `ERROR` instead of `ERR` in all instances, but that's just stylistic.

gomesalexandre (approved):
LGTM.

cjthompson:
> I like it.
> 
> I'd prefer to see `ERROR` instead of `ERR` in all instances, but that's just stylistic.

I was just following how [node.js does their errors](https://nodejs.org/api/errors.html#nodejs-error-codes).

PR #526, opened 2022-04-15
feat(swapper): add getByPair

DaoDev44:
- Add `isSupportedAssets` to SwapperManager.
- This is needed to chose swappers by supported assets instead of hard coding the swapper to use.

Closes [Add logic to get swapper by pair](https://github.com/shapeshift/lib/issues/523) #523

PR #542, opened 2022-04-22
chore: remove unused swapper methods

DaoDev44:
**NOTE: This is technically a breaking change, however, the methods being removed are not used anywhere in our codebase, so it is a "safer" breaking change.** 

- Remove `canTradePair` and `getAvailableAssets` from Swapper. 
- As a part of the Swapper refactor work, we have created methods that will replace these methods ([PR](https://github.com/shapeshift/lib/pull/526)) that improve on the naming and implementation. As we move away from using `ChainTypes` and move more to strictly using CAIP19s, removing these methods will bring us closer to that goal.

### DIFF
--- a/packages/swapper/src/api.ts
+++ b/packages/swapper/src/api.ts
@@ -33,16 +33,6 @@ export interface Swapper {
   getQuote(input: GetQuoteInput, wallet?: HDWallet): Promise<Quote<ChainTypes, SwapperType>>
 
   /**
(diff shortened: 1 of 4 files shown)


gomesalexandre (approved):
LGTM! Searched for these two methods across the org codebases and that removes the references correctly, indeed they aren't consumed by any repo currently 👍🏽

cjthompson (approved):

elmutt (approved):

PR #545, opened 2022-04-25
feat(logger): support custom logging output function

cjthompson:
* feat(logger): support custom logging ouput function
* feat(logger): support formatting Errors with cause, details, and code
* chore(logger): update README to describe custom output logging functions and add some comments

## Details
You can now pass in `logFn` to the `Logger` constructor to provide a custom log output function. The function is provided with the log level and a `FormattedObject` of data.

### DIFF
--- a/packages/logger/src/format.ts
+++ b/packages/logger/src/format.ts
@@ -1,4 +1,4 @@
-import { FormattedObject } from './logger.type'
+import { FormattedError, FormattedObject } from './logger.type'
 
 const objectProto = Object.getPrototypeOf(Object())
 function isObject(value: unknown): value is Record<string, unknown> {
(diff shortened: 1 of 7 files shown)


packages/logger/src/format.ts
```diff
@@ -20,19 +20,41 @@ function isError(value: unknown): value is Error {
...
+/**
+ * Support errors with additional properties
+ * @shapeshiftoss/errors implement these additional properties
+ */
+function isErrorWithDetails(value: unknown): value is Error & {
+  code?: string
+  details?: Record<string, unknown>
+  cause: unknown
+} {
+  return isError(value) && 'cause' in value
```
0xApotheosis, line 32: Should this be checking for `details`?

```suggestion
  return isError(value) && 'details' in value
```

packages/logger/src/logger.type.ts
```diff
@@ -47,10 +73,32 @@ export interface FormattedError {
...
+  status: LogLevel
+}
+
+/**
+ * A function responsible for processing log data
+ */
+export type LoggerFunction = (
+  level: LogLevel.TRACE | LogLevel.DEBUG | LogLevel.INFO | LogLevel.WARN | LogLevel.ERROR,
+  data: LogData
+) => void
```
0xApotheosis, line 104:
```suggestion
export type LoggerFunction = (level: Exclude<LogLevel, LogLevel.NONE>, data: LogData) => void
```

0xApotheosis (approved):
Looks good, a couple of thoughts.

packages/logger/src/format.ts
cjthompson: No because details is optional but `cause` always exists even if the value is undefined.

packages/logger/src/logger.type.ts
gomesalexandre: https://www.typescriptlang.org/docs/handbook/utility-types.html

gomesalexandre (approved):

PR #546, opened 2022-04-25
feat(caip): banxa chain map

stackedq:
this PR adds the map of converting our chains to the Banxa blockchain name.

web PR: https://github.com/shapeshift/web/pull/1478
this is where this function is used: https://github.com/shapeshift/web/pull/1478/files#diff-d11d8939921408eeb21dd7f5115e6b0ee21aeb464708b178b08c36ba5aecd12eR34
the input for this function is the selectedAssetTicker from Banxa list, this is the list:
https://github.com/shapeshift/lib/blob/b0dd979413f400db08bb360992148d49a4728478/packages/caip/src/adapters/banxa/index.ts#L4-L33

follow-up of: https://github.com/shapeshift/web/issues/1387

### DIFF
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -1,7 +1,7 @@
 import entries from 'lodash/entries'
 import toLower from 'lodash/toLower'
 
-import { CAIP2, toCAIP2 } from '../../caip2/caip2'
+import { ChainId, toCAIP2 } from '../../caip2/caip2'
 import { fromCAIP19 } from '../../caip19/caip19'
 
 const CAIP19ToBanxaTickerMap = {
@@ -56,11 +56,11 @@ export const getSupportedBanxaAssets = () =>
  * since some Banxa assets could be on multiple chains and their default
  * chain won't be exactly the same as ours.
  */
-const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
+const chainIdToBanxaBlockchainCodeMap: Record<ChainId, string> = {
   'eip155:1': 'ETH',
   'bip122:000000000019d6689c085ae165831e93': 'BTC',
   'cosmos:cosmoshub-4': 'COSMOS'
-}
+} as const
 
 /**
  * Convert a banxa asset identifier to a Banxa chain identifier for use in Banxa HTTP URLs
@@ -73,6 +73,6 @@ export const getBanxaBlockchainFromBanxaAssetTicker = (banxaAssetId: string): st
   if (!assetCAIP19)
     throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${banxaAssetId} is not supported`)
   const { chain, network } = fromCAIP19(assetCAIP19)
-  const caip2 = toCAIP2({ network, chain })
-  return caip2ToBanxaBlockchainCodeMap[caip2]
+  const chainId = toCAIP2({ network, chain })
+  return chainIdToBanxaBlockchainCodeMap[chainId]
 }
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -59,7 +59,8 @@ const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
 
 export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
   const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
-  if (!assetCAIP19) return ''
+  if (!assetCAIP19)
+    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${asset} is not supported`)
   const { chain, network } = fromCAIP19(assetCAIP19)
   const caip2 = toCAIP2({ network, chain })
   return caip2ToBanxaBlockchainCodeMap[caip2]
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -51,16 +51,27 @@ export const getSupportedBanxaAssets = () =>
     ticker
   }))
 
+/**
+ * map CAIP2s to Banxa blockchain codes (ETH, BTC, COSMOS),
+ * since some Banxa assets could be on multiple chains and their default
+ * chain won't be exactly the same as ours.
+ */
 const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
   'eip155:1': 'ETH',
   'bip122:000000000019d6689c085ae165831e93': 'BTC',
   'cosmos:cosmoshub-4': 'COSMOS'
 }
 
-export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
-  const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
+/**
+ * Convert a banxa asset identifier to a Banxa chain identifier for use in Banxa HTTP URLs
+ *
+ * @param {string} banxaAssetId - a Banxa asset string referencing a specific asset; e.g., 'atom'
+ * @returns {string} - a Banxa chain identifier; e.g., 'cosmos'
+ */
+export const getBanxaBlockchainFromBanxaAssetTicker = (banxaAssetId: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(banxaAssetId.toLowerCase())
   if (!assetCAIP19)
-    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${asset} is not supported`)
+    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${banxaAssetId} is not supported`)
   const { chain, network } = fromCAIP19(assetCAIP19)
   const caip2 = toCAIP2({ network, chain })
   return caip2ToBanxaBlockchainCodeMap[caip2]
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -1,5 +1,8 @@
+import { ChainTypes } from '@shapeshiftoss/types'
 import entries from 'lodash/entries'
 import toLower from 'lodash/toLower'
+
+import { fromCAIP19 } from '../../caip19/caip19'
 
 const CAIP19ToBanxaTickerMap = {
   'bip122:000000000019d6689c085ae165831e93/slip44:0': 'btc',
@@ -47,3 +50,17 @@ export const getSupportedBanxaAssets = () =>
     CAIP19,
     ticker
   }))
+
+const caip10TpBanxaBlockchainMap: Record<ChainTypes, string> = {
+  [ChainTypes.Ethereum]: 'ETH',
+  [ChainTypes.Bitcoin]: 'BTC',
+  [ChainTypes.Cosmos]: 'COSMOS',
+  [ChainTypes.Osmosis]: ''
+}
+
+export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
+  if (!assetCAIP19) return ''
+  const { chain } = fromCAIP19(assetCAIP19)
+  return caip10TpBanxaBlockchainMap[chain]
+}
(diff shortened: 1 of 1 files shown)


packages/caip/src/adapters/banxa/index.ts
```diff
@@ -47,3 +50,17 @@ export const getSupportedBanxaAssets = () =>
...
+
+const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
+  'eip155:1': 'ETH',
+  'bip122:000000000019d6689c085ae165831e93': 'BTC',
+  'cosmos:cosmoshub-4': 'COSMOS'
+}
+
+export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
+  if (!assetCAIP19) return ''
```
gomesalexandre, line 62: I'd rather throw here

gomesalexandre (approved):

packages/caip/src/adapters/banxa/index.ts
stackedq: @gomesalexandre is this good:
```
if (!assetCAIP19)
    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${asset} is not supported`)
 ```

Comment (gomesalexandre) [reply packages/caip/src/adapters/banxa/index.ts:62]: Perfect Ser

File: packages/caip/src/adapters/banxa/index.ts [hunk 2]
```diff
@@ -47,3 +50,18 @@ export const getSupportedBanxaAssets = () =>
...
+}
+
+export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
+  if (!assetCAIP19)
+    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${asset} is not supported`)
+  const { chain, network } = fromCAIP19(assetCAIP19)
+  const caip2 = toCAIP2({ network, chain })
+  return caip2ToBanxaBlockchainCodeMap[caip2]
+}
```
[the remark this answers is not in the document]
cjthompson: In general, I don't like `get` (or `is`) functions to `throw` when something isn't found. I'd rather just return `undefined`. I assume this is following a pattern from another file in this package. If that's the case, I get it.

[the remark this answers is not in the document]
stackedq: @cjthompson that `throw` part is suggested by @gomesalexandre in his review.
this function gets banxaAssetTicker which they're values of `CAIP19ToBanxaTickerMap`,
so the first step is to convert Banxa asset ticker to caip19,
then get chain and network from caip19,
then create caip2 with those variables, and find the related banxa blockchain_code from `caip2ToBanxaBlockchainCodeMap`.
only "btc", "eth" and "cosmos" work with your suggested implementation.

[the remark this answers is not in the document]
cjthompson: > only "btc", "eth" and "cosmos" work with your suggested implementation.

those values are the only acceptable inputs in your implementation also. I don't see the difference.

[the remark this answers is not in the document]
stackedq: any Banxa ticker can be passed to this function, this is the list of the tickers: https://github.com/shapeshift/lib/blob/b0dd979413f400db08bb360992148d49a4728478/packages/caip/src/adapters/banxa/index.ts#L4-L33

cjthompson:
Can you link to the issue / ticket for this PR so I can get context for the requirements?

stackedq:
> Can you link to the issue / ticket for this PR so I can get context for the requirements?

@cjthompson This is the web PR: https://github.com/shapeshift/web/pull/1478
this is where this function is used: https://github.com/shapeshift/web/pull/1478/files#diff-d11d8939921408eeb21dd7f5115e6b0ee21aeb464708b178b08c36ba5aecd12eR34
the input for this function is the selectedAsset from banxa list, this is the list:
https://github.com/shapeshift/lib/blob/b0dd979413f400db08bb360992148d49a4728478/packages/caip/src/adapters/banxa/index.ts#L4-L33

gomesalexandre:
@stackedq Could you add the list of supported assets in description, either the response from `https://api.banxa.com/api/coins` or any page that lists them?
Checked again what we support and we support all of them, but it would be great to double-check against the full list (I suppose this one is just the intersection of what they support and what we support)

Also, as we've discussed on Discord, what about using IDs vs. tickers? The order API states the following for the `target` in the JSON payload:
```
target
string
required

Target currency or cryptocurrency code. This parameter indicates whether the order is a buy or a sell cryptocurrency order.
```

It should be possible to use `coin_code` which should be different from the ticker.

gomesalexandre:
> > Can you link to the issue / ticket for this PR so I can get context for the requirements?
> 
> @cjthompson This is the web PR: 

Can you link https://github.com/shapeshift/web/issues/1387 as an issue in description?

packages/caip/src/adapters/banxa/index.ts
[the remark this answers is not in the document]
gomesalexandre: > any Banxa ticker can be passed to this function, this is the list of the tickers:

`as const` would probably help here. It narrows down the typing of this object to these specific keys/values primitives (strings) vs. any string as key, any string as value as we have currently. If we go that route, we'll need a type guard as well, since with the web logic, we will try all of our supported assetIds against `CAIP19ToBanxaTickerMap` which would produce type errors in case of an assetId that's not in this object.

Something like:
```typescript
const isValidBanxaAssetId = (assetId: string): assetId is keyof typeof CAIP19ToBanxaTickerMap => Boolean(CAIP19ToBanxaTickerMap[assetId as keyof typeof CAIP19ToBanxaTickerMap])
```
([TS Playground demo](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhhAwuaB5ARgKwKbFgXhgG8AoGcmAMxBAC4YBydOAJwYBoyLmAvehgI4BXAB4MSAX3gQYoSFBKK50GAEsIANTgAbVQBMAgsGDYIEECxiEAFHGOnzLetBaqwAcwCU9OybMW1GQBrbABPEEoYKFCAB2wI6WR5DBw8KwA+GAAhGm1sODBbJBQoFNwoAG1fBwCEGBDwyOi4hIQktCxygF1PRRJlWHUAMRotXT0rQLH9Iz9HawZqEAZe5RA8gDptEHdrImHRnX0JVZLAkaEWaIAVAHcQaYnCdUfZmpYF6kub+5X+lHW2C2Oz2B2+oTuDyOehOQA))

If we have such safer access, then we wouldn't need to throw anymore. Type-level safety is probably a better option, thanks @cjthompson!

stackedq:
@gomesalexandre `https://api.banxa.com/api/coins` needs API_KEY and secret and we don't got them.
this is the list (although it looks like it has been edited): https://docs.google.com/spreadsheets/d/1KU6J1Hl4vxBTIWCwWdrFfSadNltuFheQRoJyPrd0LMQ/edit#gid=631982242
as far as I know `coin_code` is the coin ticker, I can't find any docs on them though.

gomesalexandre:
> @gomesalexandre `https://api.banxa.com/api/coins` needs API_KEY and secret and we don't got them. this is the list (although it looks like it has been edited): https://docs.google.com/spreadsheets/d/1KU6J1Hl4vxBTIWCwWdrFfSadNltuFheQRoJyPrd0LMQ/edit#gid=631982242 as far as I know `coin_code` is the coin ticker, I can't find any docs on them though.

We should get in touch with Banxa to get the relevant IDs as well. We have the mapping for a reason (reliability towards CAIP19s and lookup supported assets to make the intersection) and if we end up using tickers as values, it doesn't solve the first point anymore.

decentralized-j:
gm sers! just chiming in here around these points. @stackedq Banxa assured that we do not need API_KEY or secret to integrate - all that is needed is to embed the referral link (https://shapeshift.banxa.com/) to a button, further details are here including ability to pass thru more info from the front end https://docs.banxa.com/docs/referral-method

I have just reached out to Banxa for ID mapping, will update here asap

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -47,3 +50,18 @@ export const getSupportedBanxaAssets = () =>
...
   }))
+
+const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
+  'eip155:1': 'ETH',
+  'bip122:000000000019d6689c085ae165831e93': 'BTC',
+  'cosmos:cosmoshub-4': 'COSMOS'
+}
+
+export const getBanxaBlockchainFromBanxaAssetTicker = (asset: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(asset.toLowerCase())
```
cjthompson, line 61:
```suggestion
```suggestion
/**
 * Convert a banxa asset identifier to a Banxa chain identifier for use in Banxa HTTP URLs
 *
 * @param {string} banxaAssetId - a Banxa asset string referencing a specific asset; e.g., 'atom'
 * @returns {string} - a Banxa chain identifier; e.g., 'cosmos'
 */
export const getBanxaBlockchainFromBanxaAssetTicker = (banxaAssetId: string): string => {
  const assetCAIP19 = banxaTickerToCAIP19(banxaAssetId.toLowerCase())
```

gomesalexandre:
As we've discussed on Discord, Banxa has confirmed us they use tickers, not IDs. The map has two purposes in Gem:
1. Provide safe access to the intersection of assets the ramp supports and the ones we support
2. Protect against them rugging us with changing the ticker, by using IDs and having a mapping(caip19 -> id)

While the second point doesn't apply anymore, there is still value in going with this map to ensure safe access IMO.

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -47,3 +50,29 @@ export const getSupportedBanxaAssets = () =>
     CAIP19,
     ticker
   }))
+
+/**
+ * map CAIP2s to Banxa blockchain codes (ETH, BTC, COSMOS),
+ * since some Banxa assets could be on multiple chains and their default
+ * chain won't be exactly the same as ours.
+ */
+const caip2ToBanxaBlockchainCodeMap: Record<CAIP2, string> = {
```
gomesalexandre, line 59: We might even want to `as const` this if this is compatible with the `Record<CAIP2, string>` type

gomesalexandre (dismissed):
LGTM, CI is happy both on remote and locally which makes sense given the fact that this is a newly added domain

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -1,6 +1,9 @@
 import entries from 'lodash/entries'
 import toLower from 'lodash/toLower'
 
+import { CAIP2, toCAIP2 } from '../../caip2/caip2'
```
0xdef1cafe, line 4: can we please update this to use `ChainId` over `CAIP2`?

packages/caip/src/adapters/banxa/index.ts
0xdef1cafe: agree with as const, and please rename to `chainIdToBanxaBlockchainCodeMap`

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -47,3 +50,29 @@ export const getSupportedBanxaAssets = () =>
...
+ * @param {string} banxaAssetId - a Banxa asset string referencing a specific asset; e.g., 'atom'
+ * @returns {string} - a Banxa chain identifier; e.g., 'cosmos'
+ */
+export const getBanxaBlockchainFromBanxaAssetTicker = (banxaAssetId: string): string => {
+  const assetCAIP19 = banxaTickerToCAIP19(banxaAssetId.toLowerCase())
+  if (!assetCAIP19)
+    throw new Error(`getBanxaBlockchainFromBanxaAssetTicker: ${banxaAssetId} is not supported`)
+  const { chain, network } = fromCAIP19(assetCAIP19)
+  const caip2 = toCAIP2({ network, chain })
+  return caip2ToBanxaBlockchainCodeMap[caip2]
```
0xdef1cafe, line 77:
```suggestion
  const chainId = toCAIP2({ network, chain })
  return chainIdToBanxaBlockchainCodeMap[chainId]
```

0xdef1cafe (changes requested):
if you make these updates i'll approve

0xdef1cafe (approved):

cjthompson (approved):

PR #560, opened 2022-04-26
feat(caip): flatten exports

### LINKED ISSUE #548: Flatten exports from caip packages
### Overview

When first implemented we namespaced all the fucntions under caip2 or caip19 - let's flatten these out for ergonomics

https://github.com/shapeshift/lib/blob/main/packages/caip/src/index.ts#L2

### Acceptance Criteria

Functions like `toCAIP19` or `fromCAIP19` can be imported directly from the caip packages.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

gomesalexandre:
This flattens the caip exports instead of exporting each of the namespaces as their own module export. 

Following the release of this, https://github.com/shapeshift/web/pull/1591 should be merged, updating the usages.

- partially tackles #548

### DIFF
--- a/packages/asset-service/src/generateAssetData/ethTokens/ethereum.ts
+++ b/packages/asset-service/src/generateAssetData/ethTokens/ethereum.ts
@@ -1,4 +1,4 @@
-import { caip19 } from '@shapeshiftoss/caip'
+import { fromCAIP19 } from '@shapeshiftoss/caip'
 import { BaseAsset, TokenAsset } from '@shapeshiftoss/types'
 import axios from 'axios'
 import chunk from 'lodash/chunk'
@@ -43,7 +43,7 @@ export const addTokensToEth = async (): Promise<BaseAsset> => {
   for (const [i, batch] of tokenBatches.entries()) {
     console.info(`processing batch ${i + 1} of ${tokenBatches.length}`)
     const promises = batch.map(async (token) => {
-      const { chain } = caip19.fromCAIP19(token.assetId)
+      const { chain } = fromCAIP19(token.assetId)
       const { info } = generateTrustWalletUrl({ chain, tokenId: token.tokenId })
       return axios.head(info) // return promise
     })
@@ -70,7 +70,7 @@ export const addTokensToEth = async (): Promise<BaseAsset> => {
         }
         return uniqueTokens[key] // token without modified icon
       } else {
-        const { chain } = caip19.fromCAIP19(uniqueTokens[key].assetId)
+        const { chain } = fromCAIP19(uniqueTokens[key].assetId)
         const { icon } = generateTrustWalletUrl({ chain, tokenId: uniqueTokens[key].tokenId })
         return { ...uniqueTokens[key], icon }
       }
--- a/packages/asset-service/src/generateAssetData/ethTokens/foxy.ts
+++ b/packages/asset-service/src/generateAssetData/ethTokens/foxy.ts
@@ -1,4 +1,4 @@
-import { AssetNamespace, caip2, caip19 } from '@shapeshiftoss/caip'
+import { AssetNamespace, toCAIP2, toCAIP19 } from '@shapeshiftoss/caip'
 import { AssetDataSource, ChainTypes, NetworkTypes, TokenAsset } from '@shapeshiftoss/types'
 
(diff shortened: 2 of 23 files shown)


0xdef1cafe:
@gomesalexandre this should be a major version bump as it's a breaking API change, and given this is the *only* CAIP implementation i know of, we should be cognizant if others are using this

0xApotheosis (approved):
🍬

0xApotheosis:
All changes appear to continue existing behavior 👍  agree re: major version.

gomesalexandre:
> @gomesalexandre this should be a major version bump as it's a breaking API change, and given this is the _only_ CAIP implementation i know of, we should be cognizant if others are using this

Good idea, this is indeed a breaking change - updated the package.json to reflect this, and also lerna bootstrapped all the other packages using the previous CAIP major to use 3.0.0.

cjthompson:
> > @gomesalexandre this should be a major version bump as it's a breaking API change, and given this is the _only_ CAIP implementation i know of, we should be cognizant if others are using this
> 
> Good idea, this is indeed a breaking change - updated the package.json to reflect this, and also lerna bootstrapped all the other packages using the previous CAIP major to use 3.0.0.

Don't edit the version number manually. semantic release process will bump it when given the right commit messages.

cjthompson:
This should be 2 PRs due to how semantic release works. 1 PR needs to be just for the `caip` package with the `BREAKING CHANGE` note in the commit messages.

Then a second PR should update the imports in the other packages.

gomesalexandre:
> https://blog.greenkeeper.io/introduction-to-semantic-release-33f73b117c8

Thanks a lot for that. Updated commit header/body to match what's needed for the semantic release to trigger a feature release.

cjthompson:
This PR needs to be updated to remove all changes not inside "caip" so that we don't cause major version bumps on other packages.

gomesalexandre:
> This PR needs to be updated to remove all changes not inside "caip" so that we don't cause major version bumps on other packages.

Done. Created https://github.com/shapeshift/lib/pull/565 for it, will open it after new caip version is published on npm.

cjthompson:
> > This PR needs to be updated to remove all changes not inside "caip" so that we don't cause major version bumps on other packages.
> 
> Done. Created #565 for it, will open it after new caip version is published on npm.
> 
> However, build is faling for the other packages since lerna links all the packages inside the monorepo together. Is it OK to have the CI fail in this particular case? Will it still publish caip changes only?

Hmm... I'm not sure about that.

gomesalexandre:
> Hmm... I'm not sure about that.

Let's give it a try but my understanding is CI would fail since the `yarn build` effectively fails. Might have to revert it to the previous changes including the usages update on the other packages as well.

elmutt (approved):

0xApotheosis:
> > Hmm... I'm not sure about that.
> 
> Edit2: Went with @cjthompson suggestion of manually bumping the version in the package.json of caip ([0cc5626#diff-ae32105bdd75af136c77218eca5f677270ea3efd92c13a2b2e33d97a2c90ac3dR3](https://github.com/shapeshift/lib/commit/0cc5626edcd3369f286cf02dc71107dd6202b9c7#diff-ae32105bdd75af136c77218eca5f677270ea3efd92c13a2b2e33d97a2c90ac3dR3)) and manually referencing it in every package ([8a80f32](https://github.com/shapeshift/lib/commit/8a80f326ab19e2d8341dbcb09cce010d6ac858e6)). Used `chore:` as a commit message/PR title so this should trigger a minor release for all except caip. ~Edit: Brought it back. This will fail at `yarn build` - in theory semantic release would be able to bump a single package depending on the affected files, but we'll be effectively blocked from doing so at build step.~ ~Let's give it a try but my understanding is CI would fail since the `yarn build` effectively fails. Might have to revert it to the previous changes including the usages update on the other packages as well.~

@gomesalexandre don't do this - it does not work, as I learned the hard way.

The semantic bot will override your manual bump, and because it's a chore it'll be a patch release (`BREAKING CHANGE` in the commit footer gets ignored).

You'll need to use `perf` - an example is in https://github.com/shapeshift/lib/pull/617.

0xdef1cafe (approved):

elmutt (approved):

PR #563, opened 2022-04-26
fix(chain-adapters): don't set amount on MsgWithdrawDelegationReward

mrnerdhair:
Including a zero-valued field here breaks with amino signers (like KK).

### DIFF
--- a/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
@@ -350,10 +350,6 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<ChainTypes.Cosmos> {
           {
             type: 'cosmos-sdk/MsgWithdrawDelegationReward',
             value: {
-              amount: {
-                amount: bnOrZero(0).toString(), // amount here is required to broadcast, but we are withdrawing ALL rewards
-                denom: 'uatom'
-              },
               delegator_address: from,
               validator_address: validator
             }


packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
```diff
@@ -350,10 +350,6 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<ChainTypes.Cosmos> {
           {
             type: 'cosmos-sdk/MsgWithdrawDelegationReward',
             value: {
-              amount: {
-                amount: bnOrZero(0).toString(), // amount here is required to broadcast, but we are withdrawing ALL rewards
```
cjthompson, line 354: So is the comment here wrong or was `hdwallet` fixed so that this isn't needed anymore?

cjthompson (approved):

0xdef1cafe (approved):
confirmed that at least this broadcasts a valid claim payload as per @DaoDev44 with a native wallet locally - need to confirm we can claim with KK

packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts
0xdef1cafe: @cjthompson see my comment below, @DaoDev44 was able to claim with native so at least the payload is valid.

mrnerdhair: I think this was required by `proto-tx-builder` at the time of writing; it's not anymore in `^0.2.1`.

PR #569, opened 2022-04-27
chore: remove swappertype from quote fee type

elmutt:
Theres no reason to have the SwapperType generic. 

It was added prematuraly with an assumption that thorchain would need it. This simplifies lots of things


0xdef1cafe:
as discussed, a potential fee abstraction (in your future PRs) could look like the following

```
export type TradeFees = {
  broadcastFee: {
    /**
     * e.g. in the case of a THORChain ETH -> BTC swap,
     * the broadcastFee is taken from the sellAsset, in this case the broadcastFee is paid in ETH
     */
    feeAssetId: AssetId
    fee: string // base units, e.g. wei or sats
  }
  tradeFee: {
    /**
     * e.g. in the case of a THORChain ETH -> BTC swap,
     * the trade fee is taken from the buy asset, in this case the tradeFee is paid in BTC
     */
    tradeFeeId: AssetId
    fee: string // base units, e.g. wei or sats
  }
}
```

DaoDev44 (approved):

cjthompson:
This is a breaking change and needs to have `BREAKING CHANGE` in the squash commit message if it's merged.

elmutt:
> This is a breaking change and needs to have `BREAKING CHANGE` in the squash commit message if it's merged.

Will do!

cjthompson (approved):

PR #582, opened 2022-04-28
fix: remove getmaxamount unnecessary code

elmutt:
It does not make sense to calculate max trade amount in the swapper. 

Doing so keeps us tightly coupled to chain adapters and swapper should not care about the users balance

All of this code can be removed once the following web PR is merged:

https://github.com/shapeshift/web/pull/1626


0xdef1cafe (approved):

gomesalexandre (approved):
LGTM
- grepped for all occurences of `(send)?maxAmount" and there are none remaining
- all references used by the removed code are removed from imports
- CI is very happy

PR #584, opened 2022-04-29
fix: blacklist 0xeee burn address

0xdef1cafe:
we had two ETH assets - one was picking up the 0xeee burn address which we don't want to display in
the UI


NeOMakinG (approved):
Seems fine

The fact that some new assets added to every API we use are also generated at the same time feels weird and makes the review harder but I guess we can't do anything about it

cjthompson (approved):

PR #586, opened 2022-04-29
fix: removes 3 unnecessary api calls from zrxswapper. 

elmutt:
Removes 3 unnecessary API calls in the ZrxSwapper. 

- ZrxApproveInfinite doesnt need to make another call to the /quote api. It is already available on the existing quote
- ZrxApprovalNeeded doesnt need to make another call to the /quote api. It is already available on the existing quote
- minimumPrice in getZrxMinMax isnt used anywhere and can be entirely removed

Should speed things up


packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
```diff
@@ -14,31 +13,16 @@ export const getZrxMinMax = async (
...
-      sellToken,
-      buyToken,
-      sellAmount: normalizeAmount(minimumWeiAmount.toString())
-    }
-  })
-  const minimumPrice = new BigNumber(minimumPriceResult?.data?.price).toString()
 
   return {
     minimum,
-    minimumPrice,
```
elmutt, line 41: This is entirely unused by web or anywhere so we can just remove it

GMSteuart:
@0xdef1cafe should this pull request have a commit including `BREAKING CHANGE:` that affects a `@shapeshiftoss/types` file?

theoboldfrazier (approved):

DaoDev44 (approved):

PR #588, opened 2022-05-02
fix: unbreak lib build

gomesalexandre:
This unbreaks the `yarn build` step, which is currently broken in latest develop since https://github.com/shapeshift/lib/commit/e02a2f4046e7a734f6e92d5ec448234a17818092:

image

- update `@shapeshiftoss/types` to latest where needed
- add missing `tokens` property in `transformValidator` return type
- narrow down the adapter type to be `ChainTypes.Ethereum` vs. `ChainTypes` currently in `investor-foxy`. This gives it a proper `T` generic type that's passed down to various methods and solves many type errors. This also gives it typing parity with `investor-yearn`, where the typing is actually the correct narrowed down one:
https://github.com/shapeshift/lib/blob/bf5aa33d3f3396ae652e618d3e1d5e4440711792/packages/investor-yearn/src/api/api.ts#L32
https://github.com/shapeshift/lib/blob/bf5aa33d3f3396ae652e618d3e1d5e4440711792/packages/investor-yearn/src/api/api.ts#L40
- add `ChainAdapter<ChainTypes.Ethereum>` type casting to adapter property in `FoxyApi` args. Until we narrow down the types in `chain-adapters` `byChainId`, that's probably the easiest way to go 
- bump `@shapeshiftoss/chain-adapters` to ^2.13.0 in `investor-yearn` for version parity with `investor-foxy` 
- import correct, non-namespaced path in `packages/market-service/src/market-service-manager.ts`:  `import { FindAllMarketArgs } from '@shapeshiftoss/types/src'` -> `import { FindAllMarketArgs } from '@shapeshiftoss/types'`
https://github.com/shapeshift/lib/blob/597dc6ed2a8367656204ba2687cd31466a963e61/packages/market-service/src/market-service-manager.ts#L9

### DIFF
--- a/packages/asset-service/package.json
+++ b/packages/asset-service/package.json
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@ethersproject/providers": "^5.5.2",
     "@shapeshiftoss/caip": "^2.0.0",
-    "@shapeshiftoss/types": "^2.1.0",
+    "@shapeshiftoss/types": "^3.0.0",
     "@types/lodash": "^4.14.172",
     "@yfi/sdk": "^1.0.30",
     "dotenv": "^14.3.0",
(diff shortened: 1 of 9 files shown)


gomesalexandre:
@cjthompson Since this bumps deps to new major one, do we want this to be squashed to a `BREAKING CHANGE: ` body?

cjthompson (approved):

0xdef1cafe (approved):

Issue #579, opened 2022-04-28
Independent Submodule Versioning


GMSteuart:
### Overview

As a developer, I want the ability to independently version the packages in this `lib` repository to have more control over the publishing process. Additionally, each submodule should be able to build separately without causing issues. Submodules should only be version bumped if a code or file change occurred in the package's directory.

### References and additional details

- `package.json` files will be useful
- semantic-release
- semantic-release-monorepo

### Acceptance Criteria

- [ ] All submodules successfully build
- [ ] Independent submodule versioning and publishing
- [ ] Implement test script for release using a local package manager

### Estimated effort

One day

GMSteuart:
Missed opportunity to put @0xdef1cafe and @majorhayes under "References and additional details"

0xean:
closing per @0xdef1cafe

PR #589, opened 2022-05-02
chore: add e2e release test

GMSteuart:
Closes issue #579 by implementing a continuous integration test script. Script will install a [Verdaccio](https://verdaccio.org/) locally for pushing and pulling all packages; this is done via `npx` to reduce project install time. When script finishes, everything is cleaned up.

### DIFF
--- a/.verdaccio/config.yaml
+++ b/.verdaccio/config.yaml
@@ -1,0 +1,41 @@
+# path to a directory with all packages
+storage: ../build/local-registry/storage
+
+auth:
+  htpasswd:
+    file: ./htpasswd
+
+# a list of other known repositories we can talk to
+uplinks:
+  npmjs:
+    url: https://registry.npmjs.org/
+    cache: true
+  yarn:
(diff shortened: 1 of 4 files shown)


elmutt (approved):

theoboldfrazier (approved):
Looks good. Tested out locally and ran well. Nice work.

0xdef1cafe:
@GMSteuart do we still need this? @0xApotheosis and @kaladinlight were able to publish major versions of packages yesterday without this.

GMSteuart:
@0xdef1cafe this PR is the code I used to test that our release process works as intended. Merging it wont have any side effects, it lays the foundation for locally testing package releases for full e2e testing, and can be used as a reference for running a registry locally.

cjthompson:
It doesn't seem like this does what the issue is requesting. This does the release process to a local repository instead of npm but it doesn't fix the issue of major version bumps.

cjthompson (approved):

PR #595, opened 2022-05-03
chore: remove more unnecessary swapper code

elmutt:
- Slippage doesnt make sense coming back from the quote (it should be determined by a user selection on the frontend)
- Slippage doesnt make sense passing into getQuote because its not neecded until buildQuoteTx
- It doesnt make sense to pass minimum into getQuote. ZrxSwapper now determines its own minimum

Removes unnecessary types & related code

Tested & working on linked locally to web. Slippage & minimum aren't actually being passed in anywhere from web so this is safe


packages/swapper/src/swappers/zrx/getZrxQuote/getZrxQuote.ts
```diff
@@ -33,12 +27,11 @@ export async function getZrxQuote(input: GetQuoteInput): Promise<Quote<ChainType
...
 
-  let minQuoteSellAmountWei = null
-  if (minQuoteSellAmount) {
-    minQuoteSellAmountWei = new BigNumber(minQuoteSellAmount as string).times(
-      new BigNumber(10).exponentiatedBy(sellAsset.precision)
-    )
-  }
+  const { minimum: minQuoteSellAmount, maximum: maxSellAmount } = await getZrxMinMax(input)
+
+  const minQuoteSellAmountWei = new BigNumber(minQuoteSellAmount as string).times(
```
kaladinlight, line 32: It looks like `minQuoteSellAmount` should already be a string, so we can remove the redundant type cast

packages/types/src/base.ts
```diff
@@ -125,14 +125,14 @@ export type QuoteResponse = {
   sources?: Array<SwapSource>
 }
 
-export type Quote<C extends ChainTypes> = {
+export type Quote<SellAssetType extends ChainTypes> = {
```
kaladinlight, line 128: standard generic convention makes use of single character uppercase names ie. `T` for type. this is usually sufficient for clarity especially when dealing with a single generic parameter. we may want a quick discussion on if we think this is reasonable, or if we would like to make use of more verbose and descriptive generic naming conventions. not a blocker, just something of note if anyone has any strong opinions one way or the other.

kaladinlight (dismissed):
approved with two small comments

packages/swapper/src/swappers/zrx/getZrxQuote/getZrxQuote.ts
elmutt: true. done

packages/types/src/base.ts
elmutt: I changed it back. going convention is fine with me

packages/swapper/src/swappers/zrx/getZrxQuote/getZrxQuote.ts
```diff
@@ -33,12 +27,11 @@ export async function getZrxQuote(input: GetQuoteInput): Promise<Quote<ChainType
...
 
-  let minQuoteSellAmountWei = null
-  if (minQuoteSellAmount) {
-    minQuoteSellAmountWei = new BigNumber(minQuoteSellAmount as string).times(
-      new BigNumber(10).exponentiatedBy(sellAsset.precision)
-    )
-  }
+  const { minimum: minQuoteSellAmount, maximum: maxSellAmount } = await getZrxMinMax(input)
+
+  const minQuoteSellAmountWei = new BigNumber(minQuoteSellAmount).times(
```
0xdef1cafe, line 32: If we're changing to this let's make it bnorzero

0xdef1cafe (changes requested):
bnOrZero plz if we're gonna change this

packages/swapper/src/swappers/zrx/getZrxQuote/getZrxQuote.ts
elmutt: yep. done

cjthompson:
Wait why doesn't it make sense for slippage to come from the quote? Isn't the slippage determined by the size of the trade?

0xdef1cafe (approved):

elmutt:
> Wait why doesn't it make sense for slippage to come from the quote? Isn't the slippage determined by the size of the trade?

The "slippage" we refer to here should be called the "maximumSlippage". It represents what the user is telling 0x it wants it to fail the trade if its worse than that.

We dont currently have a concept of "priceImpact" which is I think what you are referring to

theoboldfrazier (approved):

PR #601, opened 2022-05-04
chore: add noUnusedLocals TS rule

0xApotheosis:
## Description

- Add `noUnusedLocals` TS rule.
- Fix existing code that breaks the rule

Note, this PR uses the name-spaced `caip2` import to match our existing, but incorrect convention - this better implementation is addressed separately in https://github.com/shapeshift/lib/pull/560.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

N/A

## Testing

N/A

## Screenshots (if applicable)

N/A

### DIFF
--- a/packages/caip/src/index.ts
+++ b/packages/caip/src/index.ts
@@ -4,6 +4,6 @@ import * as caip10 from './caip10/caip10'
 import * as caip19 from './caip19/caip19'
 
 export { adapters, caip2, caip19, caip10 }
-export { ChainId, CAIP2 } from './caip2/caip2'
+export { ChainId, CAIP2, ChainReference } from './caip2/caip2'
 export { AccountId, CAIP10 } from './caip10/caip10'
 export { AssetId, AssetNamespace, AssetReference, CAIP19 } from './caip19/caip19'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -28,7 +28,7 @@ jobs:
         run: printf '%s' "$TITLE" | npx commitlint
       - name: Build
         run: yarn build
-      - name: Type check
+      - name: TypeScript compilation check
         run: yarn tsc --project ./tsconfig.json --noEmit
       - name: Lint
         run: yarn lint
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -26,6 +26,8 @@ jobs:
         env:
           TITLE: ${{ github.event.pull_request.title }}
         run: printf '%s' "$TITLE" | npx commitlint
+      - name: Type check
+        run: yarn tsc --project ./tsconfig.json --noEmit
       - name: Build
         run: yarn build
       - name: Lint
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -26,10 +26,10 @@ jobs:
         env:
           TITLE: ${{ github.event.pull_request.title }}
         run: printf '%s' "$TITLE" | npx commitlint
-      - name: Type check
-        run: yarn tsc --project ./tsconfig.json --noEmit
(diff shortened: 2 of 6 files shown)


packages/caip/src/index.ts
```diff
@@ -4,6 +4,6 @@ import * as caip10 from './caip10/caip10'
 import * as caip19 from './caip19/caip19'
 
 export { adapters, caip2, caip19, caip10 }
-export { ChainId, CAIP2 } from './caip2/caip2'
+export { ChainId, CAIP2, ChainReference } from './caip2/caip2'
```
gomesalexandre, line 7: Revert this hunk

gomesalexandre (dismissed):
Looks good.
Tested it out locally and compiler violations are properly reported:
```shell
$ lerna run type-check
lerna notice cli v4.0.0
lerna info versioning independent
lerna info Executing command in 7 packages: "yarn run type-check"
lerna info run Ran npm script 'type-check' in '@shapeshiftoss/types' in 1.1s:
$ tsc --project ./tsconfig.build.json --noEmit
lerna info run Ran npm script 'type-check' in '@shapeshiftoss/errors' in 1.1s:
$ tsc --project ./tsconfig.json --noEmit
lerna ERR! yarn run type-check exited 2 in '@shapeshiftoss/caip'
lerna ERR! yarn run type-check stdout:
$ tsc --project ./tsconfig.build.json --noEmit
src/index.ts:11:7 - error TS6133: 'unusedLocal' is declared but its value is never read.

11 const unusedLocal = 'test'
```

As we've discussed in Discord, having caip exported the old, consistent way is the way to go for now since https://github.com/shapeshift/lib/pull/560 flattens exports.

kaladinlight (dismissed):

kaladinlight:
I would raise the same question about what compiler/linter we want to add rules to where supported. tsc also supports noUnusedLocals, we just don't have it set currently. This may be one that eslint fix would automatically clean up whereas tsc would not.

gomesalexandre:
> I would raise the same question about what compiler/linter we want to add rules to where supported. tsc also supports noUnusedLocals, we just don't have it set currently. This may be one that eslint fix would automatically clean up whereas tsc would not.

Actually
image


We do have a rule for that in eslint as well, coming from [`eslint:recommended`](https://github.com/eslint/eslint/blob/ab37d3ba302856007beb833c34b56658a34bbb5d/conf/eslint-recommended.js#L67) rules:
https://github.com/shapeshift/lib/blob/e4562f7aed5cc1cfd3961ce28a86cd7d47759646/.eslintrc#L4 but it is then disabled by [plugin:@typescript-eslint/recommended](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts#L29) so we would have to re-enable it as its own rule.

WRT autofix, there is no autofix rule for `no-unused-vars`, as a design choice from the eslint team: https://github.com/eslint/eslint/issues/11640 so we might be better off with having this specific one in tsconfig only.

0xdef1cafe (approved):

gomesalexandre (approved):
Let's get this in, we can always remove the unused `ChainReference` export later.

PR #604, opened 2022-05-05
chore(types): replace caip properties with their high-level counterparts

0xApotheosis:
**BREAKING CHANGE: caip-specific type properties have been renamed with their generic counterparts.**

This is a major version bump to the `@shapeshiftoss/types` package ~and uses semantic release notation~ - manually bumped the version as per https://github.com/shapeshift/lib/pull/560#issuecomment-1111182529.

Follow-on PRs will update all affected packages in `lib`, as well as `web` by handling the breaking change and bumping `@shapeshiftoss/types`.

Contributes to https://github.com/shapeshift/unchained/issues/398.

PR #524, opened 2022-04-14
fix: add KICK to the blacklist

NeOMakinG:
## Context

The KICK token is shown on the `web` app:
image

We need to add the token to the blacklist, because it was missing, it's a scam token.

## Result
image

## Warning

It seems that rebuilding the `generatedAssetData.json` file made the job by itself, probably because `yearnfinance` removed the asset, but I would prefer to blacklist it in case the asset is inside another datasource in the future.

closes #509

PR #605, opened 2022-05-05
perf(assetService): replace caip properties with their high-level counterparts

0xApotheosis:
Updates `@shapeshiftoss/asset-service` to support breaking changes introduced in https://github.com/shapeshift/lib/pull/604.

Note, requires https://github.com/shapeshift/lib/pull/604 to be published, and assumes the new version number will be `4.0.0`.

Contributes to https://github.com/shapeshift/unchained/issues/398.

BREAKING CHANGE: Handle removal of CAIP vernacular in types package.


kaladinlight:
There are some additional references to caip naming in the `generateAssetData`, `AssetService.test.ts`, `generatedAssetData.json` (regen after generate scripts updated), and the `README.md` in reference to blacklisting

kaladinlight:
Seeing a few remaining references:

```
./src/service/AssetService.test.ts:  caip2: 'eip155:3',
./src/service/AssetService.test.ts:        caip2: 'eip155:3',
./src/generateAssetData/ethTokens/yearnVaults.ts:      caip2: caip2.toCAIP2({
./src/generateAssetData/ethTokens/yearnVaults.ts:      caip2: caip2.toCAIP2({
./src/generateAssetData/ethTokens/yearnVaults.ts:      caip2: caip2.toCAIP2({
./src/generateAssetData/ethTokens/yearnVaults.ts:      caip2: caip2.toCAIP2({
./src/generateAssetData/cosmos/getOsmosisAssets.ts:      caip2: 'cosmos:osmosis-1',
./README.md:You can contribute to blacklist a flagged token by adding its [caip19](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md) to the [blacklist.json](https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/generateAssetData/blacklist.json) list.
```

And we will want to regenerate the asset data to remove caip references in the generated output as well.

kaladinlight:
Also, would it be worth waiting for the flattened exports to go along in conjunction with these changes or planning on just having a follow up set of PR's for those changes. Just seems like it is going to hit all of the same places, whatever is easier I guess.

gomesalexandre:
> Also, would it be worth waiting for the flattened exports to go along in conjunction with these changes or planning on just having a follow up set of PR's for those changes. Just seems like it is going to hit all of the same places, whatever is easier I guess.

I think it makes more sense to keep the PRs separate to keep them more atomic.

In addition to the comment about remaining occurrences, what about renaming this to what it *does* (remove CAIP references) as opposed to *why* it does it (the update of `@shapeshiftoss/types`?) given we have a lot of similar PRs here, it would probably make history cleaner.

0xApotheosis:
> Seeing a few remaining references:

This is intentional @kaladinlight - this PR isn't meant to do everything at once, but rather just the things that need to be done to handle the breaking change. Most of these PRs are min-diff where possible (there was some creep with variable renaming in well-encapsulated functions).

There are still many remaining caip references after this group of PRs, but that is by design.

0xApotheosis:
> > Also, would it be worth waiting for the flattened exports to go along in conjunction with these changes or planning on just having a follow up set of PR's for those changes. Just seems like it is going to hit all of the same places, whatever is easier I guess.
> 
> I think it makes more sense to keep the PRs separate to keep them more atomic.
> 
> In addition to the comment about remaining occurrences, what about renaming this PR to what it _does_ (remove CAIP references) as opposed to _why_ it does it (the update of `@shapeshiftoss/types`?) given we have a lot of similar PRs here, it would probably make history cleaner.

Sounds good to me @gomesalexandre, I'll update the titles accordingly.

gomesalexandre:
> Seeing a few remaining references:
> And we will want to regenerate the asset data to remove caip references in the generated output as well.

While most of these are not related to this PR, the CAIP19/CAIP2 references should indeed be removed in `packages/asset-service/src/service/AssetService.test.ts` for `Asset` type.

0xApotheosis:
This should be good to go, though it concerns me that the diff is so large from the asset generation - did we blacklist a lot of assets recently?

NeOMakinG:
> This should be good to go, though it concerns me that the diff is so large from the asset generation - did we blacklist a lot of assets recently?

There is a problem with the generation, misses most of ERC20 etc...

https://github.com/shapeshift/lib/blob/main/packages/asset-service/src/generateAssetData/generateAssetData.ts#L21

Noticed that we use an orderBy with a caip19, I guess this can't work anymore, maybe this is the problem!

NeOMakinG (changes requested):
I didn't test it, but there is definitely a problem with the `generatedAssetData.json` content from an external point of view, my guess is that if you test it by linking to web, most of tokens won't be there

Added a small comment which should probably be a good start

gomesalexandre:
> This should be good to go, though it concerns me that the diff is so large from the asset generation - did we blacklist a lot of assets recently?

There's been two recently: https://github.com/shapeshift/lib/pull/584, https://github.com/shapeshift/lib/pull/524

gomesalexandre:
> Noticed that we use an orderBy with a caip19, I guess this can't work anymore, maybe this is the problem!

Nice catch @NeOMakinG ! We should change the sorting criteria in a separate PR before this can be merged. Otherwise the diff on the generated assets data will be huge, the reason for that being the filtered assets are not properly sorted out anymore.

0xApotheosis:
Nailed it @NeOMakinG!

I'll put this back in draft until https://github.com/shapeshift/lib/pull/620 is in, then regenerate the list.

kaladinlight (dismissed):
Looks solid. I did a fresh yarn and yarn generate and got slightly different generated asset data, but it mostly looked like new/changes symbol names so we are probs good, or can just generate fresh one more time for good measure. Also, be sure to check in the yarn.lock diff as there are a few additions there as well it looks like

0xApotheosis:
Thanks @kaladinlight, should be good to go - I regenerated the asset data for good measure.

NeOMakinG (approved):
Looks fine now, hope we didn't forget another reference but I couldn't find one so!

gomesalexandre (approved):
Looking good, ran `yarn && yarn build && cd packages/asset-service && yarn generate`.
The diff is only on the `packages/asset-service` scope which ensures a major bump will happen and be published.
You might want to add the now infamous `BREAKING CHANGE: <breaking changes here` to the description and update it since the PR you mentioned is now merged.

I'm again getting a diff on `generatedAssetData.json` but that's not related to these changes so probably no point to regenerate it again, your call:

```diff
diff --git a/packages/asset-service/src/service/generatedAssetData.json b/packages/asset-service/src/service/generatedAssetData.json
index 71d78e9..b39d58f 100644
--- a/packages/asset-service/src/service/generatedAssetData.json
+++ b/packages/asset-service/src/service/generatedAssetData.json
@@ -1855,7 +1855,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x0202Be363B8a4820f3F4DE7FaF5224fF05943AB1/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/12819/thumb/UniLend_Finance_logo_PNG.png?1602748658",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "UFT"
@@ -2650,7 +2650,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x044727e50ff30DB57fad06Ff4F5846eAb5eA52a2/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/19385/thumb/2Hy412Bd_400x400.png?1635146893",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "KITTY"
@@ -10054,7 +10054,7 @@
         "assetId": "eip155:1/erc20:0x19e2a43fbbc643c3b2d9667d858d49cad17bc2b5",
         "chainId": "eip155:1",
         "dataSource": "coingecko",
-        "name": "BNS Token",
+        "name": "BNS",
         "precision": 8,
         "tokenId": "0x19e2a43fbbc643c3b2d9667d858d49cad17bc2b5",
         "contractType": "erc20",
@@ -11200,7 +11200,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x1d350417d9787E000cc1b95d70E9536DcD91F373/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/13649/thumb/mirror_logo_transparent.png?1611565655",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "MIAU"
@@ -11305,7 +11305,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x1dA01e84F3d4e6716F274c987Ae4bEE5DC3C8288/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/12344/thumb/IMG_20200919_115022_477.png?1600739441",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "BID"
@@ -32404,7 +32404,7 @@
         "assetId": "eip155:1/erc20:0x586aa273f262909eef8fa02d90ab65f5015e0516",
         "chainId": "eip155:1",
         "dataSource": "coingecko",
-        "name": "Fixed Income Asset Token",
+        "name": "Fixed Income Asset",
         "precision": 18,
         "tokenId": "0x586aa273f262909eef8fa02d90ab65f5015e0516",
         "contractType": "erc20",
@@ -54595,7 +54595,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x986EE2B944c42D017F52Af21c4c69B84DBeA35d8/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/5236/thumb/bitmart-token.png?1548835017",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "BMX"
@@ -54745,7 +54745,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x9903A4Cd589DA8e434f264deAFc406836418578E/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/12196/thumb/cc2016f6-0c74-4a95-b89b-b1684c7b9426.png?1597991823",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "FIRST"
@@ -62320,7 +62320,7 @@
         "contractType": "erc20",
         "color": "#FFFFFF",
         "secondaryColor": "#FFFFFF",
-        "icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0xaDA62f7CCd6af6cAcff04ACCBC4f56f3D4FFd4Ef/logo.png",
+        "icon": "https://assets.coingecko.com/coins/images/10313/thumb/BNXcW4kq_400x400.jpg?1577833532",
         "sendSupport": true,
         "receiveSupport": true,
         "symbol": "PLF"
```

0xApotheosis:
Thorough as always, cheers @gomesalexandre!

I think the "BREAKING CHANGE:" copy just needs to be in a commit footer, which gets wrapped up in the squash merge - but I've added it to the PR description just in case...

Re: the asset generation, yeah - every time I run it I get slightly different results, I'm not too worried about a small diff there.

NeOMakinG:
It's related to every API pushing changes, I was surprised too the first time I worked on lib, I even blacklisted an asset that wasn't even inside the new build

elmutt (approved):

PR #606, opened 2022-05-05
perf(chainAdapters): replace caip properties with their high-level counterparts

0xApotheosis:
Updates `@shapeshiftoss/chainAdapters` to support breaking changes introduced in https://github.com/shapeshift/lib/pull/604.

Note, requires https://github.com/shapeshift/lib/pull/604 to be published, and assumes the new version number will be `4.0.0`.

Contributes to https://github.com/shapeshift/unchained/issues/398.

PR #609, opened 2022-05-05
chore: major bump dependencies on @shapeshiftoss/types

0xApotheosis:
Updates packages to use the latest major version of `@shapeshiftoss/types`.

Note, requires https://github.com/shapeshift/lib/pull/604 to be published, and assumes the new version number will be `4.0.0`.

Contributes to https://github.com/shapeshift/unchained/issues/398.


kaladinlight:
https://github.com/shapeshift/lib/pull/606 will also result in not needing to bump types in investor packages here as well

PR #610, opened 2022-05-05
feat: add getSupportedAccountTypes to ChainAdapterManager

gomesalexandre:
This adds `getSupportedAccountTypes` into `ChainAdapterManager`, effectively giving this guy a better house:
https://github.com/shapeshift/web/blob/89e26b31f9e46638e9578544486edc663311954e/src/state/slices/portfolioSlice/portfolioSliceCommon.ts#L4-L10

Take one, not sure about this way of abstracting it and having it all as one object in the manager.

A second PR (#611) has an alternative implementation, having each adapter define its own supported account types, which for all chains apart from UTXOs means `[undefined]` for now, but when thinking with the future of what we house as "accounts" (e.g account abstractions like ERC4437), they might be populated with some items at some point in the future.

### DIFF
--- a/packages/chain-adapters/src/ChainAdapterManager.ts
+++ b/packages/chain-adapters/src/ChainAdapterManager.ts
@@ -1,5 +1,5 @@
 import { CAIP2, caip2 } from '@shapeshiftoss/caip'
-import { ChainTypes } from '@shapeshiftoss/types'
+import { ChainTypes, UtxoAccountType } from '@shapeshiftoss/types'
 import * as unchained from '@shapeshiftoss/unchained-client'
 
 import { ChainAdapter } from './api'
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/ChainAdapterManager.ts
```diff
@@ -90,6 +90,22 @@ export class ChainAdapterManager {
     return Array.from(this.supported.keys())
   }
 
+  getSupportedAccountTypes(): Record<ChainTypes, Array<UtxoAccountType | undefined>> {
```
0xdef1cafe, line 93: this should call the individual adapters and return what they respond with, otherwise this is just coupling the adapters to the manager?

0xdef1cafe (changes requested):
call adapters rather than having the manager respond with this

gomesalexandre:
Closed in favor of #611

PR #620, opened 2022-05-06
feat: regenerate asset data with assetId as sorting criteria

gomesalexandre:
This:
- changes the sorting criteria for `filteredAssetData` from `caip19` to `assetId` to match the new high-level terminology we use. It will be needed for https://github.com/shapeshift/lib/pull/605 since it also regenerates asset data, but with the old criteria.
- regenerates `packages/asset-service/src/generateAssetData/generateAssetData.ts`.


NeOMakinG (approved):
Didn't test it yet but looks fine 👍

0xApotheosis (approved):

0xdef1cafe (approved):

PR #607, opened 2022-05-05
perf(marketService): replace caip properties with their high-level counterparts

0xApotheosis:
Updates `@shapeshiftoss/market-service` to support breaking changes introduced in https://github.com/shapeshift/lib/pull/604.

Note, requires https://github.com/shapeshift/lib/pull/604 to be published, and assumes the new version number will be `4.0.0`.

Contributes to https://github.com/shapeshift/unchained/issues/398.

PR #622, opened 2022-05-06
perf(caip): remove caip vernacular

0xApotheosis:
BREAKING CHANGE


packages/caip/README.md
```diff
@@ -10,20 +10,20 @@ https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md
...
 expect(result).toEqual('eip155:1')
 ```
 
-### `fromCAIP2`
+### `fromChainId`
 
 ```ts
 const ethCaip2 = 'eip155:1'
-const { chain, network } = fromCAIP2(ethCaip2)
+const { chain, network } = fromChainId(ethCaip2)
```
gomesalexandre, line 26:
```suggestion
const { chain, network } = fromChainId(ethChainId)
```

gomesalexandre (changes requested):
Looking good, almost all references are gone!
The `CAIP-2` section variable naming in the README example, as well as all references to the vernacular in the `CAIP-19` section should also be updated.

gomesalexandre:
> Also, while all of these changes are applicable, I was thinking this PR would just include the caip naming updates in the caip package only. Then the caip updates would be brought in alongside the type updates in the existing open PR's for the remaining packages.

> This will result in a single major version bump for each package for all encompassed breaking changes.

Agreed with this order, we don't want all to major bump. For reference, proceeded this way for https://github.com/shapeshift/lib/pull/560 which only minor bumped caip as well as the affected packages (since it kept the old implementation, I could use chore type for the commit/PR) then https://github.com/shapeshift/lib/pull/623, a perf type, which should major bump caip only since it's now just having a diff on caip.

kaladinlight:
> > Also, while all of these changes are applicable, I was thinking this PR would just include the caip naming updates in the caip package only. Then the caip updates would be brought in alongside the type updates in the existing open PR's for the remaining packages.
> 
> > This will result in a single major version bump for each package for all encompassed breaking changes.
> 
> Agreed with this order, we don't want all to major bump. For reference, proceeded this way for #560 which only minor bumped caip as well as the affected packages (since it kept the old implementation, I could use chore type for the commit/PR) then #623, a perf type, which should major bump caip only since it's now just having a diff on caip.

I did comment on https://github.com/shapeshift/lib/pull/623 that those changes will be included in this PR and can be closed. Same story for https://github.com/shapeshift/lib/pull/624

gomesalexandre:
> I did comment on #623 that those changes will be included in this PR and can be closed. Same story for #624

Yep! But the logic will be the same as the (intended) one.

0xApotheosis:
> The order should be as follows:
> 
> 1. package/types v4.0.0 (done)
> 2. package/caip v3.0.0 (this PR - caip packages changes only)
> 3. [perf(chainAdapters): replace caip properties with their high-level counterparts #606](https://github.com/shapeshift/lib/pull/606) with additional caip updates and `unchained-client` updates included
> 4. [perf(assetService): replace caip properties with their high-level counterparts #605](https://github.com/shapeshift/lib/pull/605) with additional caip updates
> 5. [perf(marketService): replace caip properties with their high-level counterparts #607](https://github.com/shapeshift/lib/pull/607) with additional caip updates

Sounds good, I'll work some git magic and update each PR accordingly.

0xApotheosis:
Closing in favor of https://github.com/shapeshift/lib/pull/625

PR #634, opened 2022-05-11
fix: dont crash when passing undefined to filterBuyAssetsBySellAssetId

elmutt:
Fixes: https://github.com/shapeshift/web/issues/1746

### DIFF
--- a/packages/swapper/src/swappers/zrx/ZrxSwapper.ts
+++ b/packages/swapper/src/swappers/zrx/ZrxSwapper.ts
@@ -83,13 +83,11 @@ export class ZrxSwapper implements Swapper {
   }
 
   filterBuyAssetsBySellAssetId(args: BuyAssetBySellIdInput): AssetId[] {
-    const { assetIds, sellAssetId } = args
-    return assetIds?.filter(
-      (id) => id.startsWith('eip155:1') && sellAssetId?.startsWith('eip155:1')
-    )
+    const { assetIds = [], sellAssetId } = args
+    return assetIds.filter((id) => id.startsWith('eip155:1') && sellAssetId?.startsWith('eip155:1'))
   }
 
-  filterAssetIdsBySellable(assetIds: AssetId[]): AssetId[] {
-    return assetIds?.filter((id) => id.startsWith('eip155:1'))
+  filterAssetIdsBySellable(assetIds: AssetId[] = []): AssetId[] {
+    return assetIds.filter((id) => id.startsWith('eip155:1'))
   }
 }
--- a/packages/swapper/src/swappers/zrx/ZrxSwapper.ts
+++ b/packages/swapper/src/swappers/zrx/ZrxSwapper.ts
@@ -84,10 +84,12 @@ export class ZrxSwapper implements Swapper {
 
   filterBuyAssetsBySellAssetId(args: BuyAssetBySellIdInput): AssetId[] {
     const { assetIds, sellAssetId } = args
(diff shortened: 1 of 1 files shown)


DaoDev44 (dismissed):

0xdef1cafe (approved):
makes sense, approved, but curious how this got past the type system?

https://github.com/shapeshift/lib/blob/main/packages/swapper/src/api.ts#L25

this fixes the symptom, but where is the root cause? i suspect a selector in web is return a type sig of `AssetId[]` when in reality it's returning `AssetId[] | undefined`

packages/swapper/src/swappers/zrx/ZrxSwapper.ts
```diff
@@ -84,10 +84,12 @@ export class ZrxSwapper implements Swapper {
 
   filterBuyAssetsBySellAssetId(args: BuyAssetBySellIdInput): AssetId[] {
     const { assetIds, sellAssetId } = args
-    return assetIds.filter((id) => id.startsWith('eip155:1') && sellAssetId.startsWith('eip155:1'))
+    return assetIds?.filter(
```
kaladinlight, line 87: this will return undefined whereas the function sig implies we would always like to return an array.

some other options:

default value - `const { assetIds = [], sellAssetId } = args`
nullish operator - `return (assetIds ?? []).filter` or `return assetIds?.filter() ?? []`

0xdef1cafe (changes requested):
requesting changes as per @kaladinlight comment - he's correct

elmutt:
> 

donno

elmutt:
> makes sense, approved, but curious how this got past the type system?
> 
> https://github.com/shapeshift/lib/blob/main/packages/swapper/src/api.ts#L25
> 
> this fixes the symptom, but where is the root cause? i suspect a selector in web is return a type sig of `AssetId[]` when in reality it's returning `AssetId[] | undefined`

not sure how it made it in

kaladinlight:
> makes sense, approved, but curious how this got past the type system?
> 
> https://github.com/shapeshift/lib/blob/main/packages/swapper/src/api.ts#L25
> 
> this fixes the symptom, but where is the root cause? i suspect a selector in web is return a type sig of `AssetId[]` when in reality it's returning `AssetId[] | undefined`

I am curious about this too. I wonder if some sort of inferred return type rule is in place that is covering this up. I prefer explicit return types for this reason (though it is explicit in this case). Def strange...

packages/swapper/src/swappers/zrx/ZrxSwapper.ts
elmutt: good feedback, thanks

0xdef1cafe:
> 

this is the root cause - we should enable explicit return types. i'm always banging the drum about writing a function type sig first and the implementation last.

0xdef1cafe:
@0xApotheosis this one will make your ears burn - suggesting https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

elmutt:
> requesting changes as per @kaladinlight comment - he's correct

done

0xdef1cafe (approved):

kaladinlight (approved):

0xApotheosis:
> @0xApotheosis this one will make your ears burn - suggesting https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

I'm all for it. 181 functions currently break that rule - hopefully most are easy fixes, but I expect some will propagate changes throughout the codebase and require special attention.

PR #639, opened 2022-05-12
chore: use bn where appropriate

0xApotheosis:
## Description

We use `bnOrZero` in places we shouldn't.

This PR removes this pattern to stop it from propagating.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

Minimal.

## Testing

Everything should work as it does now.

## Screenshots (if applicable)

N/A

### DIFF
--- a/packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
+++ b/packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
@@ -1,16 +1,12 @@
-import { ChainTypes, GetQuoteInput, MinMaxOutput } from '@shapeshiftoss/types'
+import { Asset, MinMaxOutput } from '@shapeshiftoss/types'
 
 import { bn } from '../utils/bignumber'
 import { MAX_ZRX_TRADE } from '../utils/constants'
 import { getUsdRate } from '../utils/helpers/helpers'
 import { ZrxError } from '../ZrxSwapper'
 
-export const getZrxMinMax = async (
-  input: Pick<GetQuoteInput, 'sellAsset' | 'buyAsset'>
-): Promise<MinMaxOutput> => {
-  const { sellAsset, buyAsset } = input
-
-  if (sellAsset.chain !== ChainTypes.Ethereum || buyAsset.chain !== ChainTypes.Ethereum) {
+export const getZrxMinMax = async (sellAsset: Asset, buyAsset: Asset): Promise<MinMaxOutput> => {
+  if (sellAsset.chainId !== 'eip155:1' || buyAsset.chainId !== 'eip155:1') {
     throw new ZrxError('getZrxMinMax - must be eth assets')
   }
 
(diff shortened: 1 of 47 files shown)


packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
```diff
@@ -19,7 +19,7 @@ export const getZrxMinMax = async (
     tokenId: sellAsset.tokenId
   })
 
-  const minimum = bnOrZero(1).dividedBy(bnOrZero(usdRate)).toString()
+  const minimum = bn(1).dividedBy(bnOrZero(usdRate)).toString()
```
kaladinlight, line 22: This seems like we need to check and error on invalid `usdRate` to avoid divide by zero error which looks like it results in string `Infinity`

elmutt: is it a good idea to assume getUsdRate will always behave that way?

elmutt (approved):

packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
0xApotheosis: I think so, as we maintain it - a change in behavior should be caught in review.

Addressed in https://github.com/shapeshift/lib/pull/639/commits/b294ba1441c3bb2fe451344d6418d83dfd1971ff

elmutt (dismissed):

kaladinlight (approved):
Approved assuming there was discussion about removing the previous `bnOrZero` -> `bn` changes.

packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
```diff
@@ -15,7 +15,7 @@ export const getZrxMinMax = async (sellAsset: Asset, buyAsset: Asset): Promise<M
     tokenId: sellAsset.tokenId
   })
 
-  const minimum = bn(1).dividedBy(bnOrZero(usdRate)).toString()
+  const minimum = bn(1).dividedBy(usdRate).toString()
```
0xdef1cafe, line 18: can't safely divide here - plz use bnorzero

0xdef1cafe (changes requested):

packages/swapper/src/swappers/zrx/getZrxMinMax/getZrxMinMax.ts
0xApotheosis: Right you are - addressed in https://github.com/shapeshift/lib/pull/639/commits/27b8bebf2fc5d05e2ba07e9212ee26b796aed351

0xApotheosis:
Incorporated in https://github.com/shapeshift/lib/pull/641

PR #644, opened 2022-05-12
fix: correct fee calculation for eth/erc20 sends

kaladinlight:
Fixes issue: https://github.com/shapeshift/web/issues/1716

- Use standard fee as the normalization constant
- Use fast instead of instant (instant was often lower than slow for unknown reasons)
- Use the correct fee data for each return object (slow, average, fast)

There is a larger question/concern of relying on https://gas.api.0x.org/ for our fee calculations. Using the `MEDIAN` return value from this api results in a lower gas price than that returned by geth or something like https://etherchain.org/tools/gasnow. The result lines up more with https://ethgasstation.info/.

There are clearly fee discrepancies across the ecosystem depending on where you look, so we should have a discussion on whether we are ok with the returned data from 0x or if we want to build out the unchained fees response to include slow/average/fast results that we can adjust and control ourselves.

It is also worth noting that the difference between slow/average/fast from the 0x gas api is very small resulting in very little diff between the levels to the point they don't even matter much (specifically for sending eth/erc20s). This may just be the current network state, but something to keep our eyes on.

### DIFF
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
@@ -172,27 +172,27 @@ describe('EthereumChainAdapter', () => {
           average: {
             chainSpecific: {
               gasLimit: '21000',
-              gasPrice: '50180000000',
+              gasPrice: '45000000000',
               maxFeePerGas: '300',
               maxPriorityFeePerGas: '10'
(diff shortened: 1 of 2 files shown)


elmutt (approved):

0xdef1cafe (approved):
much noicer

elmutt (approved):

PR #647, opened 2022-05-15
fix: sort fiat price history by time correctly

0xdef1cafe:
the exchange rate host was returning newer fiat price history data first, we should be returning
oldest at the beginning of the array, like we do with crypto data.

also uses bnOrZero and deletes redundant NaN test as a result


0xApotheosis (approved):
Tight 🍫 

Linked + tested locally, no apparent regressions.

gomesalexandre (approved):
Linked this one locally and LGTM, didn't find any regressions testing against prod.

PR #655, opened 2022-05-17
chore: remove getinfo from cosmos base adapter

elmutt:
getInfo is not part of the chain adapter interface and is not used anywhere.

We noticed this trying to implement osmosis and believe it should be removed


DaoDev44 (approved):

0xApotheosis (changes requested):
Good catch. It looks like it's mocked 24 times in various `chain-adapter` tests, too.

Can we also remove those references?

kaladinlight (approved):

0xdef1cafe (approved):

Issue #550, opened 2022-04-25
Using strings over enums


0xdef1cafe:
### Overview

We want to move away from using enums for ChainNamespace, ChainReference, AssetNamespace, AssetReference and move to strings instead.

We don't gain anything from enums and the compiler is smart enough to infer/autocomplete strings instead.

### References and additional details

https://github.com/shapeshift/lib/blob/main/packages/caip/src/caip19/caip19.ts#L10

Note - this may be difficult to do without breaking changes, consider using a new name such as `ChainNamespaceString`

### Acceptance Criteria

We have string versions of these enums instead.

PR #656, opened 2022-05-17
perf: replace enums with dynamically derived object literals

0xApotheosis:
Replaces `ChainNamespace`, `ChainReference`, `AssetNamespace`, `AssetReference` enums with types dynamically derived from object literals.

This gives us literal string union types, a single source of truth for each reference/namespace, and object literals for good developer experience. String literals will also type check correctly.

For example:

```ts
export type AssetNamespace = typeof ASSET_NAMESPACE[keyof typeof ASSET_NAMESPACE]

const test1: AssetNamespace = 'cw20' // ✅ passes type check
const test2: AssetNamespace = ASSET_NAMESPACE.CW20 // ✅ passes type check
const test3: AssetNamespace = 'invalid-string' // 🛑 fails type check
```

Closes https://github.com/shapeshift/lib/issues/550

Screen Shot 2022-05-17 at 7 16 29 pm


packages/caip/src/adapters/osmosis/utils.ts
```diff
@@ -39,11 +39,11 @@ export const parseOsmosisData = (data: OsmosisCoin[]) => {
 
     if (isNativeAsset) {
       // TODO(ryankk): remove `toString` when AssetReferences are changed to strings
-      assetReference = isOsmo ? AssetReference.Osmosis.toString() : denom
-      assetNamespace = isOsmo ? AssetNamespace.Slip44 : AssetNamespace.NATIVE
+      assetReference = isOsmo ? ASSET_REFERENCE.Osmosis.toString() : denom
```
kaladinlight, line 42: Can remove `toString()` and comment above

packages/investor-foxy/src/api/api.ts
```diff
@@ -53,14 +53,14 @@ import {
 
 export * from './foxy-types'
 
+// EthereumMainnet, EthereumRinkeby, EthereumRopsten
+type Network = '1' | '3' | '4'
```
kaladinlight, line 57: Reason to not use constant strings over hardcode?

elmutt (approved):

packages/caip/src/adapters/osmosis/utils.ts
0xApotheosis: Sounds good - removed in https://github.com/shapeshift/lib/pull/656/commits/f2b3ea985a4f48f4d3e7ba2e14bd3491cfecf48c

packages/investor-foxy/src/api/api.ts
0xApotheosis: There was an issue using the `CHAIN_REFERENCE` constant due to a circular import, or something similar.

It was a battle I chose not to fight last night - I'll take another look today and add an inline comment for clarity if I can't work it out.

kaladinlight: Circular import strikes me as odd considering the caip package where they are defined should not be importing any "feature" packages in lib. But that sounds good, let me know.

0xApotheosis: Committed what we landed on in https://github.com/shapeshift/lib/pull/656/commits/8226d6f9308f7999c4154eb4c3b36ae6be5c7d39

kaladinlight (approved):

0xApotheosis:
Don't merge yet - this should technically be a breaking change.

elmutt (approved):

0xApotheosis:
Converting to draft until I get a corresponding web PR ready.

PR #661, opened 2022-05-18
chore: bump peer dependencies and remove src imports

0xApotheosis:
- Update peer dependencies to breaking changes in https://github.com/shapeshift/lib/pull/656
- Remove 2 erroneous imports from `/src`

PR #663, opened 2022-05-18
fix: bump caip package version in dependencies

cjthompson:


0xdef1cafe:
@cjthompson https://github.com/shapeshift/lib/pull/661 this just fixed the same thing

PR #667, opened 2022-05-19
fix: timeframes for fiat rates

0xdef1cafe:
see the tests - we weren't returning any data for the hourly timeframe with the previous changes. ensures this works across all timeframes.


0xApotheosis (dismissed):
Non-blocking comments.

cjthompson (approved):

PR #668, opened 2022-05-19
feat(chain-adapters): change byChainId to by synchronous

cjthompson:
`ChainAdapter.getChainId()` was changed to a synchronous function a while back but the `byChainId` function was never updated to match.

`byChainId` was async for legacy reasons when we used to get the ChainId from the backend. We don't do that anymore and `byChain` was already changed to synchronous but `byChainId` didn't get changed at the same time.

This isn't a breaking change because you can `await` a synchronous function and you'll get back the correct result. This will only break cases where we use `.then` or `.catch` but I don't think we do.

### DIFF
--- a/packages/chain-adapters/package.json
+++ b/packages/chain-adapters/package.json
@@ -41,7 +41,7 @@
     "@shapeshiftoss/hdwallet-core": "^1.20.0",
     "@shapeshiftoss/hdwallet-native": "^1.20.0",
     "@shapeshiftoss/types": "^4.0.0",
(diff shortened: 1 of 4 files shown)


0xApotheosis (approved):
Nice catch 🍭

gomesalexandre (approved):
LGTM, built and tested locally, tests are happy and implementation of `byChainId` is indeed fully synchronous now 🍬

0xdef1cafe (approved):

PR #670, opened 2022-05-19
feat: add supply and maxSupply fields to market service response

chxash:
This PR implements supply and maxSupply fields for `findAll` and `findByAssetId` methods in market-service.
For coingecko, I choose to rely on total_supply field when necessary and commented accordingly why in the code.
This is needed in web for https://github.com/shapeshift/web/issues/1771

testing : answer from market-service can be tested using marketCli command, by modifying marketDataCLI.ts, supply and maxSupply fields are present when values are present in provider answer.

Code for marketDataCLI.ts : 

```ts
import { CoinGeckoMarketService } from './coingecko/coingecko'

const main = async (): Promise<void> => {

  const cgMarketService = new CoinGeckoMarketService()
  const data = await cgMarketService.findAll()
  console.log({ data })
}

main()
  .then(() => console.info('Done'))
  .then(() => process.exit(0))
```


gomesalexandre (commented):
Tested locally with the market-service CLI and LGTM. Also compared the FOX values to the ones from the Coingecko website which also looks correct.
The compromise of using coingecko's total supply when max supply is unavailable seems sound to me. Will take another look at the diff itself before stamping this!

packages/market-service/src/osmosis/osmosis.test.ts
```diff
@@ -43,39 +43,46 @@ describe('osmosis market service', () => {
...
+      const result = await osmosisMarketService.findByAssetId(args)
+
+      expect(result).toEqual(expect.objectContaining(expectedResult))
+      expect(parseFloat(result!.supply!)).toBeCloseTo(3867804.88, 2)
     })
 
     it('should return market data for Ion', async () => {
       const args = { assetId: 'cosmos:osmosis-1/native:uion' }
-      const result: MarketData = {
+      const expectedResult = {
```
gomesalexandre, line 62: Ditto

gomesalexandre (approved):
Few typings comments but overall LGTM

packages/market-service/src/coincap/coincap.ts
```diff
@@ -82,7 +85,9 @@ export class CoinCapMarketService implements MarketService {
         price: marketData.priceUsd,
         marketCap: marketData.marketCapUsd,
         changePercent24Hr: parseFloat(marketData.changePercent24Hr),
-        volume: marketData.volumeUsd24Hr
+        volume: marketData.volumeUsd24Hr,
+        supply: marketData.supply,
+        maxSupply: marketData.maxSupply?.toString()
```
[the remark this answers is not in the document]
chxash: maxSupply returned by this method should be maxSupply?: string as a member of MarketData structure

packages/market-service/src/osmosis/osmosis.test.ts
chxash: fixed !

elmutt (approved):

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -100,11 +105,19 @@ export class CoinGeckoMarketService implements MarketService {
 
       const currency = 'usd'
       const marketData = data?.market_data
+      /* max_supply may be null on coingecko while available on other sources (coincap)
+      hence choosing to take value from total_supply if existing
+      Also a lot of time when max_supply is null, total_supply is the maximum supply on coingecko
+      We can reassess in the future the degree of precision we want on that field */
```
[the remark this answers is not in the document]
theoboldfrazier: Given this issue, we may consider moving coincap to the beginning of the list in data sources to read from.

Before, we never really had any reason that coingecko was the first data source other than it was the first one implemented.

If you are finding in your testing that coincap is more reliable in returning some of this data and if moving to the first data source doesn't cause any obvious issues, this would be a good reason to do that.

Would just be moving coincap above coingecko in `packages/market-service/src/market-providers.ts`

theoboldfrazier (approved):
Left a non-blocking comment.

Looks good!

packages/market-service/src/coingecko/coingecko.ts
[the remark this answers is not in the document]
chxash: @theoboldfrazier That is a good suggestion and I actually contemplated doing it while developping this. Can I create a separate issue for this ? I am afraid doing it now would increase the scope of our issue maybe a bit too much.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -100,11 +105,19 @@ export class CoinGeckoMarketService implements MarketService {
...
       return {
         price: marketData?.current_price?.[currency],
         marketCap: marketData?.market_cap?.[currency],
         changePercent24Hr: marketData?.price_change_percentage_24h,
-        volume: marketData?.total_volume?.[currency]
+        volume: marketData?.total_volume?.[currency],
+        supply: marketData?.circulating_supply,
+        maxSupply: marketData?.max_supply
+          ? marketData?.max_supply
+          : marketData?.total_supply?.toString()
```
0xdef1cafe, line 120:
```suggestion
        maxSupply: marketData?.max_supply ?? marketData?.total_supply?.toString()
```

0xdef1cafe: non blocking comment but i see this unnecessary ternary pattern a lot

0xdef1cafe (approved):

0xdef1cafe (changes requested):
requesting changes so this doesn't get broken until types is updated

0xdef1cafe (approved):

0xdef1cafe:
merging as this was previously approved - changes are updating types package from a separate PR, thank you ser

PR #677, opened 2022-05-23
chore: add zrx specific trade type to hold zrx specific trade data

elmutt:
Add zrx specific trade type to hold zrx specific trade tx data.

This is safe to merge because web uses the `Trade` type and `ZrxTrade` extends `Trade`

### DIFF
--- a/packages/swapper/src/api.ts
+++ b/packages/swapper/src/api.ts
@@ -83,9 +83,12 @@ export interface TradeQuote<C extends SupportedChainIds> extends TradeBase<C> {
 }
 
 export interface Trade<C extends SupportedChainIds> extends TradeBase<C> {
+  receiveAddress: string
+}
+
+export interface ZrxTrade<C extends SupportedChainIds> extends Trade<C> {
   txData: string
   depositAddress: string
-  receiveAddress: string
 }
 
 export type ExecuteTradeInput<C extends SupportedChainIds> = {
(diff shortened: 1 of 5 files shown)


packages/swapper/src/swappers/zrx/zrxExecuteTrade/zrxExecuteTrade.test.ts
```diff
@@ -40,7 +40,7 @@ describe('ZrxExecuteTrade', () => {
         chainSpecific: { approvalFee: '123600000', estimatedGas: '1235', gasPrice: '1236' }
       },
       sources: []
-    },
+    } as ZrxTrade<'eip155:1'>,
```
0xdef1cafe, line 43: please implement this as `const trade: ZrxTrade<'eip155:1> = ...` rather than casting - casting in tests is still dangerous

packages/swapper/src/api.ts
```diff
@@ -83,9 +83,12 @@ export interface TradeQuote<C extends SupportedChainIds> extends TradeBase<C> {
 }
 
 export interface Trade<C extends SupportedChainIds> extends TradeBase<C> {
+  receiveAddress: string
+}
+
+export interface ZrxTrade<C extends SupportedChainIds> extends Trade<C> {
```
0xdef1cafe, line 89: please comment in https://discord.com/channels/554694662431178782/886041582422356018/978367019651047454 to find a better home for `SupportedChainIds` before we spread this pattern further

0xdef1cafe (changes requested):

packages/swapper/src/api.ts
elmutt: Based on @majorhayes response in the discord discussion I feel like we need to talk about it further and not necessarily make the decision with this PR

0xdef1cafe (approved):

DaoDev44 (approved):

PR #678, opened 2022-05-23
chore: tighten up normalizeamount

elmutt:
tighten up normalizeAmount

### DIFF
--- a/packages/swapper/src/swappers/zrx/utils/helpers/helpers.ts
+++ b/packages/swapper/src/swappers/zrx/utils/helpers/helpers.ts
@@ -44,8 +44,7 @@ type GrantAllowanceArgs = {
  * ETH or ERC20, we still keep 7 decimal places.
  * @param amount
  */
-export const normalizeAmount = (amount: string | undefined): string | undefined => {
-  if (!amount) return
+export const normalizeAmount = (amount: string): string => {
   return bnOrZero(amount).toNumber().toLocaleString('fullwide', { useGrouping: false })
 }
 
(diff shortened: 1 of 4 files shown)


packages/swapper/src/swappers/zrx/utils/helpers/helpers.ts
```diff
@@ -44,8 +44,7 @@ type GrantAllowanceArgs = {
  * ETH or ERC20, we still keep 7 decimal places.
  * @param amount
  */
-export const normalizeAmount = (amount: string | undefined): string | undefined => {
-  if (!amount) return
+export const normalizeAmount = (amount: string): string => {
```
0xdef1cafe, line 47:
```suggestion
export const normalizeAmount = (amount: string | number | BigNumber): string => {
```

don't we want this to be more flexible, it was the return type that was the problem? returning `string` is correct - the previous signature of `string | undefined` didn't make sense given the name of the function.

kaladinlight (dismissed):

packages/swapper/src/swappers/zrx/utils/helpers/helpers.ts
elmutt: gotcha. makes sense - I re added the input types

kaladinlight (dismissed):

packages/swapper/src/swappers/zrx/zrxBuildTrade/zrxBuildTrade.ts
```diff
@@ -77,7 +77,7 @@ export async function zrxBuildTrade(
         params: {
           buyToken,
           sellToken,
-          sellAmount: normalizeAmount(sellAmount?.toString()),
+          sellAmount: normalizeAmount(sellAmount.toString()),
```
0xdef1cafe, line 80: now this can just be `sellAmount` because you made the input more flexible?

0xdef1cafe (approved):

DaoDev44 (approved):

PR #676, opened 2022-05-23
feat(caip): update to & from type signatures

0xApotheosis:
- Adds `chainNamespace` and `chainReference` to CAIP input/output helper type signatures
- By proxy, removes `ChainTypes` and `NetworkTypes` enums from the CAIP package
- Adds type guards and assertion functions to all related types

Implements the CAIP package portions of https://github.com/shapeshift/lib/issues/547 and https://github.com/shapeshift/web/issues/1570 - all changes are isolated to `packages/caip`.

It's a big one due to the number of places these helpers are used (especially tests). Besides some quality of life refactoring along the way and type narrowing for safety, this is about as atomic as we'll get for this piece of work.

Note that this is a breaking change for the CAIP package.

---

This is **Part 1** of a 3-part process.
**Part 2:** update affected `lib` packages
**Part 3:** update `web` to handle breaking changes

PR #682, opened 2022-05-24
chore(marketService): bump market-service CAIP dependency to 5.0.0

0xApotheosis:
- Updates `market-service`'s CAIP package dependency to `5.0.0`, a breaking change: https://github.com/shapeshift/lib/pull/676

---

This is a patch release of `@shapeshiftoss/market-service`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2810,6 +2810,11 @@
     tsoa "^3.4.0"
     ws "^8.3.0"
 
+"@shapeshiftoss/caip@^4.0.1":
+  version "4.0.1"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -2810,6 +2810,11 @@
     tsoa "^3.4.0"
     ws "^8.3.0"
 
+"@shapeshiftoss/caip@^4.0.1":
```
kaladinlight, line 2813: This should go away after caip package v5.x.x published correctly

0xApotheosis: Seems like it won't until all of these dependent PRs are in.

0xdef1cafe (approved):

PR #684, opened 2022-05-24
chore(assetService): bump asset-service CAIP dependency to 5.0.0

0xApotheosis:
- Updates `asset-service`'s CAIP package dependency to `5.0.0`, a breaking change: https://github.com/shapeshift/lib/pull/676

No asset re-generation should be necessary as this is a refactor only.

---

This is a patch release of `@shapeshiftoss/asset-service`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2810,6 +2810,11 @@
     tsoa "^3.4.0"
     ws "^8.3.0"
 
+"@shapeshiftoss/caip@^4.0.1":
+  version "4.0.1"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -2810,6 +2810,11 @@
     tsoa "^3.4.0"
     ws "^8.3.0"
 
+"@shapeshiftoss/caip@^4.0.1":
```
kaladinlight, line 2813: Regenerate to remove diff after caip v5.x.x published

0xdef1cafe (approved):

Issue #556, opened 2022-04-25
Remove ChainTypes - many many tickets


0xdef1cafe:
### Overview

This will be the last ticket - once chain adapters relies on CAIPs, and web/lib/unchained do not refer to ChainTypes, it can die.

### References and additional details

https://github.com/shapeshift/lib/blob/main/packages/types/src/base.ts#L14-L15

### Acceptance Criteria

Good riddance `ChainTypes`

PR #688, opened 2022-05-24
feat(types): remove dataSource, NetworkTypes, and ChainTypes

0xApotheosis:
- Removes the `dataSource` property from `AbstractAsset`
- Removes `NetworkTypes` enum
- Removes `ChainTypes` enum

Contributes to https://github.com/shapeshift/lib/issues/556 and https://github.com/shapeshift/web/issues/1570

This is a breaking change to `shapeshiftoss/types`.

This is ready for review, but I'll leave it in draft and hold off on merging until we are closer to handling the breaking changes in `web` and consuming packages in `lib`.


packages/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shapeshiftoss/types",
-  "version": "4.3.0",
+  "version": "5.0.0",
```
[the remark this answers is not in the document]
kaladinlight: Unless we have determined it works correctly with `feat()` I do see the `BREAKING CHANGE` in a commit message, so maybe we are good.

0xdef1cafe:
while we're at it, i think we can remove `sendSupport` and `receiveSupport`

packages/types/package.json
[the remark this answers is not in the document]
0xApotheosis: There is a `BREAKING CHANGE: ` commit in this PR, which will get included in the squashed message: https://github.com/shapeshift/lib/pull/688/commits/59db7fee14b4df56f6766598dc4174051dff6c16

I've also tested major version bumps with `feat()` and it works - an example is https://github.com/shapeshift/lib/pull/676.

packages/types/src/base.ts
```diff
@@ -8,32 +8,14 @@ export type BIP44Params = {
...
-  'cosmos:cosmoshub-4',
-  'cosmos:osmosis-1'
-] as const
-
-export type SupportedChainIds = typeof supportedChainIds[number]
+export const SUPPORTED_CHAIN_IDS = {
+  EthereumMainnet: 'eip155:1',
+  BitcoinMainnet: 'bip122:000000000019d6689c085ae165831e93',
+  CosmosMainnet: 'cosmos:cosmoshub-4',
+  OsmosisMainnet: 'cosmos:osmosis-1'
```
kaladinlight, line 15: Something I have been thinking about with naming related to cosmos assets is including the version number in the key name (ie. `CosmosMainnet4`) to differentiate between mainnet chain upgrades.

packages/types/package.json
```diff
@@ -21,7 +21,11 @@
     "test": "jest test",
     "type-check": "tsc --project ./tsconfig.build.json --noEmit"
   },
+  "peerDependencies": {
```
kaladinlight, line 24: Do the peer/dev dependencies result in circular dependency warning since caip imports types?

packages/types/src/chain-adapters/index.ts
```diff
@@ -1,3 +1,4 @@
+import { btcChainId, cosmosChainId, ethChainId, osmosisChainId } from '@shapeshiftoss/caip'
```
kaladinlight, line 1: Another thought is while we are doing all of this type change/refactor, would this be a good time to move chain-adapters specific types back into the chain-adapter package and ensure only shared types live in the types package?

0xdef1cafe:
closing as stale/redundant as part of today's merge fest

PR #699, opened 2022-05-25
feat(caip): use branded types

0xApotheosis:
A proof of concept demonstrating the use of nominal typing on `AssetId`, `AccountId`, and `ChainId`.

These are currently structurally typed as `string`s, which is not useful as they aren't checked by the compiler. 
An example is outlined below:

```ts
type Brand<K, T> = K & { __brand: T }

type StructuralChainId = string
type BrandedChainId = Brand<string, 'chainId'>

const notAChainId = '1234'

const acceptsStructuralChainIdArg = (chainId: StructuralChainId) => chainId
const acceptsBrandedChainIdArg = (chainId: BrandedChainId) => chainId

acceptsStructuralChainIdArg(notAChainId) // 😞 No type error!
acceptsBrandedChainIdArg(notAChainId) // 😁 Argument of type 'string' is not assignable to parameter of type 'BrandedChainId'
```

---

BREAKING CHANGE: changes AssetId, AccountId, and ChainId types

### DIFF
--- a/packages/caip/src/accountId/accountId.ts
+++ b/packages/caip/src/accountId/accountId.ts
@@ -1,8 +1,9 @@
 import { ChainId, ChainNamespace, ChainReference, fromChainId, toChainId } from '../chainId/chainId'
 import { CHAIN_NAMESPACE } from '../constants'
 import { assertIsChainId, assertIsChainNamespace, assertIsChainReference } from '../typeGuards'
+import { Brand } from '../utils'
 
-export type AccountId = string
+export type AccountId = Brand<string, 'AccountId'>
 
 type ToAccountIdWithChainId = {
   chainId: ChainId
@@ -42,7 +43,7 @@ export const toAccountId: ToAccountId = ({
   const outputAccount =
     chainNamespace === CHAIN_NAMESPACE.Ethereum ? account.toLowerCase() : account
 
-  return `${chainId}:${outputAccount}`
+  return `${chainId}:${outputAccount}` as AccountId
 }
 
 type FromAccountIdReturn = {
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -1,7 +1,7 @@
 import entries from 'lodash/entries'
 import toLower from 'lodash/toLower'
 
-import { fromAssetId } from '../../assetId/assetId'
+import { AssetId, fromAssetId } from '../../assetId/assetId'
 import { ChainId } from '../../chainId/chainId'
 
 const AssetIdToBanxaTickerMap = {
@@ -57,9 +57,9 @@ export const getSupportedBanxaAssets = () =>
  * chain won't be exactly the same as ours.
  */
 const chainIdToBanxaBlockchainCodeMap: Record<ChainId, string> = {
-  'eip155:1': 'ETH',
-  'bip122:000000000019d6689c085ae165831e93': 'BTC',
-  'cosmos:cosmoshub-4': 'COSMOS'
+  ['eip155:1' as ChainId]: 'ETH',
+  ['bip122:000000000019d6689c085ae165831e93' as ChainId]: 'BTC',
+  ['cosmos:cosmoshub-4' as ChainId]: 'COSMOS'
 } as const
 
 /**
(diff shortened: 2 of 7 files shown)


packages/caip/src/accountId/accountId.ts
```diff
@@ -42,7 +43,7 @@ export const toAccountId: ToAccountId = ({
   const outputAccount =
     chainNamespace === CHAIN_NAMESPACE.Ethereum ? account.toLowerCase() : account
 
-  return `${chainId}:${outputAccount}`
+  return `${chainId}:${outputAccount}` as AccountId
```
0xApotheosis, line 46: Forces the implementer to be explicit about the nominal type of return strings.

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -57,9 +57,9 @@ export const getSupportedBanxaAssets = () =>
  * chain won't be exactly the same as ours.
  */
 const chainIdToBanxaBlockchainCodeMap: Record<ChainId, string> = {
-  'eip155:1': 'ETH',
-  'bip122:000000000019d6689c085ae165831e93': 'BTC',
-  'cosmos:cosmoshub-4': 'COSMOS'
+  ['eip155:1' as ChainId]: 'ETH',
+  ['bip122:000000000019d6689c085ae165831e93' as ChainId]: 'BTC',
+  ['cosmos:cosmoshub-4' as ChainId]: 'COSMOS'
```
0xApotheosis, line 62: These type-casted string literals will go away once we properly implement `SUPPORTED_CHAIN_IDS` or similar.

Example implementation in https://github.com/shapeshift/lib/pull/688/commits/a26d0315119e7d7527476890000b529008bd4248#diff-bc09084a5595dff54122b864f8a3718cf0403c148a7b6b667452b710bdaaf8b5

packages/caip/src/constants.ts
```diff
@@ -1,15 +1,15 @@
...
+export const osmosisAssetId = 'cosmos:osmosis-1/slip44:118' as AssetId
 
-export const ethChainId = 'eip155:1'
-export const btcChainId = 'bip122:000000000019d6689c085ae165831e93'
-export const cosmosChainId = 'cosmos:cosmoshub-4'
-export const osmosisChainId = 'cosmos:osmosis-1'
+export const ethChainId = 'eip155:1' as ChainId
+export const btcChainId = 'bip122:000000000019d6689c085ae165831e93' as ChainId
+export const cosmosChainId = 'cosmos:cosmoshub-4' as ChainId
+export const osmosisChainId = 'cosmos:osmosis-1' as ChainId
```
0xApotheosis, line 12: By being explicit here we ensure these strings can't be used as args to functions that can't handle them.

elmutt:
Is this overkill on types?

0xApotheosis:
> I'm uncomfortable with the idea of strongly typing chainId. It means that adding any level of support for another chain can't just be done in a plug-in. The global types have to be updated first.

I don't think that's the case - a nominal type has no concept of _which_ `ChainId`s are supported, only that the strings that are supported _are_ `ChainId`s

0xApotheosis:
> Is this overkill on types?

I don't think so. We use types for a reason and currently, we type strings in a way that does not work, in that they are not enforced by the compiler.

This PR makes string-type aliases actually work instead of them being widened back to a `string` before type-checking.

packages/caip/src/constants.ts
```diff
@@ -1,15 +1,15 @@
...
-export const ethChainId = 'eip155:1'
-export const btcChainId = 'bip122:000000000019d6689c085ae165831e93'
-export const cosmosChainId = 'cosmos:cosmoshub-4'
-export const osmosisChainId = 'cosmos:osmosis-1'
+export const ethChainId: ChainId = 'eip155:1'
+export const btcChainId: ChainId = 'bip122:000000000019d6689c085ae165831e93'
+export const cosmosChainId: ChainId = 'cosmos:cosmoshub-4'
+export const osmosisChainId: ChainId = 'cosmos:osmosis-1'
 
 export const CHAIN_NAMESPACE = {
```
GMSteuart, line 14: I understand this comment is outside of your changes but this caught my attention.

`CHAIN_NAMESPACE` is essentially an enum and I'd consider removing "_NAMESPACE" since `namespace` is a keyword and `Chain` would suffice.

0xApotheosis: @GMSteuart I get where you're coming from here, though in this case the enum represents an array of possible values of type `ChainNamespace`.

The "namespace" nomenclature is one we've chosen to represent the parts of a `ChainId`, which is `ChainNamespace:ChainReference`.

Similarly, we have `AssetNamespace` and `AssetReference`, along with an `AssetId` that encompasses them both.

Agree that there might be an argument for not using a TypeScript keyword in the name, but that ship has sailed.

0xdef1cafe:
closed by https://github.com/shapeshift/lib/pull/774

PR #706, opened 2022-05-26
chore: move types to swapper package & remove unused

elmutt:
- Replace usages of ExecQuoteOutput with TradeResult and remove ExecQuoteOutput
- Move swapper types from types package to swapper

This PR is only moving types around and does not have any functional changes. It is safe for web because web does not refernce any of the moved types


DaoDev44 (approved):
🔥

PR #716, opened 2022-05-31
feat: thorchain memo support

elmutt:
Add a function to return thorchain memo shortened to a max of 80 characters as described:

https://dev.thorchain.org/thorchain-dev/memos


DaoDev44 (dismissed):

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo.test.ts
```diff
@@ -0,0 +1,33 @@
+import { makeSwapMemo } from './makeSwapMemo'
+
+describe('makeSwapMemo', () => {
+  it('make a trade to usdc memo with', () => {
```
kaladinlight, line 4: nitpick that test descriptions start with `should` and this one isn't a complete sentence

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo.ts
```diff
@@ -0,0 +1,43 @@
+import { adapters } from '@shapeshiftoss/caip'
+
+import { SwapError, SwapErrorTypes } from '../../../api'
+
+// BTC (and likely other utxo coins) can only support up to 80 character memos
+const MAX_LENGTH = 80
```
kaladinlight, line 6: We are going to need to think about how to do a reverse fuzzy lookup on abbreviated assets in order to turn it into a caip assetId if there is not a ticket for that already

elmutt: oh thats a fun one

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @returns thorchain memo shortened to a max of 80 characters as described:
```
kaladinlight, line 9: better link to show memo message limits
https://dev.thorchain.org/thorchain-dev/memos#mechanism-for-transaction-intent

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (fullMemo.length <= MAX_LENGTH) return fullMemo
+
+  const fullMemoLength = fullMemo.length
+  const truncateAmount = fullMemoLength - MAX_LENGTH
+  const thorIdSymbol = thorId.slice(0, thorId.indexOf('-'))
+  const thorIdReferenceTruncated = thorId.slice(thorId.length - truncateAmount, thorId.length)
+  const truncatedThorId = `${thorIdSymbol}-${thorIdReferenceTruncated}`
+  const shortenedMemo = `s:${truncatedThorId}:${destinationAddress}:${limit}`
+  return shortenedMemo
```
kaladinlight, line 42: suggestion: thorchain verbiage, comment on abbreviation, catch if we don't have an asset to abbreviate but the memo is too long (not sure how this case would be handled?), and one liner to abbreviate
```suggestion
  const memo = `s:${thorId}:${destinationAddress}:${limit}`

  if (memo.length <= MAX_LENGTH) return memo

  const abbreviationAmount = memo.length - MAX_LENGTH

  // delimeter between ticker and id allowing us to abbreviate the id: https://dev.thorchain.org/thorchain-dev/memos#asset-notation
  const delimeterIndex = memo.indexOf('-')

  if (delimeterIndex === -1) {
    throw new SwapError('[makeSwapMemo] - unable to abbreviate asset, no delimeter found', {
      code: SwapErrorTypes.MAKE_MEMO_FAILED
    })
  }

  return memo.replace(memo.slice(delimeterIndex + 1, delimeterIndex + abbreviationAmount), '')
```

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo.ts
0xdef1cafe: @elmutt i agree with @kaladinlight  and that logic absolutely needs to go into the caip adapter

DaoDev44 (approved):

PR #719, opened 2022-06-01
chore: use map to convert thor id to asset id

elmutt:
use map to convert thor id to asset id and vice versa. instead of deriving it programmatically

This will occasionally need to be updated with 1 line as we add assets to thor swapper


DaoDev44 (dismissed):

packages/caip/src/adapters/thortrading/index.ts
```diff
@@ -1,82 +1,63 @@
...
 
-  switch (chain) {
-    case 'ETH': {
-      if (ticker === 'ETH') return ethAssetId
+const assetIdToPoolAssetIdMap = invert(thorPoolIdAssetIdSymbolMap)
 
-      // We make Eth contract addresses lower case to simplify comparisons
-      const assetReference = contractAddress?.toLowerCase()
-      if (!assetReference) return undefined
+export const poolAssetIdToAssetId = (id: string): string | undefined =>
```
0xdef1cafe, line 59:
```suggestion
export const poolAssetIdToAssetId = (id: string): AssetId | undefined =>
```

kaladinlight (approved):

0xdef1cafe (approved):

Issue #658, opened 2022-05-17
Unchained-Client: Generic approve function parsing


kaladinlight:
### Overview

We are currently detecting approve contract calls against our yearn router contract only, but this is supported by many contracts and should be detected no matter what contract a user is interacting with.

### References and additional details

See the yearn parser for an example of how to instantiate an abi interface, and parse transaction input data

### Acceptance Criteria

- Detect approve contract call for any transaction that matches the function sigHash `approve(address,uint256)`

### Estimated effort

2 days.

0xApotheosis:
Turning into a bounty reserved for @vividwood.

vividwood:
So, I created simple subparser for approve sighash and got following result.
![image](https://user-images.githubusercontent.com/101632255/169588265-486f47b3-61ab-42d7-b2a4-ff9a99a81cdc.png)
But, in my opinion this looks kinda useless without showing an asset. 
So I changed api a little and got this
![image](https://user-images.githubusercontent.com/101632255/169588712-c94968da-c072-4b10-88dd-b2d84d175acd.png)
There is no way to know contract address transaction interacted with, so I had to extend `StandardTxMetadata`. We can discuss other ways to send contract address if we go in this direction. Or I can just stick to first example.
@kaladinlight @0xApotheosis  what do you think?

kaladinlight:
> So, I created simple subparser for approve sighash and got following result. ![image](https://user-images.githubusercontent.com/101632255/169588265-486f47b3-61ab-42d7-b2a4-ff9a99a81cdc.png) But, in my opinion this looks kinda useless without showing an asset. So I changed api a little and got this ![image](https://user-images.githubusercontent.com/101632255/169588712-c94968da-c072-4b10-88dd-b2d84d175acd.png) There is no way to know contract address transaction interacted with, so I had to extend `StandardTxMetadata`. We can discuss other ways to send contract address (like adding `to`) if we go in this direction. Or I can just stick to first (asset-less) example. @kaladinlight @0xApotheosis what do you think?

[...]

PR #731, opened 2022-06-01
feat: generic erc20 approve

vividwood:
Added subparser for ERC20 approve method
Closes https://github.com/shapeshift/lib/issues/658

### DIFF
--- a/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
+++ b/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts
@@ -390,7 +390,7 @@ export class ChainAdapter implements IChainAdapter<ChainTypes.Ethereum> {
           txid: tx.txid,
           ...(tx.data && {
             data: {
-              method: tx.data.method,
+              ...tx.data,
               parser: tx.data.parser ?? 'unknown'
             }
           })
--- a/packages/unchained-client/src/ethereum/parser/__tests__/ethereum.test.ts
+++ b/packages/unchained-client/src/ethereum/parser/__tests__/ethereum.test.ts
@@ -731,7 +731,7 @@ describe('parseTx', () => {
         chainId: 'eip155:1',
         confirmations: tx.confirmations,
         data: {
-          assetId: "eip155:1/erc20:0x470e8de2ebaef52014a47cb5e6af86884947f08c",
+          assetId: 'eip155:1/erc20:0x470e8de2ebaef52014a47cb5e6af86884947f08c',
           method: TxMethod.Approve,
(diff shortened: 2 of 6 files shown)


vividwood:
re @kaladinlight
 
>This has become a little more of a grey area now with the inclusion of unchained-client in lib and essentially providing what we consider a good human readable representation of a tx. I could see the case being made at this point to just use the unchained-client types directly (or we can always create a type alias in chain-adapters if we still want the possible flexibility to change in the future without requiring the unchained-client to change as well). We can then reduce the somewhat confusing duplication of types across packages and make our lives a bit easier.

After some experimentations I ended up just casting `TxMetadata` in `web` and then use type narrowing. I decided to not make any breaking changes in `chain-adapters` because it's way out of scope of this issue.

> The only annoyance is having to add to the TxMetadata union and the exclude list for each new specific tx metadata type. I think it is a place to start though and we can always try to incorporate more typescript magic to make it dynamically driven off a given set of standard and specific tx parsers. I wasn't able to figure out a way easily at first take

Yes and also you have to explicitly set `StandardTxMetadata` in other parsers, which is annoying. I actually just moved `TxMetadata` to `ethereum/types` (like in `cosmos`) and it works just fine.

vividwood:
`web` https://github.com/shapeshift/web/pull/1921

kaladinlight:
> re @kaladinlight
> 
> > This has become a little more of a grey area now with the inclusion of unchained-client in lib and essentially providing what we consider a good human readable representation of a tx. I could see the case being made at this point to just use the unchained-client types directly (or we can always create a type alias in chain-adapters if we still want the possible flexibility to change in the future without requiring the unchained-client to change as well). We can then reduce the somewhat confusing duplication of types across packages and make our lives a bit easier.
> 
> After some experimentations I ended up just casting `TxMetadata` in `web` and then use type narrowing. I decided to not make any breaking changes in `chain-adapters` because it's way out of scope of this issue.
> 
> > The only annoyance is having to add to the TxMetadata union and the exclude list for each new specific tx metadata type. I think it is a place to start though and we can always try to incorporate more typescript magic to make it dynamically driven off a given set of standard and specific tx parsers. I wasn't able to figure out a way easily at first take
> 
> Yes and also you have to explicitly set `StandardTxMetadata` in other parsers, which is annoying. I actually just moved `TxMetadata` to `ethereum/types` (like in `cosmos`) and it works just fine. I created new `EthereumTxParser` enum (named it so to avoid confusion with `TxParser`) which can be used for new specific parsers

This all looks and sounds great. Thanks and sorry for the delay in review!

Issue #721, opened 2022-06-01
Create skeleton CowSwapper api


chxash:
### Overview

Create `CowSwapper.ts` in the correct directory structure `cow/ ` (same as ZrXSwapper), implementing interface Swapper with dummy methods.
Create unit test in `SwapperManager.test.ts` about `CowSwapper`

### References and additional details

https://github.com/shapeshift/lib/issues/702 for additional details

### Acceptance Criteria

`CowSwapper.ts` implements `Swapper` interface with dummy methods

### Estimated effort

2 hours

PR #732, opened 2022-06-02
feat: implements cowswapper skeleton

chxash:
This adds a skeleton for CowSwapper interface implementation of  `Swapper` api with dummy methods.
Not much to test for the moment except that the few unit tests are green.

Related issue : https://github.com/shapeshift/lib/issues/721

### DIFF
--- a/packages/swapper/src/api.ts
+++ b/packages/swapper/src/api.ts
@@ -137,6 +137,7 @@ export type ApprovalNeededOutput = {
 export enum SwapperType {
   Zrx = '0x',
   Thorchain = 'Thorchain',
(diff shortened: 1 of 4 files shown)


0xdef1cafe (approved):
LGTM - just make sure this doesn't get registered in web and the client attempt to use it

chxash:
> LGTM - just make sure this doesn't get registered in web and the client attempt to use it

Using it in web will be our very last task for CowSwap for sure !

DaoDev44 (approved):
LGTM

elmutt (approved):

Issue #722, opened 2022-06-01
CowSwap : Create cowService and use it in getUsdRate method implementation


chxash:
### Overview

We need to create `cowService` in order to use `CowSwap` api.
First use will be to implement the `getUsdRate` method of `Swapper` interface for `CowSwapper`

### References and additional details

https://github.com/shapeshift/lib/issues/702 for additional details

endpoint for `getUsdRate` : https://api.cow.fi/mainnet/api/v1/markets/{baseToken}-{quoteToken}/{kind}/{amount}

### Acceptance Criteria

- `cowService` is implemented
- `getUsdRate` method is implemented
- unit tests

### Estimated effort

4 hours

PR #733, opened 2022-06-02
feat: implements getUsdRate method

chxash:
This implements `getUsdRate()` method in CowSwapper for `Swapper` interface.
Related issue is : https://github.com/shapeshift/lib/issues/722

To be able to test this : 
- Verify the unit test linked to this change are green
- Use swapcli with any parameter after having added the following lines into `swappercli.ts`

```
import { CowSwapper } from './swappers/cow/CowSwapper'

const cowSwapperDeps = {
    apiUrl: 'https://api.cow.fi/mainnet/api/'
  }

  const cowSwapper = new CowSwapper(cowSwapperDeps)
  const testFox = await cowSwapper.getUsdRate(
    assetMap['eip155:1/erc20:0xc770eefad204b5180df6a14ee197d99d808ee52d']
  )
  const testWbtc = await cowSwapper.getUsdRate(
    assetMap['eip155:1/erc20:0x2260fac5e5542a773aa44fbcfedf7c193bc2c599']
  )
```


packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+import { bn, bnOrZero } from '../../../utils/bignumber'
+import { CowSwapperDeps } from '../../CowSwapper'
+import { CowSwapPriceResponse } from '../../types'
+import { cowService } from '../cowService'
+
+const USDC_CONTRACT_ADDRESS = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+
+export const getUsdRate = async (
+  { assetService }: CowSwapperDeps,
+  input: Pick<Asset, 'symbol' | 'assetId'>
```
DaoDev44, line 15:
```suggestion
  input: Pick<Asset, 'assetId'>
```

packages/swapper/src/swappers/cow/utils/cowService.ts
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios'
+
+const axiosConfig = {
+  baseURL: 'https://api.cow.fi/mainnet/api/',
```
DaoDev44, line 4: Does cowswap support any other networks other than mainnet? If so, it may be a good idea to pass the baseUrl as a dep so it can be set by web instead of hard coding it here. It may be out of the scope of this PR (given this is how zrxSwapper does it), but it might be worth looking at.

here is a reference of how it's being done in [thorchainSwapper](https://github.com/shapeshift/lib/blob/5cd7af8106c3d7e3bc619e993d2f1a85ba4f4a11/packages/swapper/src/swappers/thorchain/utils/getUsdRate/getUsdRate.ts#L25)

packages/swapper/src/swappers/cow/utils/helpers/helpers.test.ts
```diff
@@ -0,0 +1,114 @@
+import { AssetService } from '@shapeshiftoss/asset-service'
+import { ChainTypes, NetworkTypes } from '@shapeshiftoss/types'
+
+import { FOX, WBTC } from '../../../utils/test-data/assets'
+import { CowSwapperDeps } from '../../CowSwapper'
+import { cowService } from '../cowService'
+import { getUsdRate } from '../helpers/helpers'
+
+const axios = jest.createMockFromModule('axios')
```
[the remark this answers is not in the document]
chxash: Removed ! thanks !

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
chxash: Fixed !

packages/swapper/src/swappers/cow/utils/cowService.ts
chxash: Indeed, there are other networks supported. I did similar thing as in thorchainSwapper.

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const asset = assetService.byTokenId({
+    chain: ChainTypes.Ethereum,
+    network: NetworkTypes.MAINNET,
+    tokenId: erc20Address
+  })
+
+  try {
+    // rate is imprecise for low $ values, hence asking for $1000
```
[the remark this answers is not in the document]
elmutt: maybe best to define 1000 in a named const for clarity and derive all values from that (its not clear what 1000000000 represents for example)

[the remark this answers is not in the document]
chxash: > How is it imprecise for low values? That might be ok? $1000 feels like an arbitrary value

I chose $1000 through trial and testing, I corrected similar bug couple months ago for Zrx where getUsdRate() was returning very imprecise values.
Extracted 1000 to a const for more clarity !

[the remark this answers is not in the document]
elmutt: Can you give me an example of the price difference? im curious now

I remember seeing that code in zrx swapper when we were cleaning it up and I didnt feel like it was necessary for zrx because its an AMM and large amounts shouldnt make much different. 

Buit maybe its more important for cowswap because it uses an order book and small prices can be very far off? I just want to make sure I fully understand the implications of changing this amount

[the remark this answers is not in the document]
chxash: @elmutt 
One example with cowswap : 

WBTC value currently : **$29522**

** usdRate fetching $1

Api call : https://api.cow.fi/mainnet/api//v1/markets/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48-0x2260fac5e5542a773aa44fbcfedf7c193bc2c599/buy/1000000

Api response : {"amount":"3277","token":"0x2260fac5e5542a773aa44fbcfedf7c193bc2c599"}
usdRate calculation leads to **WBTC 30515.71**

** usdRate fetching $1000
Api call : https://api.cow.fi/mainnet/api//v1/markets/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48-0x2260fac5e5542a773aa44fbcfedf7c193bc2c599/buy/1000000000

Api response : {"amount":"3389831","token":"0x2260fac5e5542a773aa44fbcfedf7c193bc2c599"}
usdRate calculation leads to **WBTC 29499.99**

I witnessed same behaviour with multiple tokens (FOX etc).

CowSwap ultimately will look into 0x if needed so there is a good chance the same behaviour on 0x is simply present on CowSwap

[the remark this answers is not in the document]
elmutt: I dont think its a "bug" for 0x to use an amount of $1 because the price is always going to be different depending on the amount. Im think for 0x using $1000 is just as arbitrary as $1 -- It will ee wrong if they do am amount higher or lower than $1000. The price would be correct for a trade amount of X but its likely the user does a different amount than we are using to estimate the price.

Ultimately usdRate is only used as an estimate and its impossible for to be perfect without specifying an amount. Using $1000 here is fine if you feel it gets us the closest estimate


Its strange to me that the cowswap rate gets better as amount increases, I would expect it to be the opposite because of slippage. Do you know what that is?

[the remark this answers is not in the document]
chxash: I agree with you that $1 is not any more "right" than $1000, but ultimately some api calls will be more relevant regarding the usdRate estimation we want, and that's what I looked for ...

For the rate getting better as amount increase, it's really an effect I noted going from $1 to $1000 (for both 0x and cowswap). I would say slippage is minimal for those "low" values of liquid coins like WBTC (also it seems slippage is not taken into account in this api from their doc : https://docs.cow.fi/off-chain-services/api/price-estimation)

I am not sure but I think it's more a fees problem.  Since fees are taken into account with the amount of token we can expect from a buy, it may make sense that fee represent a very large percentage of the transaction value for tiny values on ETH network, hence this bad estimation we are seeing with low value orders. 

I will certainly get more information as I dig more into the cowswap api ...

elmutt (dismissed):

elmutt:
Looks great so far! approved (noticed some conficts, thats probably my fault as I removed getMinMax from the api but should be an easy fix)

chxash:
@elmutt no issue for the conflicts, I rebased main into this branch

packages/swapper/src/swappers/cow/utils/helpers/helpers.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { FOX, WBTC } from '../../../utils/test-data/assets'
+import { CowSwapperDeps } from '../../CowSwapper'
+import { cowService } from '../cowService'
+import { getUsdRate } from '../helpers/helpers'
+
+jest.mock('../cowService')
+jest.mock('@shapeshiftoss/asset-service')
+
+// @ts-ignore
```
0xdef1cafe, line 11: do this properly please

```
jest.mock('@shapeshiftoss/asset-service')
const mockAssetService = jest.mocked('assetService', true)
mockAssetService.mockImplementation(...
```

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { assetReference: erc20Address, assetNamespace } = fromAssetId(assetId)
+
+  if (assetNamespace !== 'erc20') {
+    throw new SwapError('[getUsdRate] - unsupported asset namespace', {
+      code: SwapErrorTypes.USD_RATE_FAILED,
+      details: { assetNamespace }
+    })
+  }
+
+  const assets = assetService.getAll()
```
0xdef1cafe, line 29: if this is why we added the asset service dependency, don't do it this way.

in the near term future, the client will be aware of more assets than are statically defined in the service.

just get the assets from the store in the client (which is the source of truth) and pass that into the swapper instead

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const getUsdRate = async (
+  { assetService, apiUrl }: CowSwapperDeps,
+  input: Pick<Asset, 'assetId'>
+): Promise<string> => {
+  const { assetId } = input
+
+  const { assetReference: erc20Address, assetNamespace } = fromAssetId(assetId)
+
+  if (assetNamespace !== 'erc20') {
```
0xdef1cafe, line 22: this will fail on ETH

0xdef1cafe: check for `slip44` too and comment as to why

0xdef1cafe (changes requested):

packages/swapper/src/swappers/cow/utils/helpers/helpers.test.ts
chxash: Removed asset-service dependency so this line does not exist anymore

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
chxash: Fixed this by adding `assets` dependency to swapper

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
chxash: I did this on purpose so it would fail on ETH, because only WETH is supported on CowSwap (ETH needs to be wrapped to WETH before doing anything on CowSwap). Please suggest if I did something wrong, should I return the USD rate of WETH when called with ETH ?

0xdef1cafe: ah, in that case we're all good

packages/swapper/src/swappers/cow/CowSwapper.ts
```diff
@@ -37,9 +52,8 @@ export class CowSwapper implements Swapper {
     throw new Error('CowSwapper: getTradeQuote unimplemented')
   }
 
-  getUsdRate(input: Pick<Asset, 'symbol' | 'assetId'>): Promise<string> {
-    console.info(input)
-    throw new Error('CowSwapper: getUsdRate unimplemented')
+  async getUsdRate(input: Pick<Asset, 'symbol' | 'assetId'>): Promise<string> {
```
0xdef1cafe, line 55: looking at this further, you can simplify this more

https://github.com/shapeshift/lib/blob/main/packages/swapper/src/api.ts#L192

just pass the full asset in as the input, like the base class does, and then you can pull the precision from the asset itself, and you don't need to inject all the assets, just to take the asset id, to look up the full asset again...

0xdef1cafe (changes requested):

packages/swapper/src/swappers/cow/CowSwapper.ts
chxash: Oops yes ! done !

DaoDev44 (approved):

0xdef1cafe (dismissed):

DaoDev44 (dismissed):

PR #738, opened 2022-06-02
feat: support coingecko pro with api key

0xdef1cafe:
we we're getting rate limited by coingecko - add support for using the coingecko pro url and
injecting the api key if we have one


0xean (approved):
LGTM

0xean (approved):

PR #741, opened 2022-06-02
fix: dependency inject market service manager

0xdef1cafe:
we were building packages that were depending on `process.env.FOO`

`process.env` is not available in the browser, and isn't replaced by webpack in the build process in
web.

the impact of this is that we aren't able to pass API keys into the market service at runtime - this refactor to dependency injection allows us to do that.

the market service and yearn sdk were the only things that actually require environment variables
that will have user facing impacts. the yearn market service is currently being rewritten anyway, and doesn't have any user facing bugs.

unchained-client also wants a log level, but this is low priority for now

the rest of the references to `process.env` are CLI/dev tools that are running locally in the
context of node


cjthompson (approved):

theoboldfrazier (dismissed):
Looks good!

0xdef1cafe:
merging as paired with @gomesalexandre and @theoboldfrazier - previously approved by @cjthompson architecturally and tests are updated and passing

0xApotheosis:
This was a breaking change to `web`, though CI still passing thanks to caching. 

Change handled as part of https://github.com/shapeshift/web/pull/1908 in https://github.com/shapeshift/web/pull/1908/commits/7de0ec154b0fd5958b5c47d0d638e9242e116a98 - or we can cherry-pick for expediency.

PR #712, opened 2022-05-30
feat: normalize asset service

0xApotheosis:
This is a placeholder PR to track changes - we might want to break this up by package and/or extend upon it.

- Nomalizes `Asset` data
- Removes unused fields
- Removes `BaseAsset` and `TokenAsset`, consolidated with `Asset`
- Uses static asset data in the Asset Service

Contributes to https://github.com/shapeshift/lib/issues/711

TODO:

- Create PRs updating each consuming packages
- `BREAKING CHANGES: ` commit

---

Note: this is a breaking change to:

- `@shapeshiftoss/asset-service`
- `@shapeshiftoss/chain-adapters`
- `@shapeshiftoss/swapper`
- `@shapeshiftoss/types`

Issue #711, opened 2022-05-30
Normalize asset data


0xApotheosis:
### Overview

We have both used and redundant data stored on our assets - remove these.

### References and additional details

Examples include these properties on `AbstractAsset`: `dataSource`, `silp44`, `secondaryColor`, `sendSupport`, `receiveSupport`.

We can consolidate the following types into `Asset`: `TokenAsset`, `TokenAsset`, `AbstractAsset`.

### Acceptance Criteria

All redundant/unneeded properties and types are removed.

PR #743, opened 2022-06-03
feat(types): remove Asset properties for normalization

0xApotheosis:
- Nomalizes `Asset` data
- Removes `BaseAsset` and `TokenAsset`, consolidated with `Asset`

This is the first of many PRs that will follow across `lib` packages and `web` to consume this change.

Contributes to https://github.com/shapeshift/lib/issues/711

This is an extract from the work started by @0xdef1cafe in https://github.com/shapeshift/lib/pull/712

This PR contains breaking changes to `@shapeshiftoss/types`.

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shapeshiftoss/types",
-  "version": "4.3.2",
+  "version": "5.0.0",
   "description": "Common types shared across packages",
   "repository": "https://github.com/shapeshift/lib",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


packages/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shapeshiftoss/types",
-  "version": "4.3.2",
+  "version": "5.0.0",
```
gomesalexandre, line 3: Do we need this?

gomesalexandre (approved):
LGTM, one concern WRT manual bumping of types version but otherwise looks good

packages/types/package.json
cjthompson: semantic release will overwrite this when it does the version bump. It's needed so that the yarn install at the beginning of the build won't symlink the package for the other packages that still want the 4.x version.

If we don't update it like this then the build will fail.

0xApotheosis: This ^ it gives us the green dot in CI so we can check that nothing _inside_ the types package broke.

kaladinlight (approved):

PR #748, opened 2022-06-03
chore: remove swapper unused

elmutt:
Removed unused code from swapper:

- GetMinMax is not used by web and can be removed from the interface (min/max comes back with quote)
- ApprovalNeeded should not be returning gas info


kaladinlight (approved):

theoboldfrazier (approved):
Looks good!

PR #750, opened 2022-06-03
fix: pick assetId instead of tokenid from asset

elmutt:
fix type


0xApotheosis (approved):

0xApotheosis (changes requested):
I'm going to work out how to fix CI whilst you're out @elmutt

0xdef1cafe:
fixed in https://github.com/shapeshift/lib/pull/744

Issue #723, opened 2022-06-01
CowSwap : getMinMax / filterBuyAssetsBySellAssetId / filterAssetIdsBySellable methods implementation


chxash:
### Overview

We need to implement those methods in ` CowSwapper` similarly to what has been done for `ZrxSwapper`.
Notable difference is we need to filter out ETH asset because only WETH is supported on CowSwap.
This is dependent on https://github.com/shapeshift/lib/issues/722.

### References and additional details

See https://github.com/shapeshift/lib/issues/702 for additional details

### Acceptance Criteria

- All methods are implemented
- All methods are unit tested

### Estimated effort

3 hours

PR #757, opened 2022-06-07
feat: cowswap getcowswapminmax and filtering methods implementation

chxash:
This implements the outlined method. `getCowSwapMinMax` method is not yet used but will be used when implementing `getTradeQuote` method.

closes https://github.com/shapeshift/lib/issues/723

No real testing possible except seeing unit tests for those methods are green / correct.


packages/swapper/src/swappers/cow/CowSwapper.ts
```diff
@@ -69,13 +70,25 @@ export class CowSwapper implements Swapper {
   }
 
   filterBuyAssetsBySellAssetId(args: BuyAssetBySellIdInput): AssetId[] {
-    console.info(args)
-    return []
+    const { assetIds = [], sellAssetId } = args
+    if (
+      !sellAssetId?.startsWith('eip155:1/erc20') ||
```
elmutt, line 75: Is there a way to ask cowswap what assets it supports?

We had to do it this way for zrx because theres no good way to get all the assets it supports from an api.

But if cowswap has an api with all supported assets it would be better to load those up in the initialize() function and do it how we do with thorchain swapper

chxash: I did not find a way to fetch supported assets but will look again

chxash: @elmutt unfortunately it seems there is no api / no easy way to get supported assets for cowswap

elmutt (approved):
Looks good to me

0xdef1cafe (approved):

packages/swapper/src/swappers/cow/getCowSwapMinMax/getCowSwapMinMax.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getCowSwapMinMax = async (
+  deps: CowSwapperDeps,
+  sellAsset: Asset,
+  buyAsset: Asset
+): Promise<MinMaxOutput> => {
+  try {
+    if (
+      !sellAsset.assetId.startsWith('eip155:1/erc20') ||
+      !buyAsset.assetId.startsWith('eip155:1/erc20')
```
0xdef1cafe, line 17: this is not ideal coupling this implementation to mainnet. cowswap supports gnosis too and we should be considering this when implementing

PR #758, opened 2022-06-07
feat: remove `ChainTypes`

0xApotheosis:
**WIP**

- Move `chain-adapter` types out of the `shapeshiftoss/types` package and into `shapeshiftoss/types`. This is needed to prevent circular dependencies
- Added nominal typing to CAIP types
- Replace `ChainTypes` with `ChainNamespace`
- Make `chain` and `network` Asset properties optional, and remove respective data from mocks
- Replace `SupportedChainId` chain-adapter generic type with something more generic so that any valid `ChainId` can be used
- Remove `ChainAdapterManager` - it's not actually needed

**TODO**

- De-structure `chain-adapter` types into common directories
- Remove "chain type" vernacular from affected code
- Update READMEs
- Address FIXMEs

**Reviewers**

Key changed files:

- `packages/chain-adapters/src/types/index.ts` - `ChainTypes` changed to `SupportedChainId`
- `packages/chain-adapters/src/ChainAdapterManager.ts` - switch on `SupportedChainId`.

### DIFF
--- a/packages/caip/src/assetId/assetId.test.ts
+++ b/packages/caip/src/assetId/assetId.test.ts
@@ -648,14 +648,14 @@ describe('assetId', () => {
       expect(() => fromAssetId('invalid')).toThrow()
     })
 
-    it('errors for invalid chaintype', () => {
+    it('errors for invalid ChainNamespace', () => {
       expect(() => fromAssetId('invalid:cosmoshub-4/slip44:118')).toThrow()
     })
 
-    it('errors for invalid network type', () => {
+    it('errors for invalid ChainReference type', () => {
       expect(() => fromAssetId('cosmos:invalid/slip44:118')).toThrow()
     })
-    it('errors for invalid osmosis asset namespace', () => {
+    it('errors for invalid osmosis AssetNamespace', () => {
       expect(() => fromAssetId('cosmos:osmosis-1/invalid:118')).toThrow()
     })
   })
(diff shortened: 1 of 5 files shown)


packages/chain-adapters/src/ChainAdapterManager.ts
```diff
@@ -81,15 +99,19 @@ export class ChainAdapterManager {
...
     if (typeof chain !== 'string' || typeof factory !== 'function') {
       throw new Error('Parameter validation error')
     }
     this.supported.set(chain, factory)
   }
 
-  removeChain<T extends ChainTypes>(chain: T): void {
-    if (!Object.values(ChainTypes).includes(chain)) {
+  removeChain<T extends SupportedChainId>(chain: T): void {
+    // FIXME: This is gross
```
kaladinlight, line 113: do we really even need this? T is already constrained to `SupportedChainId` and seem redundant (and gross as mentioned) to do this additional check

gomesalexandre: Yeah this logic should live in a type guard, but if you move it to a type guard, you realize it should always return true according to the typings

packages/types/src/base.ts
```diff
@@ -15,14 +15,14 @@ export enum ChainTypes {
   Osmosis = 'osmosis'
 }
 
-const supportedChainIds = [
-  'eip155:1',
-  'bip122:000000000019d6689c085ae165831e93',
-  'cosmos:cosmoshub-4',
-  'cosmos:osmosis-1'
-] as const
+export enum SupportedChainIds {
```
gomesalexandre, line 18: I don't think `supportedChainIds`/`SupportedChainIds` is necessarily bad for now, plus we now have them indexed with a respective enum key.

packages/chain-adapters/src/ChainAdapterManager.ts
0xApotheosis: We killed `ChainAdapterManager` entirely - problem solved!

PR #772, opened 2022-06-09
refactor: add needed peer dependencies to the workspace

cjthompson:
This gets rid of warnings in yarn that there are missing peer dependencies.

Note that the yarn.lock didn't change because these dependencies were already being installed as devDependencies in each package. This change doesn't actually install any additional packages, it just shuts up `yarn`.


gomesalexandre (approved):
Tested locally, LGTM, `yarn`, `yarn build` and `yarn local-ci` still works fine.

0xdef1cafe (approved):

PR #773, opened 2022-06-09
feat(types): remove chainAdapters types

cjthompson:
* change "supportedChainIds" to an enum "KnownChainIds"

BREAKING CHANGE: Removed `chainAdapters` types. Those will now be in the `chain-adapters` package.


packages/types/src/base.ts
```diff
@@ -15,14 +15,13 @@ export enum ChainTypes {
...
-const supportedChainIds = [
-  'eip155:1',
-  'bip122:000000000019d6689c085ae165831e93',
-  'cosmos:cosmoshub-4',
-  'cosmos:osmosis-1'
-] as const
-
-export type SupportedChainIds = typeof supportedChainIds[number]
+// todo; change to object, call named chainIds
+export enum KnownChainIds {
```
kaladinlight, line 19: Why not change to object and call the desired name at this time?

cjthompson: for static typing reasons.

0xdef1cafe (approved):

packages/types/src/base.ts
kaladinlight: What is the comment about then? I just don't see why if we are making a breaking change and intend on it being an object with a different name we wouldn't just do that? Or delete the comment if it is irrelevant

cjthompson: I removed the comment. I think it was leftover from experimenting.

PR #776, opened 2022-06-09
feat(chain-adapters): replace ChainTypes with ChainId

cjthompson:
* move all ChainAdapters types from "types" package into "chain-adapters"

BREAKING CHANGE: Removed `ChainAdapterManager`. `ChainAdapterManager` wasn't doing anything a regular Map doesn't do.


elmutt (approved):

0xdef1cafe (approved):

packages/chain-adapters/src/index.ts
```diff
@@ -1,6 +1,6 @@
-export * from './ChainAdapterManager'
 export * from './api'
 export * from './utils'
+export * as chainAdapters from './types'
```
kaladinlight, line 3: I had mentioned this to @0xApotheosis but the way we are exporting types makes no sense imo.

My suggestion was:
`packages/chain-adapters/src/types.ts`
`packages/chain-adapters/src/bitcoin/types.ts`

and export the types as:
`packages/chain-adapters/src/index.ts
```
export * from './types'
```

`packages/chain-adapters/bitcoin/index.ts`
```
export * from './types`
```

That way they are associated with the appropriate chains and you would be able to just grab the types off the desired location

```
import { BaseTx, ethereum } from '@shapeshiftoss/chain-adapters`

type EthereumTx = ethereum.Tx

PR #791, opened 2022-06-12
chore(chain-adapters): refactor types and add `ChainNamespace`d exports

0xApotheosis:
As per Kevin's suggestion [here](https://github.com/shapeshift/lib/pull/776#discussion_r893837514), refactor type exports to group them by ChainNamespace and allow imports by the same.

For example, instead of this:

```ts
import { chainAdapters } from '@shapeshiftoss/chain-adapters'

type EthereumAccount = chainAdapters.ethereum.Account
```

We can now do:

```ts
import { ethereum } from '@shapeshiftoss/chain-adapters'

type EthereumAccount = ethereum.Account
```

Note this is not a breaking change, as it is backwards compatible.


packages/chain-adapters/src/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from './api'
 export * from './utils'
-export * as chainAdapters from './types'
+export * from './types'
```
gomesalexandre, line 3: If we want this not to be a breaking change, we should have both the previous line (exporting all the modules as a `chainAdapters` export) and this new one (exporting all exports from this module).

gomesalexandre (changes requested):
Checked this one out, CI is very happy.
Added one comment WRT an introduced breaking change, although from looking at web the `chainAdapters` export is not consumed in web currently, so I'm happy to accept this if we just update the description to reflect it.

elmutt (approved):

packages/chain-adapters/src/types.ts
```diff
@@ -9,9 +9,9 @@ import {
 import { BIP44Params, ChainSpecific, KnownChainIds, UtxoAccountType } from '@shapeshiftoss/types'
 
 import * as bitcoin from './bitcoin'
-import * as cosmos from './cosmos'
+import * as cosmos from './cosmossdk/cosmos'
+import * as osmosis from './cosmossdk/osmosis'
 import * as ethereum from './ethereum'
-import * as osmosis from './osmosis'
 
 export { bitcoin, cosmos, ethereum }
```
kaladinlight, line 16:
```suggestion
```

gomesalexandre: That would be a breaking change, so the PR would need to be updated to reflect this

packages/chain-adapters/src/index.ts
kaladinlight: My vote is update the description and carry on seeing as how the destructured import is not being used in web yet.

0xApotheosis: Agree. Updated accordingly.

packages/chain-adapters/src/types.ts
0xApotheosis: https://github.com/shapeshift/lib/pull/791/commits/eb85b288cc9bb392632a8f3d2ae92ff865f09791

0xApotheosis:
@gomesalexandre @kaladinlight redundant exports removed in https://github.com/shapeshift/lib/pull/791/commits/5869c357b73e21ea90db462a5c4d7f372c4fd695 and handled in https://github.com/shapeshift/lib/pull/791/commits/08fbef3b64b41346e448a7685e20928be7d2bfe6

gomesalexandre (approved):
LGTM, CI is still very happy 🍬

cjthompson (approved):

PR #797, opened 2022-06-13
chore(build): add support for next branch

cjthompson:


0xdef1cafe:
seems like a good improvement

@cjthompson my main question is who will be the owner of the `next` branch, such that it doesn't "become master" with so many required breaking changes?

0xApotheosis (approved):
I'm keen to give this a go. 

If we can avoid making `lib`s `main` unlinkable with `web`s `develop` we are in a good place.

0xdef1cafe (approved):

PR #798, opened 2022-06-13
fix(market-service): remove hard-coded chain adapter URLs

cjthompson:
We were already providing the ethereum URLs to the `MarketDataManager` so we just need to use those.


0xdef1cafe (approved):

0xApotheosis (approved):
Looks good, though that `yarn.lock` change probably doesn't belong here.

cjthompson:
> Looks good, though that `yarn.lock` change probably doesn't belong here.

It's because you bumped `types` to 6.0.0 so the requested version of 5.0 can't be linked. I could push just the yarn.lock change to main.

PR #799, opened 2022-06-13
feat(caip): added coinbase ticker/asset mapping

theoboldfrazier:
BREAKING CHANGE: updates function names for consistency


packages/caip/src/adapters/coinbase/getCoinbasePayAssets.ts
```diff
@@ -0,0 +1,67 @@
+import axios from 'axios'
+import toLower from 'lodash/toLower'
+
+import { AssetId } from '../../index'
+
+/** Helper module to generate the coinbase supported assets map in the sibling index file */
```
gomesalexandre, line 6: How are we using this, is it supposed to be used as a CLI of sorts? If so, we can source `assetIdToCoinbaseAssetIdMap` from a JSON file instead and `fs.writeFile` it or similar

gomesalexandre (approved):
LGTM overall but added some comments about typings and the use of the added `packages/caip/src/adapters/coinbase/getCoinbasePayAssets.ts` module

packages/caip/src/adapters/coinbase/getCoinbasePayAssets.ts
```diff
@@ -0,0 +1,67 @@
...
+    push_payment_methods: string[]
+    group_types: string[]
+    display_name: string | null
+    processing_time_seconds: string | null
+    min_withdrawal_amount: number | null
+    max_withdrawal_amount: number | null
+  }
+  default_network: string
+  supported_networks: string[]
+}
```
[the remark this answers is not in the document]
theoboldfrazier: Yeah, that makes sense. Done.

packages/caip/src/adapters/coinbase/getCoinbasePayAssets.ts
theoboldfrazier: Originally, I added this code because I had it in web and I didn't want to lose the functions used to generate the object.

I decided to add a generate script around this module and wrote the output into a json file that is used as the map.

elmutt (approved):

PR #817, opened 2022-06-14
feat: add getGasFeeData for Eth chain adapter

### LINKED ISSUE #786: Fallback gas calculation for thor eth trades
### Overview

Our eth thor trading gas calculation function call can fail in some cases causing quote to fail.

1) add gasPrice / eip1559 function to eth chain adapters to return gas price info (gasprice / eip1559 variables)
2) catch failure case for from existing gas call. If fails, fall back to a known good hard coded gas limit * the new gas price we added to chain adapters

Also add eip1559 info to the thor eth gas utility function

Our gas calculation call can fail in some cases causing quote to fail.

1) add gasPrice / eip1559 function to eth chain adapters to return gas price info (gasprice / eip1559 variables)
2) catch failure case for from existing gas call. If fails, fall back to a known good hard coded gas limit * the new gas price we added to chain adapters

Also add eip1559 info to the thor eth gas utility function

### Acceptance Criteria

Our gas calculation call can fail in some cases causing quote to fail.

1) add gasPrice / eip1559 function to eth chain adapters to return gas price info (gasprice / eip1559 variables)
2) catch failure case for from existing gas call. If fails, fall back to a known good hard coded gas limit * the new gas price we added to chain adapters

Also add eip1559 info to the thor eth gas utility function

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

DaoDev44:
- add getGasFeeData to EthereumChainAdapter to return just gas fee info. This will be used to resolve an issue where eth thor trading gas calculation function call can fail in some cases causing the estimated quote to fail.

closes #786


elmutt (approved):

kaladinlight:
Both this and `buildCustomTx` are big duplications. If not improving the existing functions to handle the slight differences, can we at least re-use the logic added here in the `getFeeData` function? I just feel like these additions are just an easy route but not the most ideal.

DaoDev44:
@kaladinlight After giving it a second look, you're right. Re using the logic that is duplicated inside of getFeeData is a better way to handle this. I'll make that change.

elmutt (approved):

kaladinlight (approved):

PR #820, opened 2022-06-15
fix: expose asset-service utils in dist

gomesalexandre:
### Description

This exposes the `utils` module of `asset-service` following 94f74495b.

It is currently not exposed in the built package and thus, its consumption produces a runtime error in:
https://github.com/shapeshift/lib/blob/ef77d1dfcbc969c390320dea38e9fa4c52513c17/packages/asset-service/src/service/AssetService.ts#L5

### Issue

Relates to https://github.com/shapeshift/lib/issues/444
Relates to https://github.com/shapeshift/lib/issues/556

### DIFF
--- a/packages/asset-service/package.json
+++ b/packages/asset-service/package.json
@@ -7,9 +7,7 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "files": [
-    "dist/index.*",
-    "dist/service",
-    "dist/utils.ts"
+    "*.js"
   ],
   "publishConfig": {
     "access": "public"
--- a/packages/asset-service/package.json
+++ b/packages/asset-service/package.json
@@ -8,7 +8,8 @@
   "types": "./dist/index.d.ts",
   "files": [
     "dist/index.*",
-    "dist/service"
+    "dist/service",
+    "dist/utils.ts"
   ],
   "publishConfig": {
     "access": "public"


elmutt (dismissed):

kaladinlight:
I am confused by this PR. `utils.ts` is being built and we are importing directly from that file in `AssetService.ts`. Exporting it from `index.js` would certainly expose it to the web client if that is the intention, but I don't see it being used in there.

gomesalexandre:
> I am confused by this PR. `utils.ts` is being built and we are importing directly from that file in `AssetService.ts`. Exporting it from `index.js` would certainly expose it to the web client if that is the intention, but I don't see it being used in there.

Yeah, the intent can get quite confusing.
Basically, `utils.ts` is missing from the dist, and as such, when the dist version of `asset-service` tries to access it (in e.g web), this is the error we're getting:

```
ERROR in ./node_modules/@shapeshiftoss/asset-service/dist/service/AssetService.js 18:16-35

Module not found: Error: Can't resolve '../utils' in '/Users/alexandre.gomes/Sites/shapeshiftWeb/node_modules/@shapeshiftoss/asset-service/dist/service'
```

The fact that `asset-service` *builds* without problems doesn't mean that *the build* will actually work: if the file isn't available to consume for the dist version, the package won't work when actually installed.

kaladinlight:
> > I am confused by this PR. `utils.ts` is being built and we are importing directly from that file in `AssetService.ts`. Exporting it from `index.js` would certainly expose it to the web client if that is the intention, but I don't see it being used in there.
> 
> Yeah, the intent can get quite confusing. Basically, `utils.ts` is missing from the dist, and as such, when the dist version of `asset-service` tries to access it (in e.g web), this is the error we're getting:
> 
> ```
> ERROR in ./node_modules/@shapeshiftoss/asset-service/dist/service/AssetService.js 18:16-35
> 
> Module not found: Error: Can't resolve '../utils' in '/Users/alexandre.gomes/Sites/shapeshiftWeb/node_modules/@shapeshiftoss/asset-service/dist/service'
> ```
> 
> The fact that `asset-service` _builds_ without problems doesn't mean that _the build_ will actually work: if the file isn't available to consume for the dist version, the package won't work when actually installed.

Ok, looking more closely the issue is the `package.json` if you take a look at the `files` key, you can see we are choosing to only publish files from `service` and `index.*`. I think a better fix would be to either update `index.*` to `*.js`, or we could always leverage an `.npmignore` to ignore the `generateAssetData` directory which is the intention of the strict `files` setting I imagine.

PR #830, opened 2022-06-16
feat: use FOX CoinGecko market data for FOXy

gomesalexandre:
### Description

This 

- uses the FOX market data (`findByAssetId` and `findAll`) from CoinGecko in `FoxyMarketService` instead of the Coincap market data we're currently getting. This is to be consistent with the FOX market data. Implemented by extending `CoinGeckoMarketService` and calling the base class `findByAssetId`
- uses the FOX historical data (`findPriceHistoryByAssetId`) from Coingecko in `FoxyMarketService`, to be consistent. All references of Coincap are removed from the module, and the ALL chart for FOXy is accidentally fixed by doing so
- improves/fixes the typings of the Coingecko market data response and moves them to the `coingecko-types.ts` module. Uses `bnOrZero().toString()` to stringify the properties, now that we have the right typings for them
- updates the test cases to match fetching from Coingecko, and fixes a test that was silently failing with the previous logic. 

### Reference

https://github.com/shapeshift/web/issues/2024#issuecomment-1157536924

> > I am not sure where the discrepancies in price and change on day come from, but I assume you have a better handle on that aspect.
> 
> This one is caused by our market data providers prioritization logic. We consistently use Coincap for FOXy (which is effectively FOX market data), whereas for other ERC20s we use these market-services (in that order): Coingecko, Coincap, Yearn, Osmosis, FOXy (currently Coincap)

### Issue

- Partially tackles https://github.com/shapeshift/web/issues/2024

### Risks

In theory, possible regressions in web since the properties of the market data response we're getting were wrongly typed as strings, but in effect, the risk is small since we effectively handle the values as if they were numbers, with the right safeguards in place.

Regression tested
- Price charts/Balance charts/market data for BTC/ETH/ERC20s/Cosmos assets
- Fox Page market data
- DeFi section fiat values
- Tx history fiat values
- Account balances fiat values

### Screenshots

image
image
image
image


packages/market-service/src/foxy/foxy.test.ts
```diff
@@ -40,7 +43,7 @@ describe('foxy market service', () => {
     })
 
     it('can handle rate limiting', async () => {
-      mockedAxios.get.mockResolvedValue({ status: 429 })
+      mockedAxios.get.mockRejectedValue({ status: 429 })
```
gomesalexandre, line 46: Note: Since the underlying `CoinGeckoMarketService` has its own `RATE_LIMIT_THRESHOLDS_PER_MINUTE.COINGECKO` (500) rate limiting, we might want to ditch this test altogether, but might as well keep it for safety

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -124,13 +114,13 @@ export class CoinGeckoMarketService implements MarketService {
       Also a lot of time when max_supply is null, total_supply is the maximum supply on coingecko
       We can reassess in the future the degree of precision we want on that field */
       return {
-        price: marketData?.current_price?.[currency],
-        marketCap: marketData?.market_cap?.[currency],
+        price: bnOrZero(marketData?.current_price?.[currency]).toString(),
+        marketCap: bnOrZero(marketData?.market_cap?.[currency]).toString(),
         changePercent24Hr: marketData?.price_change_percentage_24h,
```
gomesalexandre, line 119: Note: This one is left as the number it is coming as from the API, and would require both this line and `@shapeshiftoss/types` to be updated:
https://github.com/shapeshift/lib/blob/6274ccdd7e0b5c6190da68ff1e148a125bcf5cdb/packages/types/src/market.ts#L5

theoboldfrazier (approved):
Looks good!

0xdef1cafe (changes requested):
requesting changes as this will conflict with the avalanche market service and it would be nice to get that in first

gomesalexandre:
> requesting changes as this will conflict with the avalanche market service and it would be nice to get that in first

Mergefixed and reopened after https://github.com/shapeshift/lib/pull/833 has been merged

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -122,11 +112,11 @@ export class CoinGeckoMarketService implements MarketService {
       Also a lot of time when max_supply is null, total_supply is the maximum supply on coingecko
       We can reassess in the future the degree of precision we want on that field */
       return {
-        price: marketData.current_price?.[currency].toString(),
-        marketCap: marketData.market_cap?.[currency].toString(),
-        changePercent24Hr: marketData.price_change_percentage_24h,
-        volume: marketData.total_volume?.[currency].toString(),
-        supply: marketData.circulating_supply.toString(),
+        price: bnOrZero(marketData?.current_price?.[currency]).toString(),
```
kaladinlight, line 115: There is already a null check for `marketData` above so we don't need the optional

packages/market-service/src/foxy/foxy.ts
```diff
@@ -81,9 +83,9 @@ export class FoxyMarketService implements MarketService {
       const supply = foxyTotalSupply
 
       return {
-        price: marketData.priceUsd,
+        price: cgData.price,
         marketCap: '0', // TODO: add marketCap once able to get foxy marketCap data
```
kaladinlight, line 87: Do we not want to use FOX marketCap and total volume as well?

kaladinlight: or at least we can calc marketCap based on fox price

gomesalexandre: We don't, those are two different assets. The price and price history is something we can share given FOXy is effectively "pegged" to FOX, but the rest is asset-specific

kaladinlight: Yeah that works with me

packages/market-service/src/foxy/foxy.ts
```diff
@@ -52,8 +51,11 @@ export class FoxyMarketService implements MarketService {
         return null
       }
 
-      const { data } = await axios.get(`${this.baseUrl}/assets/${FOX_COINCAP_ID}`)
-      const marketData = data.data as CoinCapMarketCap
+      const cgData = await super.findByAssetId({
```
0xApotheosis, line 54:
```suggestion
      const coinGeckoData = await super.findByAssetId({
```

packages/market-service/src/foxy/foxyMockData.ts
```diff
@@ -1,18 +1,12 @@
...
-  marketCapUsd: '76043211.3383411704757409',
-  volumeUsd24Hr: '809217.5632748945831009',
-  priceUsd: '0.2794809217688426',
-  changePercent24Hr: '2.8107676052084740',
-  vwap24Hr: '0.2717415151864480',
-  explorer: 'https://etherscan.io/token/0xc770eefad204b5180df6a14ee197d99d808ee52d'
+export const fox: CoinGeckoMarketData = {
+  circulating_supply: 272087306.915483,
+  max_supply: 1000001337.0,
+  // eslint-disable-next-line @typescript-eslint/no-loss-of-precision
```
0xApotheosis, line 6: We should aim to always explain why we disable lint by adding a comment in line.

I assume this is because it's test data and its runtime value doesn't matter.

0xApotheosis (approved):
Looks good, some non-blocking comments.

packages/market-service/src/foxy/foxy.ts
gomesalexandre: Renamed in bb6126bafa3f6fcb6e0ca3de88b08930d0ead244

0xdef1cafe (approved):

PR #833, opened 2022-06-16
feat: avalanche market data

kaladinlight:


packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -37,69 +36,68 @@ type CoinGeckoAssetData = {
...
+type CoinGeckoHistoryData = {
+  prices: [number, number][]
+}
+
+type CoinGeckoMarketServiceArgs = {
+  coinGeckoAPIKey: string
+}
+
 export class CoinGeckoMarketService implements MarketService {
+  private readonly maxCount = 2500
```
0xdef1cafe, line 48:
```suggestion
  private readonly defaultCount = 2500
```

this isn't/wasn't a max afaik

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -37,69 +36,68 @@ type CoinGeckoAssetData = {
...
-          }
-        }, {} as MarketCapResult)
+      const marketData = await Promise.all(
+        pageCount.map(async (page) => {
+          const { data } = await axios.get<CoinGeckoMarketCap>(urlAtPage(page))
+          return data ?? []
+        })
+      )
+
+      return marketData.flat().reduce<MarketCapResult>((prev, asset) => {
```
0xdef1cafe, line 86: we lost the explicit sort, but i guess if we're calling the `market_cap_desc` endpoint we can safely assume they're sorted.

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -144,64 +142,54 @@ export class CoinGeckoMarketService implements MarketService {
...
-        case HistoryTimeframe.YEAR:
-          start = end.subtract(1, 'year')
-          break
-        case HistoryTimeframe.ALL:
-          start = end.subtract(20, 'years')
-          break
-        default:
-          start = end
-      }
+      const start = (() => {
```
0xdef1cafe, line 151: 💜

packages/caip/src/adapters/coingecko/utils.ts
```diff
@@ -1,13 +1,17 @@
+/* eslint-disable no-empty */
```
0xdef1cafe, line 1: can we remove this?

0xdef1cafe (changes requested):
2 requested changes (eslint disable) and the `maxCount` thing should be `defaultCount`, rest are comments

packages/market-service/src/coingecko/coingecko.ts
kaladinlight: yeah that was the intention as it was already sorted and there was a ton of extra chaining before that was doing essentially nothing

packages/caip/src/adapters/coingecko/utils.ts
kaladinlight: yeah adding comment in the catch to explain and reduce need for the disable

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -108,30 +106,30 @@ export class CoinGeckoMarketService implements MarketService {
...
-        price: marketData?.current_price?.[currency],
-        marketCap: marketData?.market_cap?.[currency],
-        changePercent24Hr: marketData?.price_change_percentage_24h,
-        volume: marketData?.total_volume?.[currency],
-        supply: marketData?.circulating_supply,
-        maxSupply: marketData?.max_supply
-          ? marketData?.max_supply
-          : marketData?.total_supply?.toString()
+        price: marketData.current_price[currency],
+        marketCap: marketData.market_cap[currency],
```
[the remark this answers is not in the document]
kaladinlight: I will add back just to be safe, according to our type and what I've seen, we should be good if marketData exists, but doesn't hurt

packages/market-service/src/coingecko/coingecko.ts
```diff
@@ -37,101 +41,93 @@ type CoinGeckoAssetData = {
...
-      if (!adapters.assetIdToCoingecko(assetId)) return null
-      const { chainNamespace, assetReference } = fromAssetId(assetId)
-      const isToken = chainNamespace === CHAIN_NAMESPACE.Ethereum && assetReference.startsWith('0x')
-      const id = isToken ? 'ethereum' : adapters.assetIdToCoingecko(assetId)
-      const contractUrl = isToken ? `/contract/${assetReference}` : ''
-
-      const { data }: { data: CoinGeckoAssetData } = await axios.get(
-        `${this.baseUrl}/coins/${id}${contractUrl}?${this.maybeAddAPIKey()}`
-      )
+      const { data }: { data: CoinGeckoAssetData } = await axios.get(url)
```
0xdef1cafe, line 113:
```suggestion
      const { data } = await axios.get<CoinGeckoAssetData>(url)
```

0xdef1cafe (dismissed):

0xdef1cafe (approved):

gomesalexandre (approved):
Linked lib packages locally and confirmed this is working on all timeframes for Avalanche market data.
The data is matching the one on Coingecko, and the "Market Data" component is also populated with values that are all matching the ones from Coingecko.
🍬

Issue #726, opened 2022-06-01
CowSwap : Implements executeTrade / getTradeTxs methods in CowSwapper


chxash:
### Overview

We want to implement those methods in `CowSwapper` : 
- `executeTrade` with endpoint : https://api.cow.fi/mainnet/api/v1/orders
- `getTradeTxs` similarly to what was done for `ZrxSwapper`

With the way cowswap handles orders, we need to perform a message signing operation and not a tx signing one.
This is dependent on https://github.com/shapeshift/lib/issues/722

### References and additional details

See https://github.com/shapeshift/lib/issues/702 for more context

### Acceptance Criteria

- Both methods are implemented
- Both methods are unit tested

### Estimated effort

1 day

PR #837, opened 2022-06-21
feat: chainadapters signMessage method

chxash:
This implements the signMessage method in `EthereumChainAdapter`. This is needed in order to be able to sign messages for `CowSwap` (will be used in https://github.com/shapeshift/lib/issues/726)

`ChainAdapterCLI` can be ran in order to verify signMessage is working.

### DIFF
--- a/packages/chain-adapters/src/ChainAdapterCLI.ts
+++ b/packages/chain-adapters/src/ChainAdapterCLI.ts
@@ -210,7 +210,7 @@ const testEthereum = async (wallet: NativeHDWallet, broadcast = false) => {
         addressNList: [2147483692, 2147483708, 2147483648, 0, 0]
       }
     })
-    console.log('ethereum: signMessage', signedMessage)
+    console.log('ethereum: signedMessage', signedMessage)
   } catch (err) {
     console.log('ethereum: signMessage error:', err.message)
   }
--- a/packages/chain-adapters/src/ChainAdapterCLI.ts
+++ b/packages/chain-adapters/src/ChainAdapterCLI.ts
@@ -201,6 +201,19 @@ const testEthereum = async (wallet: NativeHDWallet, broadcast = false) => {
   } catch (err) {
     console.log('ethereum: tx error:', err.message)
   }
+
+  try {
+    const signedMessage = await chainAdapter.signMessage({
+      wallet,
+      messageToSign: {
+        message: 'Hello world 222',
+        addressNList: [2147483692, 2147483708, 2147483648, 0, 0]
+      }
+    })
+    console.log('ethereum: signMessage', signedMessage)
+  } catch (err) {
+    console.log('ethereum: signMessage error:', err.message)
+  }
 }
 
 // @ts-ignore:nextLine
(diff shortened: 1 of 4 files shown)


packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
```diff
@@ -420,6 +420,47 @@ describe('EthereumChainAdapter', () => {
...
+      const adapter = new ethereum.ChainAdapter(args)
+      const wallet = await getWallet()
+      wallet.ethSignMessage = async () => null
+      const message = {
+        wallet,
+        messageToSign: {
+          message: 'Hello world 111',
+          addressNList: [2147483692, 2147483708, 2147483648, 0, 0]
+        }
+      } as unknown as SignMessageInput<ETHSignMessage>
```
gomesalexandre, line 458: The `as unknown` is not needed here 
```suggestion
      } as SignMessageInput<ETHSignMessage>
```

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -201,6 +201,19 @@ const testEthereum = async (wallet: NativeHDWallet, broadcast = false) => {
...
+
+  try {
+    const signedMessage = await chainAdapter.signMessage({
+      wallet,
+      messageToSign: {
+        message: 'Hello world 222',
+        addressNList: [2147483692, 2147483708, 2147483648, 0, 0]
+      }
+    })
+    console.log('ethereum: signMessage', signedMessage)
```
kaladinlight, line 213:
```suggestion
    console.log('ethereum: signedMessage:', signedMessage)
```

packages/chain-adapters/src/ethereum/EthereumChainAdapter.test.ts
```diff
@@ -420,6 +420,47 @@ describe('EthereumChainAdapter', () => {
     })
   })
 
+  describe('signMessage', () => {
+    it('should sign a properly formatted signMessageInput object', async () => {
+      const args = makeChainAdapterArgs()
+      const adapter = new ethereum.ChainAdapter(args)
+      const wallet = await getWallet()
+      wallet.ethSignMessage = async () => ({
```
kaladinlight, line 428: we should be able to call the native wallet directly without overwriting this function since it doesn't require any interaction with the blockchain

gomesalexandre (dismissed):
Tested with `yarn adapterCli` and LGTM.
This is just exposing the `ethSignMessage` method in `chain-adapters` so as far as this goes, works as it should.

I do believe however there should also be a follow-up PR in hdwallet to update the `ethSignMessage` hdwallet implementations, which are not using the latest JSON-RPC message-signing methods, see:
https://docs.metamask.io/guide/signing-data.html#a-brief-history

To match the latest EIP-712 message signing spec, `signTypedData_v4` should be used in hdwallet, which is the EIP-712 compliant way to sign messages (cc @pastaghost)

0xdef1cafe (changes requested):

packages/chain-adapters/src/ChainAdapterCLI.ts
chxash: fixed !

kaladinlight (dismissed):

0xdef1cafe (approved):

0xdef1cafe:
merging as previous approved by big kev

PR #841, opened 2022-06-27
chore(asset-service): bump caip to latest

gomesalexandre:
This bumps caip to latest in asset-service.
Currently, when trying to call asset-service's `description`, this error is thrown:
```typescript
"TypeError: caip_1.adapters.makeCoingeckoAssetUrl is not a function
    at AssetService.description (http://localhost:3000/static/js/bundle.js:389235:35)
    at Object.queryFn (http://localhost:3000/static/js/bundle.js:87345:29)"
```
 This also fixes the `Element implicitly has an 'any' type because expression of type 'Nominal<string, "AssetId">' can't be used to index type '{}'.` type error that happens when trying to bump it, by not using `AssetId` to index the colorMap.

Unfortunately, there is no other way than string indexing, as any other type than primitive indexing isn't valid at runtime, thus making it a non-valid indexer that will always result in a type error.


theoboldfrazier (approved):

packages/asset-service/package.json
```diff
@@ -30,6 +30,7 @@
     "type-check": "tsc --project ./tsconfig.build.json --noEmit"
   },
   "dependencies": {
+    "@shapeshiftoss/caip": "^6.2.0",
```
kaladinlight, line 33: web should be controlling dependencies at this time with our usage of peer dependencies
```suggestion
```

kaladinlight (approved):

0xApotheosis (approved):

PR #845, opened 2022-06-29
feat: abstract evm base chain adapter and add avalanche chain adapter

kaladinlight:


packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.test.ts
```diff
@@ -0,0 +1,706 @@
...
+            low: 41000000000
+          }
+        ]
+      }
+    })
+  )
+}))
+
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('AvalancheChainAdapter', () => {
```
kaladinlight, line 62: unchained client does not currently support avalanche chain id, skipping tests until updated with support

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+      return { txToSign }
+    } catch (err) {
+      return ErrorHandler(err)
+    }
+  }
+
+  async getGasFeeData(): Promise<GasFeeDataEstimate> {
+    const feeData = (await this.providers.http.getGasFees()).data
+
+    const normalizationConstants = {
```
kaladinlight, line 122: magic +/- 20% fee diff for slow and fast, we can iterate to dial in more accurate fees. this differs from eth where we were using 0x gas api for fees

packages/types/src/base.ts
```diff
@@ -12,7 +12,8 @@ export enum KnownChainIds {
   EthereumMainnet = 'eip155:1',
   BitcoinMainnet = 'bip122:000000000019d6689c085ae165831e93',
   CosmosMainnet = 'cosmos:cosmoshub-4',
-  OsmosisMainnet = 'cosmos:osmosis-1'
+  OsmosisMainnet = 'cosmos:osmosis-1',
+  AvalancheMainnet = 'eip155:43114'
```
0xdef1cafe, line 16: this will prob need an accompanying web pr

packages/chain-adapters/src/utils/index.ts
```diff
@@ -35,27 +33,39 @@ export const getType = (type: TransferType): TxType => {
...
+          )
+      }
+    case CHAIN_NAMESPACE.Cosmos:
+      switch (chainReference) {
+        case CHAIN_REFERENCE.CosmosHubMainnet:
+        case CHAIN_REFERENCE.CosmosHubVega:
+          return 'cosmos'
+        case CHAIN_REFERENCE.OsmosisMainnet:
+        case CHAIN_REFERENCE.OsmosisTestnet:
+          return 'avalanche'
```
0xdef1cafe, line 62:
```suggestion
          return 'osmosis'
```

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+  public static readonly defaultBIP44Params: BIP44Params = {
+    purpose: 44,
+    coinType: Number(ASSET_REFERENCE.AvalancheC),
+    accountNumber: 0
+  }
+
+  constructor(args: ChainAdapterArgs) {
+    super({ chainId: DEFAULT_CHAIN_ID, supportedChainIds: SUPPORTED_CHAIN_IDS, ...args })
+
+    this.assetId = toAssetId({
```
0xdef1cafe, line 36: use hardcoded chain id over creating?

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  async validateAddress(address: string): Promise<ValidAddressResult> {
+    const chainLabel = chainIdToChainLabel(this.chainId)
+    const isValidAddress = WAValidator.validate(address, chainLabel)
+    if (isValidAddress) return { valid: true, result: ValidAddressResultType.Valid }
+    return { valid: false, result: ValidAddressResultType.Invalid }
+  }
+
+  async validateEnsAddress(address: string): Promise<ValidAddressResult> {
```
0xdef1cafe, line 225: does this belong in the base class? looks like avalanche has ANS as opposed to ENS, this should prob just exist in the ethereum impl

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
```diff
@@ -0,0 +1,285 @@
...
+
+export abstract class EVMBaseAdapter<T extends EVMChainIds> implements IChainAdapter<T> {
+  protected readonly chainId: ChainId
+  protected readonly supportedChainIds: Array<ChainId>
+  protected readonly providers: {
+    http: unchained.ethereum.V1Api
+    ws: unchained.ws.Client<unchained.ethereum.EthereumTx>
+  }
+
+  protected assetId: AssetId
```
0xdef1cafe, line 61:
```suggestion
  protected feeAssetId: AssetId
```

i don't really mind either wya here but this seems more explicit

0xdef1cafe (commented):
initial pass

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
kaladinlight: this would allow support for testnet ultimately since we are using `this.chainId` to create the `assetId`

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
kaladinlight: nope good catch!

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
kaladinlight: I guess we have hardcoded mainnet everywhere else though, so I suppose we could for now if we wanted to

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
kaladinlight: I am going to leave as is since it matches the rest of the chain adapters and we can always run back through and update them all at once to match if we like `feeAssetId` better

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -7,7 +7,7 @@ import dotenv from 'dotenv'
 import { ChainAdapter as BitcoinChainAdapter } from './bitcoin/BitcoinChainAdapter'
 import { ChainAdapter as CosmosChainAdapter } from './cosmossdk/cosmos'
 import { ChainAdapter as OsmosisChainAdapter } from './cosmossdk/osmosis'
-import { ChainAdapter as EthereumChainAdapter } from './ethereum/EthereumChainAdapter'
+import { ChainAdapter as EthereumChainAdapter } from './evm/ethereum'
```
gomesalexandre, line 10: ❤️

gomesalexandre (commented):
That looks good to me, added smallish comments WRT Avalanche/Ethereum chain adapter specifics

Before I stamp this, I'm wondering, what about the ERC-20 references across the Avalanche domain? Do we want to keep on calling these ERC-20s for Avalanche, or do we want to use the ARC-20 terminology?
There seems to be a bit of discrepancy in what (bridged) tokens are referred to in the C-Chain, either ERC-20s or ARC-20s but from my understanding, .e AB Bridge tokens are technically ARC-20s, which are a superset of ERC-20s (they are just a specific kind of ERC-20s that wrap a native Avalanche token), see https://docs.penguinfinance.io/getting-started/ab-bridge-guide-ethereum-to-avalanche

I would say ERC-20 is fine as this is usually how such tokens are referred to in Avalanche docs, but leaving this here for discussion!

kaladinlight:
> That looks good to me, added smallish comments WRT Avalanche/Ethereum chain adapter specifics
> 
> Before I stamp this, I'm wondering, what about the ERC-20 references across the Avalanche domain? Do we want to keep on calling these ERC-20s for Avalanche, or do we want to use the ARC-20 terminology? There seems to be a bit of discrepancy in what (bridged) tokens are referred to in the C-Chain, either ERC-20s or ARC-20s but from my understanding, .e AB Bridge tokens are technically ARC-20s, which are a superset of ERC-20s (they are just a specific kind of ERC-20s that wrap a native Avalanche token), see https://docs.penguinfinance.io/getting-started/ab-bridge-guide-ethereum-to-avalanche
> 
> I would say ERC-20 is fine as this is usually how such tokens are referred to in Avalanche docs, but leaving this here for discussion!

I think there is a slight distinction with the ARC-20s in which they are wrapped assets such as WAVAX (wrapped avax), whereas the standard ERC-20s still comply with the standard abi and don't have a `deposit()` or `withdraw()` function added. As it stands I think keeping with the `erc20` namespace vernacular as of now. I am not sure if there is a good reason to add `arc20` or `bridged` namespaces at this time for our purposes, though there are slight distinctions between those.

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
```diff
@@ -0,0 +1,277 @@
...
+      return signedTx.serialized
+    } catch (err) {
+      return ErrorHandler(err)
+    }
+  }
+
+  async signAndBroadcastTransaction(signTxInput: SignTxInput<ETHSignTx>): Promise<string> {
+    try {
+      const { txToSign, wallet } = signTxInput
+      const ethHash = await (wallet as ETHWallet)?.ethSendTx?.(txToSign)
```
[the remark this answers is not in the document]
gomesalexandre: Yeah, types are totally fine to keep as is!

gomesalexandre (approved):
Pulled locally and tested in web, did a regression testing in Ethereum and found no regressions following the `EthereumChainAdapter` changes
🍬

0xdef1cafe (approved):

0xApotheosis (approved):

Issue #725, opened 2022-06-01
CowSwap : Implements buildTrade in CowSwapper


chxash:
### Overview

We want to implement `buildTrade` method in `CowSwapper` by using endpoint https://api.cow.fi/mainnet/api/v1/quote
This is dependent on https://github.com/shapeshift/lib/issues/722

### References and additional details

https://github.com/shapeshift/lib/issues/702 for additional details

### Acceptance Criteria

- `buildTrade` is implemented
- This method is unit tested

### Estimated effort

1 day

PR #846, opened 2022-06-29
feat: cowswap buildTrade

chxash:
This implements `buildTrade` method for `CowSwapper` (https://github.com/shapeshift/lib/issues/725)
It also fixes a couple of issues related to previous PRs after successful local integration with Web, notably in `getCowSwapTradeQuote.ts`.

Some screenshots of the local integration with web 

ftxtrade1

ftxtrade2

ftxtrade3

ftxtrade4


packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.ts
```diff
@@ -0,0 +1,139 @@
...
+        sellAmountBeforeFee: normalizedSellAmount
+      })
+
+    const {
+      data: { quote }
+    } = quoteResponse
+
+    const rate = bn(quote.buyAmount)
+      .div(quote.sellAmount)
+      .times(bn(10).exponentiatedBy(sellAsset.precision - buyAsset.precision))
```
0xdef1cafe, line 77: can you explain what is going on here?

packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.ts
```diff
@@ -0,0 +1,139 @@
...
+        to: COW_SWAP_VAULT_RELAYER_ADDRESS,
+        value: '0',
+        chainSpecific: { from: receiveAddress, contractAddress: sellAssetErc20Address }
+      }),
+      getUsdRate(deps, feeAsset),
+      getUsdRate(deps, sellAsset)
+    ])
+    const feeData = feeDataOptions['fast']
+
+    const fee = bnOrZero(quote.feeAmount)
```
0xdef1cafe, line 91: similarly here, can we break up into a few lines? this logic is pretty terse.

0xdef1cafe (commented):
few questions, overall good

packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.ts
chxash: I split it into multiple lines for better comprehension.
basically I am computing the rate buyAssetAmount / sellAssetAmount

packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.ts
chxash: @0xdef1cafe agreed should have done that earlier, split it into more comprehensive code and added comments !

chxash:
@0xdef1cafe fixed comments + added one small fix into `helpers.ts/getUsdRate()` method relative to `USDC` rate

packages/swapper/src/swappers/cow/CowSwapper.test.ts
```diff
@@ -142,6 +147,16 @@ describe('CowSwapper', () => {
     })
   })
 
+  describe('buildTrade', () => {
+    it('calls cowBuildTrade on swapper.buildTrade', async () => {
```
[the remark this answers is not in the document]
kaladinlight: I see the test cases now in the separate test file, my bad...

packages/swapper/src/swappers/cow/utils/constants.ts
```diff
@@ -1,12 +1,11 @@
-export declare const MAX_ALLOWANCE = '100000000000000000000000000'
+import { AddressZero, HashZero } from '@ethersproject/constants'
```
kaladinlight, line 1: :+1:

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -57,3 +73,13 @@ export const getUsdRate = async (
     })
   }
 }
+
+export const getNowPlusThirtyMinutesTimestamp = (): number => {
```
kaladinlight, line 77: alternative
```
const ts = new Date()
ts.setMinutes(ts.getMinutes() + 30)
return Math.round(ts.getTime() / 1000)
```

packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
```diff
@@ -126,18 +117,18 @@ export async function getCowSwapTradeQuote(
       feeData: {
         fee,
         chainSpecific: {
-          estimatedGas: feeData.chainSpecific.gasLimit,
+          estimatedGas: APPROVAL_GAS_LIMIT,
```
kaladinlight, line 120: It seems like we would want to use the `feeData` gasLimit still here, so we are returning gas info for either making the trade, or performing the approval as already described by `approvalFee`.

It does make me think that our call to `getFeeData` is incorrect as it doesn't include the `appData` from the quote response to calculate the accurate gasLimit given the `inputData` required to make the trade (assuming that is what `appData` is when returned)

chxash: @kaladinlight some explanation on this : I did not include `appData` in the call to `getFeeData` because I would always get an error response while doing so. As a consequence, the `gasLimit` received from our `getFeeData` is probably not correct (insufficient to perform an approval from my blind testing), hence why I hardcoded this `APPROVAL_GAS_LIMIT`.
Not sure If I should spend time trying to bring back `appData` into `getFeeData` call.

packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.ts
```diff
@@ -0,0 +1,148 @@
...
+    const {
+      data: { quote }
+    } = quoteResponse
+
+    const buyCryptoAmount = bn(quote.buyAmount).div(bn(10).exponentiatedBy(buyAsset.precision))
+    const sellCryptoAmount = bn(quote.sellAmount).div(bn(10).exponentiatedBy(sellAsset.precision))
+    const rate = buyCryptoAmount.div(sellCryptoAmount).toString()
+
+    const [feeDataOptions, feeAssetUsdRate, sellAssetUsdRate] = await Promise.all([
+      adapter.getFeeData({
```
kaladinlight, line 80: same comment here regarding contract `inputData`. also not for this pr probably, but we should be using eip1559 fees if  supported for these trades as well.

packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
kaladinlight: At minimum, please add a comment if as to why the hard code that happens to be enough to cover the trade. I would ultimately like to understand the error case with the `appData` being passed to `getFeeData` so we can properly calculate these fees. Do you have tx reference I can look at or the returned `appData` on hand you can send my way and I can take a look?

packages/swapper/src/swappers/cow/utils/helpers/helpers.ts
```diff
@@ -75,11 +75,7 @@ export const getUsdRate = async (
...
 export const getNowPlusThirtyMinutesTimestamp = (): number => {
-  const millisecondsPerMinute = 60000
-
-  // UTC date + 30 minutes
-  const nowPlusThirtyMinutesDate = new Date(new Date().getTime() + 30 * millisecondsPerMinute)
-
-  // returning the timestamp in seconds
-  return Math.round(nowPlusThirtyMinutesDate.getTime() / 1000)
+  const ts = new Date()
+  ts.setMinutes(ts.getMinutes() + 30)
```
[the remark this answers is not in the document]
chxash: > this will still break twice a day every four years but i'm ok with that 😛

Found those DST issues as well but it seems very old ? maybe it is not an issue anymore ?

packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
kaladinlight: @chxash this looks good for accurately estimating gas for approvals. i did a little more digging and the missing piece of info for myself is that it doesn't look like there are actually any network fees for the swapping through their protocol due to their fees being calculated into the actual trade. can you confirm this?

In that case, it seems like we don't really even need the anything but the `feeData.fee` and `feeData.chainSpecific.approvalFee`. Regardless, I think it is good for us to accurately estimate fees where we can so having the approval calculation is a good improvement.

packages/swapper/src/swappers/utils/helpers/helpers.ts
```diff
@@ -168,3 +176,18 @@ export const normalizeIntegerAmount = (amount: string | number | BN): string =>
...
+
+export const getApproveContractData = async ({
+  web3,
+  sellAssetErc20Address,
+  receiveAddress
+}: GetApproveContractDataArgs): Promise<string> => {
+  const approveContract = new web3.eth.Contract(erc20AbiImported, sellAssetErc20Address)
+
+  const preApprove = await approveContract.methods.approve(
+    COW_SWAP_VAULT_RELAYER_ADDRESS,
```
kaladinlight, line 188: The spender address should be passed in seeing as this is a generic helper

packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
chxash: > @chxash this looks good for accurately estimating gas for approvals. i did a little more digging and the missing piece of info for myself is that it doesn't look like there are actually any network fees for the swapping through their protocol due to their fees being calculated into the actual trade. can you confirm this?

Exactly yes, sorry for not mentionning it earlier

packages/swapper/src/swappers/utils/helpers/helpers.ts
chxash: Fixed ! thanks

kaladinlight (approved):

0xdef1cafe (approved):

PR #847, opened 2022-06-29
chore: bump lib dependencies

kaladinlight:
fixes type errors during local linking now that all packages are pointing at the same versions


0xApotheosis (dismissed):
Nice!

I'm still getting the `web3` issue, but all of the type issues I was getting when linking appear to be fixed.

Screen Shot 2022-06-30 at 8 13 28 am

0xApotheosis:
Type narrowing on `chainSpecfic` also seems to be functioning optimally 👌

0xdef1cafe (dismissed):

0xApotheosis (approved):
`web3` issues fixed when used in combination with https://github.com/shapeshift/web/pull/2093

PR #848, opened 2022-06-30
fix(swapper): use best practice naming conventions

0xApotheosis:
Some non-standard naming conventions for file names and functions have been used, and are starting to propagate.

This PR ensures:

1. [lowerCamelCasing is used for functions](https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Code_guidelines/JavaScript#function_naming)
2. lowerCamelCasing is used for files exporting functions, or if the file exports a single class, the filename is exactly the name of the class


gomesalexandre (approved):
LGTM, linked to web as a sanity check and didn't find any regression. CI is happy, so here's a: 🍬

kaladinlight (approved):

Issue #836, opened 2022-06-20
Asset descriptions should be matching user locale


gomesalexandre:
@shapeshift/operations Reserved for @sebfk

### Overview

Currently in lib, we are using the asset descriptions from Coingecko API and getting only the `en` description from it.
Similarly, our overriden descriptions are only in English.

We should make it so that descriptions are dynamic according to the user language.

Since we are calling `asset-service`'s `description`, lib would be the best place to handle it and we could make this method accept a `locale` parameter to pick the right translation from Coingecko, or fallback to English if it isn't available for that language (empty string). See the [Ethereum Coingecko API call](https://api.coingecko.com/api/v3/coins/ethereum).

For the overriden description, there are two ways to implement this:

- Keep the naive `descriptions.json` implementation and index translations by locale: `{[assetId: AssetId]: {[locale: string]: string}` vs. `{[assetId: AssetId]: string}` currently
- Make the overriden descriptions dynamic using a similar i18n logic as we have in web, and split them by language. This way, we could call `polyglot.t()` to get the description for the specific language, or fallback to English [as we do in web currently](https://github.com/shapeshift/web/blob/a3854c60dc3a6e68e3b47778923bd7e16a785971/src/context/I18nProvider/I18nProvider.tsx#L11-L14)

A web PR should also be made to bump the `@shapeshiftoss/asset-service` dependency and consume the updated `description(assetId, locale)` signature.

### References and additional details

[asset-service's `description`](https://github.com/shapeshift/lib/blob/94f74495b8d3b082517c10c60288f8fa1e411daf/packages/asset-service/src/service/AssetService.ts#L57)
[`descriptions.json` for the overriden descriptions](https://github.com/shapeshift/lib/blob/94f74495b8d3b082517c10c60288f8fa1e411daf/packages/asset-service/src/service/descriptions.json)
[`Polyglot.js`](https://github.com/airbnb/polyglot.js/) (node-polyglot in npm)
[web translations logic](https://github.com/shapeshift/web/blob/develop/src/assets/translations/index.ts)

### Acceptance Criteria

- The Coingecko description matching the user locale is returned from asset-service's `description`, or English as a fallback if unavailable
- Overriden descriptions are implemented with internationalization and the overriden description matching the user local is returned, or English as a fallback

### Estimated effort

3 hours

Lychbot:
[...]

PR #849, opened 2022-06-30
feat(asset-service): localized asset descriptions

gomesalexandre:
### Description

This adds localization support for Coingecko/overriden asset descriptions

- splits the `src/service/descriptions.json` file into many ` `src/service/descriptions/<locale>.json` files
- supports an optional `locale` parameter to pick the localized description from Coingecko/overriden descriptions
- adds an instantiated Polyglot.js calling `polyglot.t()` on the overriden descriptions, and falling back to the en JSON if the description translation hasn't been added yet for the passed locale
- updates test to match the new logic

### Issue

- Partially tackles https://github.com/shapeshift/lib/issues/836
- Partially tackles https://github.com/shapeshift/web/issues/2052


packages/asset-service/src/service/AssetService.ts
```diff
@@ -21,22 +22,33 @@ export class AssetService {
...
+    const localeDescriptions = assetsDescriptions[locale]
     // Return overridden asset description if it exists and add isTrusted for description links
-    if (description) return { description, isTrusted: true }
+    if (assetsDescriptions.en[assetId]) {
+      const polyglot = new Polyglot({
+        phrases: localeDescriptions,
+        allowMissing: true,
+        onMissingKey: (key) => assetsDescriptions.en[key] // fallback to English overriden description, which should always be added as a base translation
+      })
+      const overriddenDescription = polyglot.t(assetId) as string
```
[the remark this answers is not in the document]
gomesalexandre: Nice one, thanks 🙏🏽 initially saw an `any` for it but reloading my language server after installing types gave me the proper typings for `polyglot.t()`!

packages/asset-service/src/service/AssetService.ts
```diff
@@ -21,22 +22,33 @@ export class AssetService {
...
 
       const url = adapters.makeCoingeckoAssetUrl(assetId)
       if (!url) throw new Error()
 
       const { data } = await axios.get<CoinData>(url)
 
-      return { description: data?.description?.en ?? '' }
+      const description = data?.description[locale]?.length
+        ? data?.description?.[locale]
+        : data?.description.en ?? ''
```
kaladinlight, line 49: alternative
`data?.description?.[locale] || data?.description?.en ?? ''`

packages/asset-service/src/service/descriptions/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import ko from './ko.json'
+import pt from './pt.json'
+import ru from './ru.json'
+import zh from './zh.json'
+
+const descriptions = { en, es, fr, id, ko, pt, ru, zh } as {
+  [locale: string]: Record<string, string>
+}
+
+export default descriptions
```
kaladinlight, line 14: not sure the need for the typing here
```suggestion
export default { en, es, fr, id, ko, pt, ru, zh }
```

kaladinlight:
Are we waiting for override description translations for these other languages I assume?

packages/asset-service/src/service/AssetService.ts
gomesalexandre: Not sure about this one, it was originally on the attempted bounty PR https://github.com/shapeshift/lib/pull/843#discussion_r908809250 but I always prefer to use the nullish coalescing operator for consistency and keep short circuits for explicitly truthy/falsy values

kaladinlight: we can also do a check on `if (!data?.description) return { description: ''}` to save reduce the optional chains if that would make it more readable

packages/asset-service/src/service/descriptions/index.ts
gomesalexandre: This will type the JSON (which are always typed as any) for consumers:

image

We currently do so by creating a new typed variable https://github.com/shapeshift/lib/pull/849/files#diff-7ac6886bda84d5c7853bc86947fb6253944e449aaef53bf61b96a656690b236bL25 but we can now do the same directly at export-level

packages/asset-service/src/service/AssetService.ts
kaladinlight: first `||` will fall back if undefined or empty string in the case no locale found so we can be sure to check for en description in both undefined or empty string case

packages/asset-service/src/service/descriptions/index.ts
kaladinlight: I tested locally and it is inferring types correctly for me which is why I commented. hmm..

kaladinlight: localeDescriptions is any though I am noticing in the asset service since local is just a string and can't be guaranteed from the inferred type

packages/asset-service/src/service/AssetService.ts
gomesalexandre: That sounds cleaner! Added in b54505b98e9ddfa2e55de5817fb6b3f464463459 and removed not-needed optional chaining. It looks like `data?.description[locale]` should always be defined (but possibly empty) so we can safely remove one more optional chaining

packages/asset-service/src/service/descriptions/index.ts
kaladinlight: It's fine, we can leave as is makes sense to avoid ambiguity

gomesalexandre: Yeah, if that was a JS file we could `as const` it and better type `locale` to properly discriminate depending on the accessor, but as this is a JSON, this is the best typings we can get for now!

packages/asset-service/src/service/AssetService.ts
gomesalexandre: > first `||` will fall back if undefined or empty string in the case no locale found so we can be sure to check for en description in both undefined or empty string case

`Boolean('')` is truthy so using `||` would just move the check back somewhere else haha

kaladinlight: > > first `||` will fall back if undefined or empty string in the case no locale found so we can be sure to check for en description in both undefined or empty string case
> 
> `Boolean('')` is truthy so using `||` would just move the check back somewhere else haha

yeah to our fallback en.

`data.description[locale]` being either undefined or `''` would fall back to `data.descriptions.en` with the `||`

kaladinlight (approved):

packages/asset-service/src/service/AssetService.ts
```diff
@@ -21,22 +22,33 @@ export class AssetService {
...
-
+  async description(assetId: AssetId, locale = 'en'): Promise<DescriptionData> {
+    const localeDescriptions = assetsDescriptions[locale]
     // Return overridden asset description if it exists and add isTrusted for description links
-    if (description) return { description, isTrusted: true }
+    if (localeDescriptions[assetId] || assetsDescriptions.en[assetId]) {
+      const polyglot = new Polyglot({
+        phrases: localeDescriptions,
+        allowMissing: true,
+        onMissingKey: (key) => assetsDescriptions.en[key] // fallback to English overriden description, which should always be added as a base translation
```
0xdef1cafe, line 32: does this not negate the need for all of the non-english files with the same translations as english?

0xdef1cafe: as discussed they're effective placeholders that lower risk for when translations are available, LGTM

0xdef1cafe (approved):

packages/asset-service/src/service/AssetService.ts
gomesalexandre: We don't need them yet, but this is provided with all translations ready to be modified so that globalization can provide them after we merge it, making things smoother 🥥

PR #851, opened 2022-07-01
feat(chain-adapters): add getRpcUrl to EVMBaseAdapter

gomesalexandre:
## Description

This adds a `rpcUrl` class property in `EVMBaseAdapter`, and a `getRpcUrl` getter.

Note that this hasn't been added to the other adapters, as the current need for getting the rpcUrl is to add an ethereum chain in https://github.com/shapeshift/hdwallet/pull/547 - if we do end up architecturally needing it in other adapters, we could add it there, but IMO it shouldn't be added just for the sake of consistency.

### DIFF
--- a/packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
+++ b/packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
@@ -33,7 +33,6 @@ export class ChainAdapter extends EVMBaseAdapter<KnownChainIds.AvalancheMainnet>
     super({ chainId: DEFAULT_CHAIN_ID, supportedChainIds: SUPPORTED_CHAIN_IDS, ...args })
 
     this.assetId = avalancheAssetId
-    this.rpcUrl = args.rpcUrl
     this.parser = new unchained.ethereum.TransactionParser({
       chainId: this.chainId,
       rpcUrl: this.rpcUrl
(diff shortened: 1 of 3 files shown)


0xApotheosis (approved):
I agree with the argument of keeping it on the EVM base adapter unless otherwise needed at a higher layer of abstraction.

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
```diff
@@ -33,16 +33,21 @@ export class ChainAdapter extends EVMBaseAdapter<KnownChainIds.AvalancheMainnet>
...
-      rpcUrl: args.rpcUrl
+      rpcUrl: this.rpcUrl
     })
   }
 
   getType(): KnownChainIds.AvalancheMainnet {
     return KnownChainIds.AvalancheMainnet
   }
 
+  getRpcUrl(): string {
```
kaladinlight, line 47: Can just add to the `EVMBaseAdapter`

0xdef1cafe:
conceptually good - but agree with others to pop this on the EVMBaseAdapter

0xdef1cafe (approved):

0xApotheosis (approved):

packages/chain-adapters/src/evm/avalanche/AvalancheChainAdapter.ts
0xApotheosis: Addressed in https://github.com/shapeshift/lib/pull/851/commits/d85a1d5864b02754a403fc8601803466105deac7

PR #853, opened 2022-07-02
feat(swapper): remove deprecated getSwapper method

0xApotheosis:
`getSwapper` is deprecated, and is now only used in tests.

This PR removes it and uses `Map.prototype.get()` in the tests instead.

### DIFF
--- a/packages/swapper/src/manager/SwapperManager.test.ts
+++ b/packages/swapper/src/manager/SwapperManager.test.ts
@@ -14,6 +14,7 @@ describe('SwapperManager', () => {
     web3: <Web3>{},
     adapter: <ethereum.ChainAdapter>{}
   }
+
   const cowSwapperDeps: CowSwapperDeps = {
     apiUrl: 'https://api.cow.fi/mainnet/api/',
     adapter: <ethereum.ChainAdapter>{},
@@ -38,7 +39,7 @@ describe('SwapperManager', () => {
     it('should add swapper', () => {
       const manager = new SwapperManager()
       manager.addSwapper(SwapperType.Thorchain, new ThorchainSwapper(thorchainSwapperDeps))
-      expect(manager.getSwapper(SwapperType.Thorchain)).toBeInstanceOf(ThorchainSwapper)
+      expect(manager.swappers.get(SwapperType.Thorchain)).toBeInstanceOf(ThorchainSwapper)
     })
 
     it('should be chainable', async () => {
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
LGTM, all references to `manager.getSwapper()` are removed in tests and replaced by `Map.prototype.get()` 🍬

packages/swapper/src/manager/SwapperManager.test.ts
```diff
@@ -59,11 +60,11 @@ describe('SwapperManager', () => {
     })
   })
 
-  describe('getSwapper', () => {
+  describe('get', () => {
```
gomesalexandre, line 63: I'd probably call this one `can get swapper instance from swappers` or something descriptive like so

0xApotheosis: Right you are, that was a lazy test name. Updated in https://github.com/shapeshift/lib/pull/853/commits/0baf714b97e3ab6b870a46bf2e3913cba454e518.

gomesalexandre (approved):

packages/swapper/src/manager/SwapperManager.test.ts
```diff
@@ -73,24 +74,10 @@ describe('SwapperManager', () => {
...
-    })
-
-    it('should throw an error if an invalid Swapper instance is passed', () => {
-      const manager = new SwapperManager()
-      // @ts-ignore
-      expect(() => manager.addSwapper(SwapperType.Thorchain, {})).toThrow(
-        '[validateSwapper] - invalid swapper instance'
-      )
+      expect(swapper.swappers.get(SwapperType.Thorchain)).toBeInstanceOf(ThorchainSwapper)
+      expect(swapper.swappers.get(SwapperType.Zrx)).toBeInstanceOf(ZrxSwapper)
```
kaladinlight, line 78:
```suggestion
```

kaladinlight:
This feels a little weird to me having `get` be the only method interacting directly with the map versus add/remove/etc using the helper functions. To me it seems like we either want to just interact with the map directly (a-la chainAdapterManager), or if there is extra necessary logic as part of the swapper manager (ie. getBest) that we don't want to duplicate in the client, then we can just have the swapper manager class and interact with the provided methods imo.

0xdef1cafe:
as discussed - let's talk to the manager through it's interfaces it exposes, rather than the raw map of swappers

0xdef1cafe:
@0xApotheosis plz undeprecate this (lol) as to not cause future confusion

PR #854, opened 2022-07-05
fix: bump osmosis gas limit

gomesalexandre:
This bumps the gas limit for osmosis Txs to 300000/500000/700000

Relevant discussion: https://github.com/shapeshift/web/pull/2067#issuecomment-1175100530

> WRT failed Txs for unstake, I've noticed that 300000 seems to be the sweet gas spot that consistently broadcasts Txs.
We should adjust the gas for Osmosis Txs to be higher than what it is currently, as unstake Txs consistently fail. Note that the needed gas can very depending on current network conditions, so we should probably not rely on this number but rather use e.g 500000 as medium gas.
This should be updated in OsmosisChainAdapter's getFeeData.


kaladinlight:
I think we should just set them all to `300000` which is enough gas to cover all tx types that we are creating

0xdef1cafe (approved):

0xApotheosis (approved):
🐏

PR #862, opened 2022-07-07
chore: avax chain adapter tests

kaladinlight:


packages/chain-adapters/src/evm/EVMBaseAdapter.ts
```diff
@@ -39,8 +39,8 @@ export type EVMChainIds = KnownChainIds.EthereumMainnet | KnownChainIds.Avalanch
 export interface ChainAdapterArgs {
   chainId?: ChainId
   providers: {
-    http: unchained.ethereum.V1Api
-    ws: unchained.ws.Client<unchained.ethereum.EthereumTx>
+    http: unchained.ethereum.V1Api | unchained.avalanche.V1Api
```
0xApotheosis, line 42: Did we want to make an `EvmV1Api` type, or might we use different versions at once?

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
```diff
@@ -54,8 +54,8 @@ export abstract class EVMBaseAdapter<T extends EVMChainIds> implements IChainAda
   protected readonly chainId: ChainId
   protected readonly supportedChainIds: Array<ChainId>
   protected readonly providers: {
-    http: unchained.ethereum.V1Api
-    ws: unchained.ws.Client<unchained.ethereum.EthereumTx>
+    http: unchained.ethereum.V1Api | unchained.avalanche.V1Api
+    ws: unchained.ws.Client<unchained.ethereum.EthereumTx | unchained.avalanche.AvalancheTx>
```
0xApotheosis, line 58: Same here, `EvmTx` or something?

0xApotheosis (approved):

packages/chain-adapters/src/evm/EVMBaseAdapter.ts
kaladinlight: I had started with abstracting the api, but with the use of `tsoa` to generate the swagger.json and routes, this wasn't trivial and punted for now. There is no real reason I couldn't just make an interface for now other than needing to expose a different set of types than just the generated. I will make an issue for both this and exposing the base common api types.

kaladinlight: https://github.com/shapeshift/unchained/issues/442

0xApotheosis: Thank you ser!

PR #874, opened 2022-07-13
feat(swapper): use hardcoded gas limit in grantAllowance

0xApotheosis:
Closes https://github.com/shapeshift/web/issues/2092

---

When we call `buildSendTransaction` for approvals we are using the estimated gas from the 0x trade quote response, which is far too high, as _approvals_ are cheaper.

We have an existing hardcoded value, `APPROVAL_GAS_LIMIT`, which we show to users in the UI, but do not use when building an approval transaction.

This leads to an issue where the "Estimated Gas Fee" we display to the user is radically different to the fee provided to the wallet, e.g. Metamask.

This PR uses the `APPROVAL_GAS_LIMIT` in the _approval infinite_ `0x` and `thor` swapper logic, which propagates to `buildSendTransaction` so that the estimate in the wallet matches the estimate shown on the ShapeShift app, and (in general) reduces the gas limit.

Screen Shot 2022-07-13 at 10 38 45 am

Screen Shot 2022-07-13 at 10 38 53 am

Longer-term we should move away from a hardcoded value and hit an endpoint that gives us an accurate gas limit for approvals.

### DIFF
--- a/packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
+++ b/packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
@@ -1,5 +1,6 @@
 import { ApproveInfiniteInput, EvmSupportedChainIds, SwapError, SwapErrorTypes } from '../../../api'
 import { erc20Abi } from '../../utils/abi/erc20-abi'
+import { APPROVAL_GAS_LIMIT } from '../../utils/constants'
 import { grantAllowance } from '../../utils/helpers/helpers'
 import { ZrxSwapperDeps } from '../types'
 import { MAX_ALLOWANCE } from '../utils/constants'
@@ -9,11 +10,21 @@ export async function zrxApproveInfinite<T extends EvmSupportedChainIds>(
   { quote, wallet }: ApproveInfiniteInput<T>
 ) {
   try {
+    const approvalQuote = {
+      ...quote,
+      sellAmount: MAX_ALLOWANCE,
+      feeData: {
+        ...quote.feeData,
+        chainSpecific: {
+          ...quote.feeData.chainSpecific,
+          // 0x approvals are cheaper than trades, but we don't have dynamic quote data for them.
+          // Instead, we use a hardcoded gasLimit estimate in place of the estimatedGas in the 0x quote response.
+          estimatedGas: APPROVAL_GAS_LIMIT
+        }
+      }
+    }
     const allowanceGrantRequired = await grantAllowance<T>({
-      quote: {
-        ...quote,
-        sellAmount: MAX_ALLOWANCE
-      },
+      quote: approvalQuote,
       wallet,
       adapter,
       erc20Abi,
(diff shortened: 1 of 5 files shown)


0xApotheosis:
Keeping this in draft for now as I need to make this ERC20/0x specific.

0xApotheosis:
`grantAllowance` is currently used by `thor`, `zrx`, and `cow` swappers.

`thor` and `zrx` use the `APPROVAL_GAS_LIMIT` constant when rendering the approval fee estimate to the user.
`cow` does not, but from my testing it seems its approval fees follow the same heuristic of being significantly cheaper, and so "fee = gasLimit * gasPrice" is innaccurate - thus it should use the constant.

As such, I've updated `cow` swapper's `approvalFee` accordingly in https://github.com/shapeshift/lib/pull/874/commits/3a0747a5aebcfe47c32a6c2472c9b4f0146c22f7.

chxash:
> `grantAllowance` is currently used by `thor`, `zrx`, and `cow` swappers.
> 
> `thor` and `zrx` use the `APPROVAL_GAS_LIMIT` constant when rendering the approval fee estimate to the user. `cow` does not, but from my testing it seems its approval fees follow the same heuristic of being significantly cheaper, and so "fee = gasLimit * gasPrice" is innaccurate - thus it should use the constant.
> 
> As such, I've updated `cow` swapper's `approvalFee` accordingly in [3a0747a](https://github.com/shapeshift/lib/commit/3a0747a5aebcfe47c32a6c2472c9b4f0146c22f7).

@0xApotheosis for `cow` in `getCowSwapTradeQuote.ts`, we are getting fee Data specifically for an approval operation (see https://github.com/shapeshift/lib/blob/main/packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts#L94). The reason is we don't perform the swap as a transaction, so we don't need TX fees (the transaction is being handled by `cowswap` protocol which charges us a fee in the process when we call their api).
Hence I think current fee estimation is correct (I did not face any issue doing approvals like 10+ times)

0xApotheosis:
Thanks for the context @chxash! I've reverted the cow swapper changes.

I think there is a better way to do this that removes the hardcoded approval value in favor of dynamic across all swappers. I'm going to make those changes, then re-open.

0xApotheosis:
I rearchitected this in https://github.com/shapeshift/lib/pull/874/commits/f986c2b8715e77684984424790b923b7955837ac to keep `grantAllowance` unaware of any hardcoded values, leaving it generic to handle dynamic values such as those in cow swapper.

Instead, the hardcode is set in the `thor` and `0x` _approve infinite_ logic, leaving `cow` unaffected.

We can tackle server-side approval estimates for `thor` and `0x` another day.

@chxash FYI 👍

packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
```diff
@@ -9,11 +10,21 @@ export async function zrxApproveInfinite<T extends EvmSupportedChainIds>(
   { quote, wallet }: ApproveInfiniteInput<T>
 ) {
   try {
+    const approvalQuote = {
```
kaladinlight, line 13: I suppose we can move forward with this, but fee management in swapper is def a bit of a mess... Another thought is we now have `approvalFee` on the `feeData.chainSpecific` data that we can garner the approval `gasLimit` from instead of writing over quote payloads.

May just want to wait until doing dynamic approval fee estimates though

0xdef1cafe (approved):

elmutt (approved):

packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
0xdef1cafe: agreed this is an improvement but let's not tackle dynamic fees here

kaladinlight (approved):

PR #877, opened 2022-07-14
fix(chain-adapters): fix showOnDevice in EvmBaseAdapter.ts

cjthompson:
Fix `showOnDevice` in `EvmBaseAdapter`.

The code was changed in v7.3.0.

The previous code:
```ts
showDisplay: Boolean(input.showOnDevice)
```

The current code:
```ts
showDisplay: showOnDevice
```

The problem here lies in that `showOnDevice` is typed as `boolean | undefined`. Passing `undefined` to HDWallet results in the default of `true` for `showOnDevice` instead of `false`.  The `Boolean()` wrapper converts `undefined` to `false`, thus keeping the expected behavior.

### DIFF
--- a/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
+++ b/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
@@ -218,7 +218,7 @@ export abstract class EvmBaseAdapter<T extends EvmChainIds> implements IChainAda
     const addressNList = bip32ToAddressNList(path)
     const address = await (wallet as ETHWallet).ethGetAddress({
       addressNList,
-      showDisplay: showOnDevice
+      showDisplay: Boolean(showOnDevice)
     })
     return address as string
   }


0xApotheosis (approved):
Nice catch.

gomesalexandre (approved):

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -218,7 +218,7 @@ export abstract class EvmBaseAdapter<T extends EvmChainIds> implements IChainAda
     const addressNList = bip32ToAddressNList(path)
     const address = await (wallet as ETHWallet).ethGetAddress({
       addressNList,
-      showDisplay: showOnDevice
+      showDisplay: Boolean(showOnDevice)
```
kaladinlight, line 221: Thanks for the fix on this. Can I suggest we use a default value for `showOnDevice` when destructuring instead of the `Boolean()` cast to make this more clear and explicit?

cjthompson: I am okay with that.

kaladinlight (approved):

gomesalexandre (approved):
LGTM. Way clearer what's the intended behavior with tests, and by not casting `showDisplay` to `Boolean()` in-place when passing it down.

PR #880, opened 2022-07-14
feat: chain adapters getDisplayName

0xdef1cafe:
chain adapters should be the source of truth - this allows web to ask each chain adapter for a
display name to be used in the view layer, rather than maintaining a map of id -> name in the
consumer.


cjthompson (approved):
I thought no it would be fine as a readoy property rather than a function but it does follow the existing pattern.

PR #883, opened 2022-07-15
feat: switch EVM chains on buildSendTransaction

gomesalexandre:
## Description

Following https://github.com/shapeshift/web/pull/2165, we do disconnect nor switch to Mainnet on MM/XDEFI if the current chain isn't mainnet. Now that we have multiple EVM chains support, we actually do not need to be on any specific chain to show on-chain data, we only need to switch chains when calling `buildSendTransaction` if there is a mismatch between the adapter-expected EVM chainId and what the current's wallet EVM chainId is

### DIFF
--- a/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
+++ b/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
@@ -1,13 +1,15 @@
-import { AssetId, ChainId, toAssetId } from '@shapeshiftoss/caip'
+import { AssetId, ChainId, fromAssetId, fromChainId, toAssetId } from '@shapeshiftoss/caip'
 import {
   bip32ToAddressNList,
   ETHSignMessage,
   ETHSignTx,
-  ETHWallet
+  ETHWallet,
+  supportsEthSwitchChain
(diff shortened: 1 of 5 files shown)


packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -84,9 +88,80 @@ export abstract class EvmBaseAdapter<T extends EvmChainIds> implements IChainAda
...
+      const txToSign = {
+        addressNList: bip32ToAddressNList(toPath(bip44Params)),
+        value: numberToHex(isErc20Send ? '0' : tx.value),
+        to: destAddress,
+        chainId: Number(fromChainId(this.chainId).chainReference),
+        data,
+        nonce: numberToHex(account.chainSpecific.nonce),
+        gasLimit: numberToHex(gasLimit),
+        ...fees
+      } as ChainTxType<T>
```
0xdef1cafe, line 158: that'll do pig @gomesalexandre

kaladinlight (approved):
:100:

elmutt (approved):

PR #884, opened 2022-07-18
feat: override Ethereum asset translations

gomesalexandre:
This overrides the Ethereum (`eip155:1/slip44:60`) asset translations, to be able to support the external links. Currently, we do not allow external links on this asset's description for security purposes since it's coming from CoinGecko's API.

Translations are copied from `https://api.coingecko.com/api/v3/coins/ethereum` for `en` (where links are currently disabled in web), and also `zh` and `ko` for consistency.

Additionally, some typos have been fixed with English language (see e.g wrong capitalization of MetaMask and MyEtherWallet on the original copies from the aforementioned API request), and valid line breaks have been added to fit the original copy, with the HTML `<br/>` tag (since regular markdown line breaks require a newline, which is invalid JSON syntax).


0xApotheosis (dismissed):
Looks good.

Noting that whilst the external link to `https://www.coingecko.com/en?category_id=29&view=market` does work, it probably doesn't go where it's meant to - it just redirects me to the home page of CoinGecko. But, it is indeed the URL returned in the response from `https://api.coingecko.com/api/v3/coins/ethereum`...

gomesalexandre:
> Looks good.
> 
> Noting that whilst the external link to `https://www.coingecko.com/en?category_id=29&view=market` does work, it probably doesn't go where it's meant to - it just redirects me to the home page of CoinGecko. But, it is indeed the URL returned in the response from `https://api.coingecko.com/api/v3/coins/ethereum`...

Nice catch! This just fixes the links for now but I've noticed the content of the copy itself has typos (which I've fixed) and is quite outdated e.g mentioning Trezor but not Ledger. I already got in touch with @firebomb1 to potentially improve them with our own, but in the meantime, we could fix this specific case by linking to `https://www.coingecko.com/en/categories/smart-contract-platform` - updated in ed5661edd45f173a568fe0a5a33fd0b742afe500

0xdef1cafe (approved):

PR #890, opened 2022-07-19
Osmosis swapper

toshiSat:
revert to previous account numbers for ibc/swap transactions


kaladinlight (approved):
Planning on looking into the unchained account return data, but to unblock osmosis dev, @toshiSat has confirmed this as working and it makes sense to me to keep moving forward right now as we can.

PR #892, opened 2022-07-20
feat(chain-adapters): export `EvmChainIds` type

0xApotheosis:
Exports `EvmChainIds` type for use in `web` to support EVM the work in https://github.com/shapeshift/web/pull/2164, allowing us to use tighter generics that are coupled with the `EvmBaseAdapter` types in `@shapeshiftoss/chain-adapters`.

[Example usage](https://github.com/shapeshift/web/blob/433824ea7438cb847243f8bac7bf90450bec73fd/src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx#L54):

```ts
const fees = estimatedFees[feeType] as FeeData<EvmChainIds>
```

Contributes to https://github.com/shapeshift/web/issues/2156.

### DIFF
--- a/packages/chain-adapters/src/evm/index.ts
+++ b/packages/chain-adapters/src/evm/index.ts
@@ -1,2 +1,4 @@
+export { EvmChainIds } from './EvmBaseAdapter'
+
 export * as ethereum from './ethereum'
 export * as avalanche from './avalanche'


gomesalexandre (approved):
LGTM 🍬 

We'll probably have to consolidate the "EVM ChainId" terminology at some point, as https://github.com/shapeshift/web/pull/2165 is also using it but is referring to the "ETH ChainId" of EVM chains vs. this referring to an actual CAIP.

In the cases we're referring to the chainReference, we could use the `ethChainId` terminology, what do you think? If we do so, we'd have to rename our current `ethChainId` to `ethereumChainId` to avoid conflicts though (it's currently the wrong terminology anyway since ethereum is the network, ether is the asset, and lowercase eth is typically used to refer to JSON-RPC calls and the like)

kaladinlight (approved):

kaladinlight:
> LGTM candy
> 
> We'll probably have to consolidate the "EVM ChainId" terminology at some point, as [shapeshift/web#2165](https://github.com/shapeshift/web/pull/2165) is also using it but is referring to the "ETH ChainId" of EVM chains vs. this referring to an actual CAIP.
> 
> In the cases we're referring to the chainReference, we could use the `ethChainId` terminology, what do you think? If we do so, we'd have to rename our current `ethChainId` to `ethereumChainId` to avoid conflicts though (it's currently the wrong terminology anyway since ethereum is the network, ether is the asset, and lowercase eth is typically used to refer to JSON-RPC calls and the like)

Not sure I like the idea of changing our caip naming conventions even more and having a diff prefix for asset vs chain. if anything I think we should refer to the chainReference value as just that `ethChainRef` or `ethChainReference` or we can refer to it as `ethNetwork` or `ethNetworkVersion` to match rpc vernacular: https://ethereum.org/en/developers/docs/apis/json-rpc/#net_version

PR #893, opened 2022-07-20
feat(types): export Asset type from asset-service

gomesalexandre:
BREAKING CHANGE: Asset is now exported from asset-service, and all consumers should now import it from it

### Description

This moves the Asset type export to asset-service, allowing us to use the `AssetId` / `ChainId` types for the `assetId` and `chainId` properties of `Asset` without any circular dependencies.

It should give us more confidence in web typings e.g accessors that should be of type `AssetId` or `ChainId`


kaladinlight:
This creates a circular dependency between caip and types. More accurate fix would be moving the asset service types out into the asset service package and then they could utilize the caip types there

gomesalexandre:
> This creates a circular dependency between caip and types. More accurate fix would be moving the asset service types out into the asset service package and then they could utilize the caip types there

That sounds good 👍🏽 less reliance on the types package, and more types living where they really should i.e in their domain package. Moved this to draft for now

kaladinlight (dismissed):
there is a new conflict, but this update looks good to me

gomesalexandre:
> there is a new conflict, but this update looks good to me

Mergefixed! Also ran `local-ci` as a sanity check which is very happy locally

gomesalexandre:
Unrugged conflicts again

kaladinlight (approved):

0xdef1cafe (approved):

elmutt (approved):

PR #894, opened 2022-07-20
feat(caip): use correct ether/ethereum/bitcoin terminology for AssetI…

gomesalexandre:
…d and ChainId constants

This changes the naming of the constants to match the right asset/network terminology:

- `eth` -> `ethereum` for the network name
- `eth` -> `ether` for the asset name
- `btc` -> `bitcoin` for the network name
- `btc` -> `bitcoin` for the asset name (to match the ether terminology using the full name, in theory it should be `Bitcoin` for the network and `bitcoin` for the asset but given the casing we use for these constants, it's impossible to respect it so `bitcoin` is the closest we can get)

With these changes in, we'll have the right terminologies, and additionally, if we agree to use that, we'll be able to use `ethChainId` instead of `evmChainId` to refer to EVM chainIds (the actual EVM chainId that's gotten from the `eth_chainId` a.k.a a `chainReference`). This way, we won't have conflicts with the new `EvmChainIds` terminology added in https://github.com/shapeshift/lib/pull/892. Currently, it both refers to a CAIP and to an actual EVM ChainId (a ChainReference).


kaladinlight:
https://github.com/shapeshift/lib/pull/892#issuecomment-1190482885

kaladinlight:
> > > Not sure I like the idea of changing our caip naming conventions even more and having a diff prefix for asset vs chain
> > 
> > 
> > I don't think this is changing any naming conventions per se, we e.g use `bitcoinChainId` and `ethereumChainId` a few places across lib. If we explicitly _want_ to use the same prefix for asset and chain, and keep it as it is, then that's fine
> > > if anything I think we should refer to the chainReference value as just that `ethChainRef` or `ethChainReference` or we can refer to it as `ethNetwork` or `ethNetworkVersion` to match rpc vernacular: https://ethereum.org/en/developers/docs/apis/json-rpc/#net_version
> > 
> > 
> > I think the latter is perfect for the purpose of matching the RPC naming, vs. sticking to the CAIP vernacular when referring to EVM-specific chainId terminology.
> 
> Ultimately I think the full "chain" name makes the most sense as a prefix and avoids ambiguity between something where the coin symbol doesn't match the chain (ie. Cosmos/ATOM). So if we were to update naming, I personally like `bitcoinChainId`, `ethereumChainId`, `cosmosChainId`, `osmosisChainId`, etc. I don't think trying to differentiate the prefix between chainId and assetId is useful though and I would just need to think of what the asset name might be I feel like instead of always just thinking in terms of chain name imo.
> 
> +1 on the rpc naming

gomesalexandre:
@kaladinlight If we're going to use the RPC naming, I'll keep this in draft for now as we discuss it.

We won't have any naming conflicts, thus we can probably keep the current naming. I still believe that going forward, we should be more specific in the const we export, not only because of the correctness of honoring the chain/asset naming conventions for each asset/network but also for the precise reason you mention, where the coin name doesn't match the chain.

Given that currently, we only export const for the native assets, what about something like:

- `ethereumNativeAssetId`
- `ethereumChainId`

? This way, we would use the network name (with the right terminology, ethereum vs. eth currently) and make it clear we're refferring to the native asset of that network specifically. If in the future, we want to export e.g other `AssetId`s, then we could do `ethereumUsdcAssetId` which follows the same syntax, making it clear we refer to USDC on the Ethereum network specifically.

PR #898, opened 2022-07-22
fix: silently fail if a swapper has already been added instead of throwing

elmutt:
fix: silently fail if a swapper has already been added instead of throwing

We need this in order to cleanly handle async swapper initialization without having to catch every error if a swapper has already been added

### DIFF
--- a/packages/swapper/src/manager/SwapperManager.test.ts
+++ b/packages/swapper/src/manager/SwapperManager.test.ts
@@ -64,13 +64,12 @@ describe('SwapperManager', () => {
       expect(manager.getSwapper(SwapperType.ZrxEthereum)).toBeInstanceOf(ZrxSwapper)
     })
 
-    it('should throw an error if adding an existing chain', () => {
-      const swapper = new SwapperManager()
-      expect(() => {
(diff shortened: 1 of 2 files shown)


0xdef1cafe:
@elmutt and i discussed this at length. this is basically required for web, unless we add the concept of swappers into plugins and move all existing swappers into plugins.

this is required to allow us to have swapper manager singleton in the client, and deal with feature flags being turned on and off dynamically at runtime.

we could get really fancy on the client side, and use the `partitionCompareWith` functions that the plugin manager already uses to add chain adapters to avoid this, but given that the swapper manager is dealing with a language-native `Map` - the language ensures we can't actually add a swapper with the same key again, so adding the same existing swapper is basically a noop, and has the same desired effect.

0xdef1cafe (dismissed):

0xdef1cafe:
@elmutt i do wonder if we should return the new swapper on add, vs the old one, such that re-adding is like an upsert?

0xdef1cafe (approved):

kaladinlight (approved):

PR #900, opened 2022-07-22
chore: cowswap remove feeAsset

### LINKED ISSUE #844: CowSwapper : tidy up feeAsset
### Overview

As `feeAsset` is always `WETH` in the context of `CowSwapper`, we want to define it as a module-level export, or a class member instead of a function parameter.

Also we need to address all remaining review comments there : https://github.com/shapeshift/lib/pull/796#pullrequestreview-1023163870

At least 2 options to do this : 

- hardcode it in a constant file
- fetch it once from assetService in the constructor / initialize method

Follow up from review comment : https://github.com/shapeshift/lib/pull/796#discussion_r907949865

### Acceptance Criteria

`feeAsset` should be statically defined

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

2 hours

0xdef1cafe:
cowswap only supports erc20s, not ETH itself.

previously we were passing in a feeAsset to check for this case, and unnecessarily polluting the
interface and making the consumption of cowswapper messier.

given this is a swapper specific issue, and not going to change based on the fundamental design of
cowswap, and we already depend on the asset service, we can safely hardcode WETH in place of ETH
here to handle this specific case and expose a cleaner interface.

closes #844

- chore: don't require feeAsset in cow swapper dependencies
- test: fix cowswap tests to remove feeAsset

### DIFF
--- a/packages/swapper/src/manager/SwapperManager.test.ts
+++ b/packages/swapper/src/manager/SwapperManager.test.ts
@@ -8,7 +8,6 @@ import { Swapper, SwapperType } from '../api'
 import { CowSwapper, ThorchainSwapper, ZrxSwapper } from '../swappers'
 import { CowSwapperDeps } from '../swappers/cow/CowSwapper'
 import { ThorchainSwapperDeps } from '../swappers/thorchain/types'
-import { WETH } from '../swappers/utils/test-data/assets'
 import { ZrxSwapperDeps } from '../swappers/zrx/types'
 import { SwapperManager } from './SwapperManager'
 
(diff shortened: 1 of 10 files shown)


elmutt (approved):

chxash (approved):
Tested locally with appropriate web branch and it works fine !

gomesalexandre (approved):
LGTM, tested linked on the relevant web PR and didn't find any regressions with the previous moo behavior

🤠

0xApotheosis (approved):
👌

kaladinlight (approved):

Issue #875, opened 2022-07-13
Cleanup `chainIdToFeeAssetId`


gomesalexandre:
### Overview

Currently, `chainIdToFeeAssetId` is used in web to get the fee/native AssetId for a given `ChainId`.

However, chain-adapters contain a `getFeeAssetId` method that will return that same `AssetId`.

It should be used instead, and the `chainIdToFeeAssetId` function removed.

### References and additional details

https://github.com/shapeshift/lib/search?q=getFeeAssetId

### Acceptance Criteria

- chainIdToFeeAssetId is removed
- Web consumptions are replaced with some usage of chain-adapters e.g `getChainAdapters().get(chainId)?.getFeeAssetId()`

0xdef1cafe:
closed by https://github.com/shapeshift/web/pull/2159 ?

gomesalexandre:
> closed by [shapeshift/web#2159](https://github.com/shapeshift/web/pull/2159) ?

Almost, this one-liner will do https://github.com/shapeshift/lib/pull/907

PR #908, opened 2022-07-25
feat(caip): remove getFeeAssetIdFromAssetId / chainIdToAssetId

gomesalexandre:
### Description 

This removes the `getFeeAssetIdFromAssetId` function and its underlying `chainIdToAssetId` mapping:

https://github.com/shapeshift/lib/blob/7b4e7565f15707d494c2c7ddfd499113f53fc04b/packages/caip/src/utils.ts#L29-L36

There is no reason to keep this mapping as it's just used by swapper and we can simply call `adapter.get(chainId)?.getFeeAssetId()` to get said native AssetId, then pass it to the swapper methods.

This way, we won't have to maintain a mapping of ChainId to native AssetId.

### Issue

- closes https://github.com/shapeshift/lib/issues/875


gomesalexandre:
> Can the thorchain part of this PR wait until buildTrade is done?
> 
> I have some overlapping changes in thor getQuote for that PR as well

Sure, I'll leave it in draft for now - I've used swapper as a scope just to make it clear in release notes this is breaking changes for swapper, but the core of this PR is really the removal of the method/mapping in caip so I'd like the diff to contain it all at once.

elmutt (dismissed):

elmutt (approved):

packages/caip/src/utils.test.ts
```diff
@@ -1,4 +1,4 @@
-import { ChainNamespace, ChainReference } from 'packages/caip/src/chainId/chainId'
+import { ChainNamespace, ChainReference } from '@shapeshiftoss/caip'
```
kaladinlight, line 1: relative import within same package?

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -53,7 +53,7 @@ export const getThorTradeQuote: GetThorTradeQuote = async ({ deps, input }) => {
 
     const sellAssetId = sellAsset.assetId
     const buyAssetId = buyAsset.assetId
-    const feeAssetId = getFeeAssetIdFromAssetId(buyAssetId)
+    const feeAssetId = adapter.getFeeAssetId()
```
kaladinlight, line 56: is this `feeAssetId` logic even necessary or valid? we aren't using it here and are ultimately doing this logic/check in `estimateTradeFee`.

also as future cleanup/issue, the usage of `chainId` in conjunction with `buyAsset` and `sellAsset` is super confusing (which chain id is it talking about), and not needed since we can get the applicable `chainId` of the desired asset.

packages/caip/src/utils.test.ts
gomesalexandre: Might as well, ty ser 0003b90fe9669ee7dab9251167f968bac07759bf

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
gomesalexandre: Yeah, I'm agnostic-ish to this but would tend to agree with it being unnecessary. I can see the rationale behind the early check and not going further down the call stack for no reason, but it's not like there's anything heavy happening before the `estimateTradeFee` check anyway:

https://github.com/shapeshift/lib/blob/25cc8b3bfd03225efaf2ded70a4887a8f301979f/packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts#L54-L55

Let's keep the check in `estimateTradeFee` to keep things simpler and remove duplication

gomesalexandre: Cleaned up in 1cfe58a9fab2852d5e834577d79afdf0a72580cd

kaladinlight: Also with the early `adapter` check and considering `getFeeAssetId` is required on the chain adapter it isn't really necessary regardless. Ultimately I feel like a lot of this interface can be cleaned up to make things more clear as to what's happening, but discussions for another day. Thanks @gomesalexandre :partying_face:

kaladinlight (dismissed):

0xdef1cafe (approved):

kaladinlight (dismissed):

kaladinlight (approved):

0xdef1cafe (approved):

PR #909, opened 2022-07-26
feat: cowswap tx parser

### LINKED ISSUE #901: cowswap tx parser
### Overview

we currently display cowswap trades as follows

image

we want to display them like this example of a 0x trade

image

implement something like this

https://github.com/shapeshift/lib/blob/main/packages/unchained-client/src/evm/ethereum/parser/zrx.ts#L5-L6

reference tx

https://etherscan.io/tx/0x6796daaa190af29881b019b8d1119543086edf96c194edaf7d49dea8d100faa9

cowswap contract

https://etherscan.io/address/0x9008d19f58aabd9ed0d60971565aa8510560ab41

### Acceptance Criteria

we can parse cowswap trades

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xdef1cafe:
closes #901 

- feat: cowswap tx parser
- chore: update user facing enum values

thanks to @kaladinlight and @gomesalexandre for context - we had a bunch of duplicate types between chain adapters and unchained client. this consolidates `TradeType`, `TransferType`, `Dex`, and `TxStatus` to `unchained-client` which `chain-adapters` imports. the values of these were also updated to user facing values which web can consume and render things more nicely to users.


packages/unchained-client/src/types.ts
```diff
@@ -1,6 +1,8 @@
+// these are user facing values, and should be rendered as such
```
0xdef1cafe, line 1: @kaladinlight plz confirm this comment is valid, and that we don't rely on these internally for anything, and they're purely presentational.

they're referenced in web here https://github.com/shapeshift/web/pull/2232/files#diff-2575b4bbecd755e30395862fd46ebb14c318ded832e13b7b14edd17cb4251c54L57

tests pass but 🤷

packages/unchained-client/src/types.ts
```diff
@@ -28,9 +30,10 @@ export interface Trade {
   type: TradeType
 }
 
+// these are user facing values, and should be rendered as such
```
0xdef1cafe, line 33: same as above

packages/unchained-client/src/evm/ethereum/parser/constants.ts
```diff
@@ -7,3 +7,4 @@ export const FOXY_STAKING_CONTRACT = '0xee77aa3Fd23BbeBaf94386dD44b548e9a785ea4b
 export const WETH_CONTRACT_MAINNET = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
 export const WETH_CONTRACT_ROPSTEN = '0xc778417E063141139Fce010982780140Aa0cD5Ab'
 export const UNI_V2_FOX_STAKING_REWARDS_V3 = '0x212ebf9FD3c10F371557b08E993eAaB385c3932b'
+export const COWSWAP_CONTRACT_MAINNET = '0x9008D19f58AAbD9eD0D60971565AA8510560ab41'
```
0xdef1cafe, line 10: reference tx for the astute reviewer

https://etherscan.io/tx/0x6796daaa190af29881b019b8d1119543086edf96c194edaf7d49dea8d100faa9

packages/unchained-client/src/evm/ethereum/parser/cowswap.ts
```diff
@@ -0,0 +1,26 @@
...
+import { COWSWAP_CONTRACT_MAINNET } from './constants'
+
+export class Parser implements SubParser<Tx> {
+  async parse(tx: Tx): Promise<TxSpecific | undefined> {
+    if (!txInteractsWithContract(tx, COWSWAP_CONTRACT_MAINNET)) return
+    if (!(tx.tokenTransfers && tx.tokenTransfers.length)) return
+
+    const trade = {
+      dexName: Dex.CowSwap,
+      type: TradeType.Trade
```
gomesalexandre, line 13: This is wrongly taken from the ZRX and Thor parser implementations.

We should either consolidate the `TradeType` enum, or use the `chain-adapters` one. Trade type Txs all fall back to the default case in web.

packages/unchained-client/src/types.ts
gomesalexandre: Yeah, these are not only user-facing values. We use them in web to render the relevant transaction type's component and they have a different casing than the similarly-named `chain-adapters` enum, which results in ZRX, Thor, and, with this diff, CowSwap Txs falling back to the generic (unknown) Tx component.

gomesalexandre (dismissed):
Tested with `unchained-client` linked locally and CowSwap Txs are parsed in a way that web wrongly switches on, resulting in unknown Tx rows, see the web PR review: 

https://github.com/shapeshift/web/pull/2232#pullrequestreview-1050785876
image

packages/unchained-client/src/types.ts
kaladinlight: That is correct, they are added as ways to narrow the trade type to a specific dex if we had different UI or it looks like we are using those strings verbatim for UI display. Nothing internally

packages/unchained-client/src/types.ts
kaladinlight: So `TransferType`, `TradeType` and `Status` were all types that lived in unchained initially and we created chain adapters types at the time so we weren't passing back end unchained types through to web. This has changed a bit now that we have the parser in lib and we can probably clean up these types so we just have one reference of them to use for both chain adapters and unchained client

kaladinlight: But yes, they are all values for the client to render as necessary ultimately

packages/unchained-client/src/types.ts
```diff
@@ -28,9 +30,10 @@ export interface Trade {
...
 export enum TradeType {
-  Trade = 'trade',
-  Refund = 'refund'
+  Trade = 'Trade',
+  Refund = 'Refund'
```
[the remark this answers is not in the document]
kaladinlight: will fix for now unless you want to fix up types along with this pr

kaladinlight (approved):

Issue #759, opened 2022-06-07
CowSwap : replace '/api/v1/market' call with '/api/v1/quote' in getUsdRate method


chxash:
### Overview

 '/api/v1/market' has just been deprecated so it may be better to use '/api/v1/quote' instead.

### References and additional details

https://api.cow.fi/docs/#/default/get_api_v1_markets__baseToken___quoteToken___kind___amount_

### Acceptance Criteria

getUsdRate method is still working and the usdRate estimation is still correct

### Estimated effort

3 hours

0xdef1cafe:
@chxash is this still valid or has this been implemented?

chxash:
@0xdef1cafe still valid yes, rather a "nice to have" than anything

0xdef1cafe:
@chxash i hope you don't mind that i assigned you to this one - i'm going to pick up the tx parsing tickets, as IMO, they're a blocker to shipping, but it would be nice if we could get this one closed out as a low priority after we ship just to tie up loose ends such that we're not using deprecated endpoints in brand new code.

chxash:
@0xdef1cafe no problem, thank you for assigning me, will work on this asap !

PR #916, opened 2022-07-28
feat: replace '/api/v1/market' call with '/api/v1/quote' in getUsdRate

chxash:
This changes the endpoint used in `getUsdRate` method, because the one currently used has been deprecated.
I tested it in web, checked rates there and performed a swap.
closes https://github.com/shapeshift/lib/issues/759

Steps to test : 

- Link with web locally
- Start swap with ERC20 tokens (not ETH)
- Verify rates are consistent with current market
- Perform swap

### DIFF
--- a/packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.test.ts
+++ b/packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.test.ts
@@ -11,7 +11,7 @@ import { ETH, FOX, WBTC, WETH } from '../../utils/test-data/assets'
 import { CowSwapperDeps } from '../CowSwapper'
 import { CowTrade } from '../types'
 import { cowService } from '../utils/cowService'
-import { CowSwapQuoteApiInput } from '../utils/helpers/helpers'
+import { CowSwapSellQuoteApiInput } from '../utils/helpers/helpers'
 import { cowBuildTrade } from './cowBuildTrade'
 
 jest.mock('@shapeshiftoss/chain-adapters')
@@ -73,7 +73,7 @@ const feeData: FeeDataEstimate<KnownChainIds.EthereumMainnet> = {
   }
 }
 
-const expectedApiInputWethToFox: CowSwapQuoteApiInput = {
+const expectedApiInputWethToFox: CowSwapSellQuoteApiInput = {
   appData: '0x0000000000000000000000000000000000000000000000000000000000000000',
   buyToken: '0xc770eefad204b5180df6a14ee197d99d808ee52d',
   from: 'address11',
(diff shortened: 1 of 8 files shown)


0xdef1cafe (changes requested):
technically looks sounds, getting a lot of errors from the quote api - our standard UX is inputting values that cause errors, e.g.

image

we'll have to handle these error conditions more gracefully

chxash:
@0xdef1cafe I was finally able to reproduce this issue
This occurs in `getCowSwapTradeQuote.ts` (where `quote` api is already used) when `CowSwap` expects a fee greater than $10 (which is our minimum when lower value is provided by UX / the user).
This is only occurring with `quote` api when we are performing a `sell` operation, so the code in this PR where I am performing a `buy` is not at fault I think.

I have tried changing `getCowSwapMinMax.ts` logic in order to perform a quote/buy operation to retrieve a meaningful fee. But the fee can be very different for buy and sell operation.
So for now I only **increased our minimum value for `CowSwap` to $20** which should not bother the user too much given fees are commonly $1 - $10. Maybe we can create a separate issue to find a more robust way of estimating fee / calculating a minimum ?

0xdef1cafe:
> @0xdef1cafe I was finally able to reproduce this issue This occurs in `getCowSwapTradeQuote.ts` (where `quote` api is already used) when `CowSwap` expects a fee greater than $10 (which is our minimum input value when lower value is provided by UX / the user). This is only occurring with `quote` api when we are performing a `sell` operation, so the code in this PR where I am performing a `buy` is not at fault I think.
> 
> I have tried changing `getCowSwapMinMax.ts` logic in order to perform a quote/buy operation to retrieve a meaningful fee. But the fee can be very different for buy and sell operation. So for now I only **increased our minimum value for `CowSwap` to $20** which should not bother the user too much given fees are commonly $1 - $10. Maybe we can create a separate issue to find a more robust way of estimating fee / calculating a minimum ?

@shapeshift/product FYI

DiggyDiggy2:
Thanks! This sounds like a good approach. When I look at the trade activity from 0x there are only a few trades that are for very small amounts, and I'd assume those are our test trades. I think a $20 minimum is fine for cowswap.

gomesalexandre (commented):
Conceptually ok with this one, discussing about a potential regression I've found before stamping it

0xdef1cafe:
with a rebase and a link locally this is behaving fine, i'm conceptually ok too.

gomesalexandre (approved):
Tested and rates still LGTM.

Note that as part of this PR, it was found that inputting a value less than the minimum amount (10$ in develop, 20$ here) will not block the flow from being continued, and we will fall back to whichever is the minimum 

Develop:

image
image

This branch: 
image
image

Thanks @chxash for testing this one out on develop as well and confirming the behavior. This is **not** a regression so:

🐄

0xdef1cafe (approved):

PR #918, opened 2022-07-28
fix: don't return zero value fiat exchange rates

0xdef1cafe:
the upstream fiat exchange rate data source will return 0 for the current day for timezones with a
positive offset, e.g. australia

this filters out these invalid results

the client consuming this data will use the most recent rate rather than zero, fixing charts


0xdef1cafe:
closes https://github.com/shapeshift/web/issues/2199

kaladinlight (approved):

PR #930, opened 2022-08-01
fix: use fee asset instead of asset in calculation

elmutt:
use buyFeeAsset instead of buyAsset in fees calculation. 

The problem was the number of precision digits in the asset vs fee asset

This was missed because most erc20's are 18 digits like eth and wasn't noticed in testing until trying usdc

### DIFF
--- a/packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts
@@ -1,5 +1,5 @@
 import { Asset, AssetService } from '@shapeshiftoss/asset-service'
-import { adapters, fromAssetId, getFeeAssetIdFromAssetId } from '@shapeshiftoss/caip'
+import { adapters, fromAssetId } from '@shapeshiftoss/caip'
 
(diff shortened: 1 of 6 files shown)


packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts
```diff
@@ -59,40 +59,41 @@ export const estimateTradeFee = async (
     })
 
   const gasRate = inboundInfo.gas_rate
-  const { chainId, assetNamespace } = fromAssetId(buyAsset.assetId)
+  const { chainId: buyChainId, assetNamespace } = fromAssetId(buyAsset.assetId)
 
-  const feeAssetId = getFeeAssetIdFromAssetId(buyAsset.assetId)
-  if (!feeAssetId)
+  const buyFeeAssetId = getFeeAssetIdFromAssetId(buyAsset.assetId)
```
0xdef1cafe, line 64: given we're trying to kill this function here https://github.com/shapeshift/lib/pull/908

and we (likely?) have the chain adapter in the deps to be able to get the fee asset, any chance you could quickly switch this over?

gomesalexandre (approved):
Untested but logic looks sound and isolated to a specific area of a feature-flagged swapper

PR #932, opened 2022-08-01
feat: add bitcoincash

kaladinlight:


packages/chain-adapters/src/utils/utxoUtils.ts
```diff
@@ -38,7 +45,7 @@ export const utxoAccountParams = (
         scriptType: BTCInputScriptType.SpendAddress,
         bip44Params: {
           purpose: 44,
-          coinType: 3,
+          coinType: Number(ASSET_REFERENCE.Dogecoin),
```
0xdef1cafe, line 48: 👍

packages/types/src/base.ts
```diff
@@ -10,6 +10,7 @@ export enum KnownChainIds {
   EthereumMainnet = 'eip155:1',
   AvalancheMainnet = 'eip155:43114',
   BitcoinMainnet = 'bip122:000000000019d6689c085ae165831e93',
+  BitcoincashMainnet = 'bip122:000000000000000000651ef99cb9fcbe',
```
gomesalexandre, line 13: To keep the casing consistent? e.g `bitcoincash` when lowercased, `bitcoinCash` when camelcased... `BitcoinCash` when PascalCased? 

```suggestion
  BitcoinCashMainnet = 'bip122:000000000000000000651ef99cb9fcbe',
```

kaladinlight: updated

0xdef1cafe (approved):

elmutt (approved):

gomesalexandre (approved):
Untested but gave it a few review passes and logic looks sound - anything weirdy here would be spotted in web but this effectively looks the same as other UTXO implementations

PR #933, opened 2022-08-01
chore(caip): add include & exclude patterns to caip's `tsconfig`

0xApotheosis:
Stops the IDE complaining about importing files not listed as part of the project, particularly in `packages/caip/src/adapters/coingecko/generated/index.ts`.

Screen Shot 2022-08-02 at 7 27 08 am


gomesalexandre (approved):
Does what it says, tested with main:

image


And with this diff built:
image

PR #935, opened 2022-08-01
feat: additonal thorchain asset support

elmutt:
- separate `buySupportedChainIds` from `sellSupportedChainIds` so we can easily add new buyAssets without having to implementing the sell logic at the same time.

- Add buy support for `doge` `ltc` `atom`
Screen Shot 2022-08-01 at 5 21 51 PM

To test:

link locally. enable thorchain ff. enable litecoin ff. see that you can trade into ltc, doge & atom

### DIFF
--- a/packages/caip/src/adapters/thortrading/index.ts
+++ b/packages/caip/src/adapters/thortrading/index.ts
@@ -1,3 +1,4 @@
+import { btcAssetId, cosmosAssetId, dogeAssetId, ethAssetId, ltcAssetId } from '../../constants'
 import { AssetId } from './../../assetId/assetId'
 
 // derived from https://midgard.thorchain.info/v2/pools
@@ -47,12 +48,12 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
     'eip155:1/erc20:0xdbdb4d16eda451d0503b854cf79d55697f90c8df',
   'ETH.AAVE-0X7FC66500C84A76AD7E9C93437BFC5AC33E2DDAE9':
     'eip155:1/erc20:0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9',
-  'BTC.BTC': 'bip122:000000000019d6689c085ae165831e93/slip44:0',
-  'ETH.ETH': 'eip155:1/slip44:60',
-  'LTC.LTC': 'bip122:12a765e31ffd4059bada1e25190f6e98/slip44:2',
-  'DOGE.DOGE': 'bip122:00000000001a91e3dace36e2be3bf030/slip44:3',
-  'GAIA.ATOM': 'cosmos:cosmoshub-4/slip44:118'
-  //  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145' // uncomment when we support bch
+  'BTC.BTC': btcAssetId,
+  'ETH.ETH': ethAssetId,
+  'LTC.LTC': ltcAssetId,
+  'DOGE.DOGE': dogeAssetId,
+  'GAIA.ATOM': cosmosAssetId
+  //  'BCH.BCH': bchAssetId // uncomment when we support bch
 }
 
 const invert = <T extends Record<string, string>>(data: T) =>
--- a/packages/caip/src/adapters/thortrading/index.ts
+++ b/packages/caip/src/adapters/thortrading/index.ts
@@ -51,8 +51,8 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
   'ETH.ETH': 'eip155:1/slip44:60',
   'LTC.LTC': 'bip122:12a765e31ffd4059bada1e25190f6e98/slip44:2',
   'DOGE.DOGE': 'bip122:00000000001a91e3dace36e2be3bf030/slip44:3',
-  'GAIA.ATOM': 'cosmos:cosmoshub-4/slip44:118',
-  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145'
+  'GAIA.ATOM': 'cosmos:cosmoshub-4/slip44:118'
+//  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145' // uncomment when we support bch
 }
 
 const invert = <T extends Record<string, string>>(data: T) =>
--- a/packages/caip/src/adapters/thortrading/index.ts
+++ b/packages/caip/src/adapters/thortrading/index.ts
@@ -52,7 +52,7 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
   'LTC.LTC': 'bip122:12a765e31ffd4059bada1e25190f6e98/slip44:2',
   'DOGE.DOGE': 'bip122:00000000001a91e3dace36e2be3bf030/slip44:3',
   'GAIA.ATOM': 'cosmos:cosmoshub-4/slip44:118'
-//  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145' // uncomment when we support bch
+  //  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145' // uncomment when we support bch
 }
 
 const invert = <T extends Record<string, string>>(data: T) =>
--- a/packages/caip/src/adapters/thortrading/index.ts
+++ b/packages/caip/src/adapters/thortrading/index.ts
@@ -48,7 +48,11 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
   'ETH.AAVE-0X7FC66500C84A76AD7E9C93437BFC5AC33E2DDAE9':
     'eip155:1/erc20:0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9',
   'BTC.BTC': 'bip122:000000000019d6689c085ae165831e93/slip44:0',
-  'ETH.ETH': 'eip155:1/slip44:60'
+  'ETH.ETH': 'eip155:1/slip44:60',
+  'LTC.LTC': 'bip122:12a765e31ffd4059bada1e25190f6e98/slip44:2',
+  'DOGE.DOGE': 'bip122:00000000001a91e3dace36e2be3bf030/slip44:3',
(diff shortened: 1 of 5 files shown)


packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
```diff
@@ -31,11 +31,21 @@ import { thorService } from './utils/thorService'
...
 export class ThorchainSwapper implements Swapper<ChainId> {
   readonly name = 'Thorchain'
-  private swapSupportedChainIds: Record<ChainId, boolean> = {
+  private sellSupportedChainIds: Record<ChainId, boolean> = {
     [KnownChainIds.EthereumMainnet]: true,
     [KnownChainIds.BitcoinMainnet]: true
   }
-  private supportedAssetIds: AssetId[] = []
+
+  private buySupportedChainIds: Record<ChainId, boolean> = {
```
0xdef1cafe, line 39: is there a reason these are maps vs just arrays?

elmutt: used as a lookup

packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts
```diff
@@ -88,21 +104,36 @@ export const estimateTradeFee = async (
...
-    case 'eip155:1':
+    case KnownChainIds.LitecoinMainnet:
+      return fromBaseUnit(
+        bnOrZero(ltcEstimate(gasRate)).times(buyFeeAssetRatio).dp(0),
+        buyFeeAsset.precision
+      )
+    case KnownChainIds.CosmosMainnet:
+      return fromBaseUnit(
+        bnOrZero(gaiaEstimate(gasRate)).times(buyFeeAssetRatio).dp(0),
+        8 // because thorchain likes to be inconsistent for no good reason
```
0xdef1cafe, line 125:
```suggestion
        THORCHAIN_FIXED_PRECISION
```

0xdef1cafe: should this actually be `THORCHAIN_FIXED_PRECISION` everywhere above for safety - this is a protocol design thing and those other assets may happen to match but only by chance?

elmutt: its just inconsistent. It wont work for eth or erc20. But it would work for utxo because it happens to be the same value. So im not sure which to use. Minimizing THORCHAIN_FIXED_PRECISION to this one spot for now seems correct enough to me

elmutt: The eth rate comes back in terms of gwei (10^9), the btc comes back in sats (which also happens to be the thorchain precision). Atom comes back in the thochain precision.

Looking at it this way it feels like atom is the special case at least until we have more chains and want to change it up

packages/caip/src/adapters/thortrading/index.ts
```diff
@@ -48,7 +48,11 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
   'ETH.AAVE-0X7FC66500C84A76AD7E9C93437BFC5AC33E2DDAE9':
     'eip155:1/erc20:0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9',
   'BTC.BTC': 'bip122:000000000019d6689c085ae165831e93/slip44:0',
-  'ETH.ETH': 'eip155:1/slip44:60'
+  'ETH.ETH': 'eip155:1/slip44:60',
+  'LTC.LTC': 'bip122:12a765e31ffd4059bada1e25190f6e98/slip44:2',
+  'DOGE.DOGE': 'bip122:00000000001a91e3dace36e2be3bf030/slip44:3',
+  'GAIA.ATOM': 'cosmos:cosmoshub-4/slip44:118'
+  //  'BCH.BCH': 'bip122:000000000000000000651ef99cb9fcbe/slip44:145' // uncomment when we support bch
```
0xdef1cafe, line 55: while we're in here we might as well use the `btcAssetId` type constants for consistency

packages/swapper/src/swappers/thorchain/utils/estimateTradeFee/estimateTradeFee.ts
```diff
@@ -87,34 +62,11 @@ export const estimateTradeFee = async (
...
-        code: SwapErrorTypes.VALIDATION_FAILED,
-        details: { chainId: buyChainId }
-      })
-  }
+  return fromBaseUnit(
+    THOR_TRADE_FEE_MULTIPLIERS[buyChainId as SUPPORTED_BUY_CHAINS]
+      .times(buyFeeAssetRatio)
+      .times(gasRate)
+      .dp(0),
+    buyFeeAsset.precision
```
0xdef1cafe, line 70: is this always going to be `THORCHAIN_FIXED_PRECISION`?

elmutt: maybe we should for consistency now that its using the map. will need to change the hard coded eth and atom values but seems a bit cleaner

elmutt: maybe better:

dont use fromBaseUnit() and have the multiplier values account for it all

packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
```diff
@@ -48,14 +58,19 @@ export class ThorchainSwapper implements Swapper<ChainId> {
         `${this.deps.midgardUrl}/pools`
       )
 
-      const supportedAssetIds = responseData.reduce<AssetId[]>((acc, midgardPool) => {
+      this.supportedSellAssetIds = responseData.reduce<AssetId[]>((acc, midgardPool) => {
         const assetId = adapters.poolAssetIdToAssetId(midgardPool.asset)
-        if (!assetId || !this.swapSupportedChainIds[fromAssetId(assetId).chainId]) return acc
+        if (!assetId || !this.sellSupportedChainIds[fromAssetId(assetId).chainId]) return acc
```
gomesalexandre, line 63: Should we keep a consistent terminology? e.g `supportedSellChainIds` for the second

packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
```diff
@@ -48,14 +58,19 @@ export class ThorchainSwapper implements Swapper<ChainId> {
...
         acc.push(assetId)
         return acc
       }, [])
 
-      this.supportedAssetIds = supportedAssetIds
+      this.supportedBuyAssetIds = responseData.reduce<AssetId[]>((acc, midgardPool) => {
+        const assetId = adapters.poolAssetIdToAssetId(midgardPool.asset)
+        if (!assetId || !this.buySupportedChainIds[fromAssetId(assetId).chainId]) return acc
```
gomesalexandre, line 70: Same question

packages/swapper/src/swappers/thorchain/utils/constants.ts
```diff
@@ -1,10 +1,31 @@
...
+// https://discord.com/channels/838986635756044328/997675038675316776/998552541170253834
+// This is still not "perfect" and tends to overestimate by a small randomish amount
+// TODO figure out if its possible to accurately estimate the outbound fee.
+// Neither the discord nor official docs are correct
+export const THOR_TRADE_FEE_MULTIPLIERS: Record<SUPPORTED_BUY_CHAINS, BigNumber> = {
+  [KnownChainIds.BitcoinMainnet]: bn(0.00002),
+  [KnownChainIds.DogecoinMainnet]: bn(0.00002),
+  [KnownChainIds.LitecoinMainnet]: bn(0.000005),
+  [KnownChainIds.CosmosMainnet]: bn(0.00000002),
+  [KnownChainIds.EthereumMainnet]: bn(0.00024)
```
gomesalexandre, line 23: Non-blocking, but `toString()` these? Passing primitives around has a lot less overhead and serializes better. Although in this case it doesn't matter as we just `THOR_TRADE_FEE_MULTIPLIERS[buyChainId].times()` without any passing around, so your call

gomesalexandre (commented):
Conceptually good with this one except a few non-blocking comments.

@elmutt Can you add testing steps before I stamp this? 🙏🏽

packages/caip/src/adapters/thortrading/index.ts
```diff
@@ -47,8 +48,12 @@ const thorPoolIdAssetIdSymbolMap: Record<string, AssetId> = {
...
   'ETH.AAVE-0X7FC66500C84A76AD7E9C93437BFC5AC33E2DDAE9':
     'eip155:1/erc20:0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9',
-  'BTC.BTC': 'bip122:000000000019d6689c085ae165831e93/slip44:0',
-  'ETH.ETH': 'eip155:1/slip44:60'
+  'BTC.BTC': btcAssetId,
+  'ETH.ETH': ethAssetId,
+  'LTC.LTC': ltcAssetId,
+  'DOGE.DOGE': dogeAssetId,
+  'GAIA.ATOM': cosmosAssetId
+  //  'BCH.BCH': bchAssetId // uncomment when we support bch
```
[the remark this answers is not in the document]
elmutt: fails on initialize in fromAssetId or something like that. cant do it yet

packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
elmutt: its only used in this file and not passed around anywhere. I dont mind this name

packages/swapper/src/swappers/thorchain/utils/constants.ts
elmutt: I agree string an be useful in some cases but I don't think theres much benefit and would still need to be converted back to a bignumber for math and then back into a string

packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
kaladinlight: will just say ocd gang agrees with @gomesalexandre , but not blocking

kaladinlight (approved):

PR #939, opened 2022-08-02
feat: detect completed txs and throw error on fail

elmutt:
Detect completed & failed transactions.

Will have a followup pr to translate error toast on trade error on web

To test:

link to web. enable thor. trade into a coin that completed quickly (eth to ltc is good) and see the trade spinner until it completes


packages/swapper/src/swappers/thorchain/types.ts
```diff
@@ -21,6 +21,21 @@ export type PoolResponse = {
...
+type action = {
+  date: string
+  height: string
+  out: [out]
+  status: string
+  type: string
+}
+
+export type ActionsResponse = {
+  actions: [action]
```
gomesalexandre, line 36: `T[]` (preferred for consistency) or `Array<T>` 🙏🏽

gomesalexandre (commented):
Tested with a monkey patch to allow for sub-minimum swap, does what it says on the label, although we will need some changes in web: Currently, the confirm spinner spins until the swap is complete, but the flow for other swaps e.g ZRX is to have the arrow icon change to a spinner while the trade isn't complete:

image

This logic happens here:

https://github.com/shapeshift/web/blob/536c9d00494edd2d81985a0c8be708d692007291/src/components/Trade/TradeConfirm/TradeConfirm.tsx#L154

packages/swapper/src/swappers/thorchain/types.ts
elmutt: it really doesnt matter

0xdef1cafe: `[action]` is an array of length exactly 1

elmutt: derp

elmutt: fixed

packages/swapper/src/swappers/thorchain/types.ts
```diff
@@ -21,6 +21,21 @@ export type PoolResponse = {
   volume24h: string
 }
 
+type MidgardActionOut = {
+  txID: string
+}
+type MidgardAction = {
+  date: string
+  height: string
+  out: [MidgardActionOut]
```
0xdef1cafe, line 30: is this guaranteed to be an array of exactly length 0?

0xdef1cafe: this is not an opinion or stylistic - `[Type]` is not the same as `Type[]`

0xdef1cafe (changes requested):

packages/swapper/src/swappers/thorchain/types.ts
elmutt: ah yes good point. long day whoops

gomesalexandre: `MidgardActionsResponse['actions']` has been updated but this is still there. Again, do we want this to be a tuple?

gomesalexandre (commented):
Question

gomesalexandre (approved):
LGTM

Swapped ATOM into LTC and got the spinner while the swap was completing on Thor / green check after it completed

image

PR #941, opened 2022-08-02
fix(chain-adapters): disable segwit for litecoin and use correct bitcoincash parser

kaladinlight:


0xApotheosis (approved):
LG, + nice file rename.

0xdef1cafe (approved):

PR #944, opened 2022-08-03
feat: improve chain adapter types to expose relevant data

kaladinlight:
- update discriminator to string union over enum value for better intellisense completion
- create chain adapters alias for pass through types from unchained
- compose chain adapters types from unchained types if there is data mutation occurring in chain adapters
- intended use would be to interact with chain adapters types specifically from web even if they are just an alias to an unchained-type under the hood to keep the types scoped to chain adapters which is what the web client is interacting with

we can decide on desired naming in an effort to reduce type changes just over a key name as well for simplicity


packages/chain-adapters/src/types.ts
```diff
@@ -83,43 +83,16 @@ export type SubscribeTxsInput = {
   accountType?: UtxoAccountType
 }
 
-export type TxFee = {
-  assetId: string
-  value: string
-}
+export type TransferType = unchained.TransferType
+export type TradeType = unchained.TradeType
```
gomesalexandre, line 89: We'll probably want to also expose the TxMetadata types to make consumption easier https://github.com/shapeshift/web/pull/2257/files#diff-1a8c76a31242bf7e97ec874047f2c1dcfb6d4e47d9d1c4811b7cc8db277f9bbaR3-R4

packages/chain-adapters/src/types.ts
```diff
@@ -83,43 +83,16 @@ export type SubscribeTxsInput = {
...
 
-export type Transaction<T extends ChainId> = {
-  address: string
-  blockHash?: string
-  blockHeight: number
-  blockTime: number
+export type Transaction<T extends ChainId> = Omit<unchained.StandardTx, 'transfers'> & {
```
gomesalexandre, line 91: Will remove the generic as the `chain` field we use it for is unused

kaladinlight: https://github.com/shapeshift/lib/pull/944/commits/820d2084349f986dbd9d0e0c1086e004ec68e572

packages/chain-adapters/src/types.ts
kaladinlight: Yeah if necessary. I would keep them localized to their appropriate chain scope for the chain specific versions (evm.TxMetadata, cosmos.TxMetadata). I also wonder if this would be necessary if using the discrimination string, but if creating type guards, makes sense

gomesalexandre:
Opening for discussion, see relevant web PR consuming these:
https://github.com/shapeshift/web/pull/2257

packages/chain-adapters/src/types.ts
```diff
@@ -85,54 +85,30 @@ export type SubscribeTxsInput = {
...
-  type: TradeType
-}
-
-export interface TxMetadata {
-  method?: string
-  parser: string
-}
+export type TransactionMetadataStandard = unchained.StandardTxMetadata
+export type TransactionMetadataEvm = unchained.evm.ERC20TxMetadata
+export type TransactionMetadataCosmos = unchained.cosmos.TxMetadata
```
kaladinlight, line 98: would like these ecosystem specific types to live in their corresponding `evm/types.ts` or `cosmossdk/type.ts` files that can be used/exported as needed (can add an evm and cosmossdk export alias to nest these ecosystem types under to keep nice namespacing)

kaladinlight: not sure how well I described this so happy to talk through, might be easier haha

gomesalexandre: Discussed this one out and decided to keep only one terminology i.e `TransactionMetadata` - using namespacing at import time: https://github.com/shapeshift/lib/pull/944/commits/e87cf4ca2f6ff35e11130641065a240e4e518a45 and https://github.com/shapeshift/lib/pull/944/commits/6c1d905b126eccca2791e5ada10a4462d7a86df1

packages/chain-adapters/src/types.ts
```diff
@@ -85,54 +85,28 @@ export type SubscribeTxsInput = {
...
-  chainId: string
-  confirmations: number
-  txid: string
-  fee?: TxFee
-  status: TxStatus
-  tradeDetails?: TradeDetails
+export type Transaction = Omit<unchained.StandardTx, 'transfers'> & {
   transfers: Array<TxTransfer>
-  data?: TxMetadata
+  data?: unchained.StandardTxMetadata | unchained.evm.TxMetadata | unchained.cosmos.TxMetadata
```
0xApotheosis, line 93: `unchained.StandardTxMetadata | unchained.evm.TxMetadata | unchained.cosmos.TxMetadata` could maybe be it's own type.

packages/unchained-client/src/types.ts
```diff
@@ -54,24 +54,13 @@ export enum TransferType {
...
-}
-
 export interface BaseTxMetadata {
   method?: string
   parser: string
 }
 
 export interface StandardTxMetadata extends BaseTxMetadata {
-  parser: TxParser
+  parser: 'cosmos' | 'yearn' | 'uniV2' | 'zrx' | 'thor' | 'foxy' | 'weth' | 'cowswap'
```
0xApotheosis, line 63: Removing this makes me a little sad. Would we still get the Intellisense improvements you are after if we kept the enum, but renamed it to `TxParsers`, and then created a type that is the string union of the enum values?

```ts
export enum TxParsers {
  Cosmos = 'cosmos',
  Yearn = 'yearn',
  UniV2 = 'uniV2',
  ZRX = 'zrx',
  Thor = 'thor',
  Foxy = 'foxy',
  WETH = 'weth',
  CowSwap = 'cowswap'
}

export type TxParser = keyof typeof TxParsers

export interface StandardTxMetadata extends BaseTxMetadata {
  parser: TxParser
}
```

Then we can keep using the enum in the tests and such instead of the hardcoded strings.

gomesalexandre: That works, best of both worlds!
image

gomesalexandre: Done in 042f8192406cbfb9697fa83c00f293c44831608f

packages/chain-adapters/src/types.ts
gomesalexandre: I do agree, and I was going for that initially but I couldn't find a naming that works.
We want to have a similar naming (although namespaced) for `TransactionMetadata` across all domains.

Looking back into it, I guess `TransactionData` works, happy to hear any idea you have for a better naming!

2e15a2b33a8d28a2c98f969d59669d79b2ecddd9

packages/unchained-client/src/types.ts
0xApotheosis: Perfect.

packages/chain-adapters/src/types.ts
0xApotheosis: It's not perfect, but I can't think of anything better right now!

0xApotheosis (approved):
Comments have been addressed, looking good!

PR #945, opened 2022-08-04
fix: fix bch swaps and fix detecting swaps into eth assets

elmutt:
1) fix swaps into bch. Memos dont support `bitcoincash:` prefix.

2) fix detecting swaps into eth assets. Eth txids coming back from midgard dont have the 0x prefix and need it added
Screen Shot 2022-08-03 at 6 09 33 PM

### DIFF
--- a/packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
@@ -27,11 +27,14 @@ export const makeSwapMemo = ({
     })
 
   // bch hack
-  const parsedDestAddress = destinationAddress.includes('bitcoincash:')
+  // Our bitcoin cash addresses are prefixed with `bitcoincash:`
+  // But thorchain memos need to be short (under 80 bytes for utxo)
+  // For this reason thorchain doesnt allow / need bitcoincash: in the memo
+  const parsedDestinationAddress = destinationAddress.includes('bitcoincash:')
     ? destinationAddress.replace('bitcoincash:', '')
     : destinationAddress
 
-  const memo = `s:${thorId}:${parsedDestAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
+  const memo = `s:${thorId}:${parsedDestinationAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
   if (memo.length <= MAX_LENGTH) return memo
   const abbreviationAmount = memo.length - MAX_LENGTH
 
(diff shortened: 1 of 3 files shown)


packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
```diff
@@ -26,7 +26,12 @@ export const makeSwapMemo = ({
       details: { buyAssetId }
     })
 
-  const memo = `s:${thorId}:${destinationAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
+  // bch hack
+  const parsedDestAddress = destinationAddress.includes('bitcoincash:')
```
gomesalexandre, line 30: Terminology pls 🙏🏽 
```suggestion
  const parsedDestinationAddress = destinationAddress.includes('bitcoincash:')
```

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
```diff
@@ -26,7 +26,12 @@ export const makeSwapMemo = ({
...
+    ? destinationAddress.replace('bitcoincash:', '')
+    : destinationAddress
+
+  const memo = `s:${thorId}:${parsedDestAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
```
gomesalexandre, line 34: We have a `makeSwapMemo`, can't we use that?

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
```diff
@@ -26,7 +26,12 @@ export const makeSwapMemo = ({
       details: { buyAssetId }
     })
 
-  const memo = `s:${thorId}:${destinationAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
+  // bch hack
```
gomesalexandre, line 29: Can we explain the hack, similar to what you did in the PR description body?
This doesn't say much, we're removing the `bitcoincash:` prefix but why? We should always aim for full context, the next person reading this would have no idea why this is

gomesalexandre (commented):
I wasn't able to test swap into ETH as I don't meet the minimum for swaps to it ATM, but BCH swaps worked.
Noticed the Txid link routed to `http://localhost:3000/undefined8ac0b823dbce6003bdd11fbc95c6aa02430ed9022d1044d05020214d0a1b3a6f` for swaps to BCH however, is that something to be fixed in web?

packages/swapper/src/swappers/thorchain/utils/makeSwapMemo/makeSwapMemo.ts
elmutt: I dont follow???

elmutt:
> 

ive seen that happening on some swaps. it will need to be investigated but I dont feel its related to this ticket.

I'll make ticket to look into that problem

elmutt:
> 

https://github.com/shapeshift/web/issues/2339

gomesalexandre (approved):
Approving as previously tested and this is just a var name change which doesn't make CI red

PR #940, opened 2022-08-02
feat: idle finance

samster91:
This is a draft PR, still have to fetch real opportunities from the production endpoint

PR #947, opened 2022-08-04
feat: idle finance

gomesalexandre:
Reopened from https://github.com/shapeshift/lib/pull/940 so we can iterate on an origin remote branch and release an initial version of the idle featureset.


packages/asset-service/src/generateAssetData/ethereum/idleVaults.ts
```diff
@@ -0,0 +1,70 @@
+import { ethChainId as chainId, toAssetId } from '@shapeshiftoss/caip'
+import { IdleSdk, IdleVault } from '@shapeshiftoss/investor-idle'
+import toLower from 'lodash/toLower'
+
+import { Asset } from '../../service/AssetService'
+import { ethereum } from '../baseAssets'
+import { colorMap } from '../colorMap'
+
+// const network = 1 // 1 for mainnet
+// const provider = new JsonRpcProvider(process.env.REACT_APP_ETHEREUM_NODE_URL)
```
gomesalexandre, line 10: @samster91 will we need these?

0xdef1cafe: i think this is pasta from another impl

0xdef1cafe (approved):

packages/asset-service/src/generateAssetData/ethereum/idleVaults.ts
gomesalexandre: Yeah but looks like the idle package won't need these to work, it doesn't accept these in its constructor like the yearn SDK would. Let's remove them on the next PR

PR #948, opened 2022-08-05
fix: osmo parser fixes

elmutt:
Parse osmo & cosmos tx history correctly for swaps.

To test:

make sure you are using the best cosmos/osmosis nodes, see: https://github.com/shapeshift/web/pull/2375

link locally, swap `atom -> osmo` and `osmo -> atom`. see transactions in history are for correct amounts of outgoing and incoming tokens.

NOTE ABOUT A KNOWN ISSUE:

You will not see incoming "atom" after an osmo->atom trade come over the websockets. you will only see this tx after reloading the page. This is a known issue: https://github.com/shapeshift/lib/issues/961

Screenshot below demonstrates  `atom -> osmo` followed by `osmo -> atom` (read from the bottom up)

You will notice transactions with no amounts, these are the `ibc atom on osmo` tokens being sent and received ephemerally during the trade

Screen Shot 2022-08-05 at 8 44 18 AM


packages/unchained-client/src/cosmos/parser/utils.ts
```diff
@@ -108,6 +115,18 @@ const virtualMessageFromEvents = (
       to: msg.to,
       origin: msg.origin
     }
+  } else if (swapEventData) {
+    const sender = swapEventData?.attributes.find((attribute) => attribute.key === 'sender')?.value
+    const swapAmount = swapEventData?.attributes.find(
```
0xApotheosis, line 120: We know `swapEventData` exists here, so we can kill the optional chaining.

gomesalexandre:
@elmutt Can we get more descriptive in what exactly this fixes i.e how broken is "broken" now and what should we expect to see when testing this?

elmutt:
> @elmutt Can we get more descriptive in what exactly this fixes i.e how broken is "broken" now and what should we expect to see when testing this?

Added testing instructions.

packages/unchained-client/src/cosmos/parser/utils.ts
```diff
@@ -108,6 +120,18 @@ const virtualMessageFromEvents = (
...
     }
+  } else if (swapEventData) {
+    const sender = swapEventData.attributes.find((attribute) => attribute.key === 'sender')?.value
+    const swapAmount = swapEventData.attributes.find(
+      (attribute) => attribute.key === 'tokens_out'
+    )?.value
+    const valueParsed = swapAmount?.slice(0, swapAmount?.length - 'uosmo'.length) ?? ''
+    if (swapAmount?.includes('uosmo'))
+      return {
+        type: 'send',
```
gomesalexandre, line 131: Don't we want to be able to parse Osmo swaps as `TradeType` Txs?
https://github.com/shapeshift/lib/blob/944c8d12dc8380734fe197a8c7555c9c1ce903f7/packages/unchained-client/src/types.ts#L35-L37

gomesalexandre (commented):
Question

packages/unchained-client/src/cosmos/parser/utils.ts
elmutt: Im not sure, maybe? The main goal here was to make sure the sends / receives totals of the swap txs are "correct" so the charts are correct

For thorchain we aren't showing them as "trades" either -- they just show up as sends and receives on different chains. It kindof makes sense to me to keep following that same pattern and not worry about displaying any one of the transactions as a "Trade". In my opinion the idea of a "trade" transaction in the parser feels like it was only originally planned for 0x swapper without considering how the others will work

We have talked about the idea of a "blockchain activity" or something more all-encompassing to display trade history. My 2 cents is this is fine until we have a better "trade history" solution for all swappers

kaladinlight: They could still be marked as trades with applicable metadata, but yes, the concept of a trade occurring across multiple transactions needs some additional thought on how to tie these together and if it can/should be done in the parser

gomesalexandre: Yeah @kaladinlight  as we've discussed with @elmutt I definitely agree this is not a task for now - happy with merging this if it fixes what it's supposed to, unfortunately I didn't get yet the Tx history to show all the Txs it should - not sure yet whether it's a regression from prod or a ws issue

kaladinlight:
Question, if the ibc events during trade are ephemeral, could/should we omit those entirely as it is a wash and just clutters up the tx history? I suppose it is technically a transaction that occurs for the user, but I don't know how useful it is to display? 

@shapeshift/product

packages/unchained-client/src/cosmos/parser/utils.ts
kaladinlight: @gomesalexandre sounds good and the cosmos stuff is the last main abstraction update that we need to do and I can take some time thinking through this more when we get to that for the parser

DiggyDiggy2:
I do think it's important that we display an "IBC send" if possible. We can make a separate ticket for it, because I dont consider it a blocker. Just thinking in the case that it gets stuck on the IBC channel, without the trade executing (for some reason) the user can see where their funds are.

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
```diff
@@ -274,6 +274,9 @@ export class OsmosisSwapper implements Swapper<ChainId> {
         })
 
       ibcSellAmount = await pollForAtomChannelBalance(receiveAddress, this.deps.osmoUrl)
+      // delay to ensure all nodes we interact with are up to date at this point
+      // seeing intermittent bugs that suggest the balances and sequence numbers were sometimes off
+      await new Promise((resolve) => setTimeout(resolve, 3000))
```
0xApotheosis, line 279: This feels a little hacky - is there a way we can instead check the response from the node to see if we need to retry it for fresher data?

0xApotheosis (commented):
Tested locally and this does what is says it does.

The issue with requiring enough balance of the receive asset likely needs to be solved before we merge.

I'm not sure of the longer-term solution, but in the short-term we can block the user from performing the trade if the receive balance is insufficient. This should only be a couple of lines to add.

Pending @shapeshift/product's thoughts on this.

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
elmutt: Not that I know of easily. Its possible we no longer need this now that we are pointing at datahub nodes (much more reliable) but I would feel safer with them in there because of some bizarre behavior we have seen

elmutt:
> Tested locally and this does what is says it does.
> 
> The issue with requiring enough balance of the receive asset likely needs to be solved before we merge.
> 
> I'm not sure of the longer-term solution, but in the short-term we can block the user from performing the trade if the receive balance is insufficient. This should only be a couple of lines to add.
> 
> Pending @shapeshift/product's thoughts on this.

I don't think we should need to hold up merging this PR waiting on an answer for the zero balance thing. We just cant release to the public without it but its not really related to this PR

0xApotheosis:
> I don't think we should need to hold up merging this PR waiting on an answer for the zero balance thing. We just cant release to the public without it but its not really related to this PR

Yep, as it's still behind a flag I'm ok with that.

Just spin up a ticket for it before this gets merged so we can track the issue.

elmutt:
> > I don't think we should need to hold up merging this PR waiting on an answer for the zero balance thing. We just cant release to the public without it but its not really related to this PR
> 
> Yep, as it's still behind a flag I'm ok with that.
> 
> Just spin up a ticket for it before this gets merged so we can track the issue.

https://github.com/shapeshift/web/issues/2377

gomesalexandre (changes requested):
Tested locally with a wallet that has enough OSMO for ATOM->OSMO swap on the last leg of the Txs (OSMO IBC channel):

Swap:

image

Tx history:

image

This doesn't look right to me. The latest Tx (OSMO receive) is for a previous swap, see last row item. 

Here are the two Txs from the last ATOM->OSMO swap:

https://www.mintscan.io/cosmos/txs/B2215B1885C97A1A048B91C80F8A7EEFD506C459C4E266E469B576C744FFB6B2
https://www.mintscan.io/osmosis/txs/87DBBE6D5BA24BD659D85C6DEBA740A310234707203D40B68E3F6534CEE6FD14

gomesalexandre:
Tested again as a sanity check:

- Swap confirmation Tx

https://www.mintscan.io/osmosis/txs/96384380E94EA8B70CE00FDB801C7B62100D97C083BF1CA12A76E85CAEF65BAF

- Tx history
image

This looks good. It seems that when there is a Tx failure, our Tx history gets shuffled and sets the previous successful swap Tx as the very last Tx (the first in the list from the top).

I'm conceptually ok with this PR as this fixes what it is supposed to fix, but after you confirm that Tx history shuffle bug  @elmutt, we should create an issue for it

elmutt:
> Tested again as a sanity check:
> 
> * Swap confirmation Tx
> 
> https://www.mintscan.io/osmosis/txs/96384380E94EA8B70CE00FDB801C7B62100D97C083BF1CA12A76E85CAEF65BAF
> 
> * Tx history
> 
> image
> 
> This looks good. It seems that when there is a Tx failure, our Tx history gets shuffled and sets the previous successful swap Tx as the very last Tx (the first in the list from the top).
> 
> I'm conceptually ok with this PR as this fixes what it is supposed to fix, but after you confirm that Tx history shuffle bug @elmutt, we should create an issue for it. Ill stamp this after.

I haven't seen that exact issue manifest before and im not sure how these parser changes would result in them being out of order. Maybe it was a one-time fluke with unchained returning txs?

The only bug I'm currently aware of is during an osmo-> atom swap the atom tx will not appear until the page is refreshed, this issue:

https://github.com/shapeshift/lib/issues/961

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
gomesalexandre: On a different note, might be time to switch to node 16 to benefit from `timers/promises` so we can `await setTimeout()` directly, one less method pushed to the next tick.
https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options

gomesalexandre:
Retested lib built and linked with the latest changes and the parsing still doesn't look right:

- ATOM -> OSMO swap

image

IBC received Tx: https://www.mintscan.io/cosmos/txs/728B14C9E4469B1A880AD6C6406BF8E6C8B82D906EC960EB0A6CBD558F87BDDF

Looks like this is wrongly parsing this intermediary Tx as the receive Tx:

https://www.mintscan.io/cosmos/txs/728B14C9E4469B1A880AD6C6406BF8E6C8B82D906EC960EB0A6CBD558F87BDDF

- OSMO -> ATOM swap looking good:

image

IBC swap exact amount Tx: https://www.mintscan.io/osmosis/txs/493E685C7C2E8E4817CF8271789E9901CF1C4506286E2386B00DFA1F17510B50

packages/unchained-client/src/cosmos/parser/utils.ts
```diff
@@ -123,9 +168,10 @@ const virtualMessageFromEvents = (
 export const valuesFromMsgEvents = (
   msg: Message,
   events: { [key: string]: Event[] },
-  assetId: string
+  assetId: string,
```
gomesalexandre, line 171:
```suggestion
  assetId: AssetId,
```

gomesalexandre (approved):
> Retested lib built and linked with the latest changes and the parsing still doesn't look right

This was a one-off issue where the IBC channel reconciliated a previously failed trade, with IBC ATOM still remaining that got sent with the next Tx. This isn't a bug with our Tx history parsing.

Approved with comments, since this functionally works, tested

- ATOM -> OSMO
- OSMO -> ATOM

and saw the correct values for both after redoing the swaps multiple times to confirm the behavior:

image

PR #950, opened 2022-08-07
fix(Transaction type): tradeDetails -> trade

gomesalexandre:
Makes Tx history great again after types refactor in https://github.com/shapeshift/lib/commit/944c8d12dc8380734fe197a8c7555c9c1ce903f7

Before:

image

After:

image


0xApotheosis (approved):
Tested locally, this does the thing it says it does:

Screen Shot 2022-08-08 at 6 38 14 am

PR #954, opened 2022-08-08
feat(swapper): add isCowTrade typeguard

gomesalexandre:
Description

This adds an `isCowTrade` type guard, since we need to consume the feeAmountInSellToken property which is specific to CowTrade.

Issue

Contributes to https://github.com/shapeshift/web/issues/2271

### DIFF
--- a/packages/swapper/src/swappers/cow/utils.ts
+++ b/packages/swapper/src/swappers/cow/utils.ts
@@ -1,0 +1,8 @@
+import { ChainId } from '@shapeshiftoss/caip'
+
+import { Trade } from '../../api'
+import { CowTrade } from './types'
+
+export const isCowTrade = <C extends ChainId>(
+  trade: CowTrade<C> | Trade<C>
+): trade is CowTrade<C> => Boolean((trade as CowTrade<C>).feeAmountInSellToken)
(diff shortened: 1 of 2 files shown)


packages/swapper/src/swappers/cow/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { ChainId } from '@shapeshiftoss/caip'
+
+import { Trade } from '../../api'
+import { CowTrade } from './types'
+
+export const isCowTrade = <C extends ChainId>(
+  trade: CowTrade<C> | Trade<C>
+): trade is CowTrade<C> => Boolean((trade as CowTrade<C>).feeAmountInSellToken)
```
0xdef1cafe, line 8:
```suggestion
): trade is CowTrade<C> => Boolean((trade as CowTrade<C>)?.feeAmountInSellToken)
```

without pulling locally it feeeeels like we'll need opt chain here?

kaladinlight: also curious if the `in` syntax works.

`'feeAmountInSellToken' in trade`

0xdef1cafe (approved):

PR #962, opened 2022-08-10
feat(swapper): add approveAmount swapper API method

0xApotheosis:
Draft PR for @NeOMakinG to take over and finish https://github.com/shapeshift/web/issues/2263

Web implementation: https://github.com/shapeshift/web/pull/2419


packages/swapper/src/api.ts
```diff
@@ -189,7 +194,7 @@ export enum SwapErrorTypes {
 }
 
 export interface Swapper<T extends ChainId> {
-  /** Human readable swapper name */
+  /** Human-readable swapper name */
```
gomesalexandre, line 197: ❤️

packages/swapper/src/api.ts
```diff
@@ -129,6 +129,10 @@ export type ApproveInfiniteInput<C extends ChainId> = {
   wallet: HDWallet
 }
 
+export type ApproveAmountInput<C extends ChainId> = ApproveInfiniteInput<C> & {
```
gomesalexandre, line 132: Rename `ApproveInfiniteInput` to `ApproveInput` and make `ApproveInfiniteInput` an alias of the latter?

packages/swapper/src/swappers/cow/CowSwapper.ts
```diff
@@ -76,6 +78,12 @@ export class CowSwapper implements Swapper<KnownChainIds.EthereumMainnet> {
     return cowApproveInfinite(this.deps, args)
   }
 
+  async approveAmount(): Promise<string> {
+    throw new SwapError('CowSwapper: approveAmount unimplemented', {
+      code: SwapErrorTypes.RESPONSE_ERROR
```
gomesalexandre, line 83: Not sure that's the best error code semantically, but I don't believe adding a code for unimplemented methods is a good solution either so let's go with this!

packages/swapper/src/swappers/cow/CowSwapper.ts
```diff
@@ -76,6 +78,12 @@ export class CowSwapper implements Swapper<KnownChainIds.EthereumMainnet> {
     return cowApproveInfinite(this.deps, args)
   }
 
+  async approveAmount(): Promise<string> {
```
gomesalexandre, line 81: - This never returns a string on the unimplemented implementations.
- The implemented ones are also not *guaranteed* to return a string, they can throw and thus return `void`.
Let's change the type definition of the implemented ones to `string | void`, and the one from the unimplemented ones to `void`?

packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
```diff
@@ -1,37 +1,67 @@
...
+  return await grantAllowance<T>({
+    quote: approvalQuote,
+    wallet,
+    adapter,
+    erc20Abi,
+    web3
+  })
+}
+
+export async function zrxApproveAmount<T extends EvmSupportedChainIds>(
```
gomesalexandre, line 41: This module is named `zrxApproveInfinite`, it should either be renamed, or this should be in a whole new module (preference for the former vs. having to search through many different modules, this should be colocated IMO)

gomesalexandre (commented):
A few initial comments but overall looks good

packages/swapper/src/swappers/cow/CowSwapper.ts
kaladinlight: More accurately, this function would be assigned a return type of never (which does satisfy the swapper interface), though we don't really interact with specific swapper types in the client.

Considering this is an async function that should be wrapped in a try/catch, I don't see too big of a problem with this current implementation and it follows how we have handled unimplemented function up until this point as far as I am aware.

packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
kaladinlight: or even better, wax all of these nested directories and just have the swapper class with the logic contained within there :) not a fan of this bloated setup personally. obv not for this PR.

I would go with rename as well if we want to change at this time

gomesalexandre (commented):
Tested locally while linked in the matching web PR and this does what it says on the box.

packages/swapper/src/swappers/zrx/zrxApproveInfinite/zrxApproveInfinite.ts
NeOMakinG: Agree, and so how would we call the module containing these 2 functions? `srxApprove`?

gomesalexandre: Precisely @NeOMakinG, that's what I had in mind

gomesalexandre (approved):
Looks good now, comments have been tackled.

Gave it a sanity check while linked to the web PR and still works as expected.

Note that until implemented, Thor/CoW exact allowances in the trade component will fail
Given:
1. This wasn't in the spec
2. Those two are under feature flag

I'm fine with getting this one (and web) in, and tackling Thor and CoW as follow-ups.

Well done ser 🍬

PR #964, opened 2022-08-11
feat(swapper): populate CowSwap `sellTxid`

0xApotheosis:
Fixes an issue where we are not passing a `sellTxid` to `web` for CowSwap trades when `cowGetTradeTxs()` is called, which is causing an exception and preventing the trade confirm screen from being shown once the transaction is complete.

image

Closes https://github.com/shapeshift/web/issues/2399


gomesalexandre (approved):
Tested linked with https://github.com/shapeshift/web/pull/2401 and the right txid is populated, allowing us to link to the Cow.fi explorer

image

PR #969, opened 2022-08-12
fix(swapper): thorchainSwapper quote inversion

0xApotheosis:
`ThorSwapper`'s `getPriceRatio` is returning an inverted response, which is called when a `sellAmount` of `0` is quoted for.

We then invert it again here:

https://github.com/shapeshift/lib/blob/d3002a6ff2c5a220bf79aa8df08e03d581da2e32/packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts#L62

This hides the `getPriceRatio` bug when the `sellAmount` is `0`, e.g. for the initial quote, but then causes an inverted rate when `sellAmount` is _not_ `0`.

**Before** 😕 

Screen Shot 2022-08-12 at 3 35 10 pm

**After** 😌 

Screen Shot 2022-08-12 at 3 35 10 pm

Closes https://github.com/shapeshift/web/issues/2417


gomesalexandre (approved):
Tested against lib main, this fixes the quote inversion and I didn't find any regression in quotes.

🥑

PR #971, opened 2022-08-13
feat(unchained-client): parse approval amount

gomesalexandre:
### Description

This parses the _amount argument of approve() calls as tx.data.parser to be consumed in web

### Issue

Relates to https://github.com/shapeshift/web/issues/2345


packages/unchained-client/src/evm/parser/erc20.ts
```diff
@@ -45,7 +55,8 @@ export class Parser<T extends Tx> implements SubParser<T> {
           assetReference: tx.to
         }),
         method: decoded.name,
-        parser: TxParser.ERC20
+        parser: TxParser.ERC20,
+        ...(approveValue ? { value: approveValue } : {})
```
gomesalexandre, line 59: Notes:
- Not sure if we should stick to the actual argument name e.g `value` for yearn, `amount` for ERC20s, matching the `_amount` arg
- Not sure if this is the best place / types for this? We might have other functions we support that might take an amount parameter, or other parameters from other functions we might want to parse as well. Maybe `args` or `arguments` as a `Record<T, U>` might be a better fit?

packages/unchained-client/src/evm/parser/erc20.ts
```diff
@@ -34,6 +34,16 @@ export class Parser<T extends Tx> implements SubParser<T> {
 
     const decoded = this.abiInterface.parseTransaction({ data: tx.inputData })
 
+    let approveValue: string | null = null
+
+    switch (txSigHash) {
+      case this.supportedFunctions.approveSigHash:
+        approveValue = decoded?.args.amount?.toString()
```
gomesalexandre, line 41: Note: ERC20 parsing works for ZRX, CowSwap, FOXy staking and UniV2 LPs. From the testing I've done, only Yearn requires separate parsing.

packages/unchained-client/src/evm/ethereum/parser/yearn.ts
```diff
@@ -61,9 +64,18 @@ export class Parser implements SubParser<Tx> {
...
+
     switch (txSigHash) {
       case this.supportedYearnFunctions.approveSigHash:
         if (decoded?.args._spender !== SHAPE_SHIFT_ROUTER_CONTRACT) return
+        approveValue = decoded?.args._value?.toString()
+        assetId = toAssetId({
+          ...fromChainId(this.chainId),
+          assetNamespace: 'erc20',
+          assetReference: tx.to
+        })
```
gomesalexandre, line 78: Not entirely sure about this one either, but IMO, better to only set `assetId` for `approve()`s.
Setting it for other calls could produce unwanted results like having the contract we're interacting to as an `AssetId` e.g ShapeShift DAO router) and would be a pain to actually parse what's an actual token contract being interacted with as an `approve()` and what's not in web

packages/unchained-client/src/types.ts
```diff
@@ -72,6 +74,8 @@ export interface BaseTxMetadata {
 
 export interface StandardTxMetadata extends BaseTxMetadata {
   parser: `${TxParser}`
+  value?: string
+  assetId?: AssetId
```
kaladinlight, line 78: we should not be polluting the standard tx metadata with any parser specific data
```suggestion
```

packages/unchained-client/src/evm/parser/erc20.ts
kaladinlight: IIFE to get rid of let and ternary below

packages/unchained-client/src/evm/ethereum/parser/yearn.ts
```diff
@@ -61,9 +64,18 @@ export class Parser implements SubParser<Tx> {
     // failed to decode input data
     if (!decoded) return
 
+    let approveValue: string | null = null
+    let assetId: string | null = null
```
kaladinlight, line 68: IFFE here as well to remove let and simplify return payload

packages/unchained-client/src/evm/parser/types.ts
```diff
@@ -11,6 +11,7 @@ export enum TxParser {
 export interface ERC20TxMetadata extends BaseTxMetadata {
   parser: TxParser.ERC20
   assetId?: string
+  value?: string
```
kaladinlight, line 14: Coincidentally this metadata now resembles both yearn and erc20. I think we have a couple options:

- Make this a less strict `EvmTxMetadata` type that all evm parsers will use and we can then narrow by `isEvmParser` or something along those lines
- Only move `TxParser.Yearn` into here and figure out a reasonable name for the type to keep things more strict/accurate
- Update to have each parser define a `TxMetadata` type (either alias to standard or some specific fields), and union all parser metadatas together under the `evm.TxMetadata` umbrella

I feel like the 3rd option would be the most explicit and flexible pattern moving forward assuming all of the narrowing still works as intended, though there will be slight extra overhead keeping the union type up to date.

packages/unchained-client/src/evm/parser/erc20.ts
kaladinlight: I say stick with `value` naming just so you can expect to see the similar data across the whole parsed tx payload imo.

A record would certainly be easier, but it would make consumption difficult client side as you have no idea what values could be in there. Depending on how complex these get, there is always the possibility of adding a contractArgs object that could be defined based on the contract method as a discriminator, but that is getting really close to just turning into ethersjs/web3 abi parsing at that point. Could be useful to pass back the `inputData` as an alternative for more custom needs as well

packages/unchained-client/src/evm/parser/types.ts
gomesalexandre: I would be more inclined towards the first solution given the overhead of the last, but agree the third one is more accurate and should allow for better discrimination, will give it a try!

gomesalexandre: Done in 598f7736693cc936f03265337657e98177f7fb0b - gave it a try while linked in web and `isTokenMetadata` still works and properly narrows TxMetadata type down with the new union :tada:

packages/unchained-client/src/evm/ethereum/parser/yearn.ts
gomesalexandre: Done in 338d2e7a9e8a1d522906358acb0913dae03ac334 / f91cd8c179d2b2ef0ca004c96bc16b9f18332464

packages/unchained-client/src/evm/parser/erc20.ts
gomesalexandre: Done in 338d2e7a9e8a1d522906358acb0913dae03ac334 / f91cd8c179d2b2ef0ca004c96bc16b9f18332464

kaladinlight (approved):

gomesalexandre (commented):
Can't approve my own PR but tested latest changes from https://github.com/shapeshift/web/pull/2460 (merged into this PR) while linked in web PR https://github.com/shapeshift/web/pull/2435 and noticed no regressions

PR #974, opened 2022-08-15
chore: set up git-blame-ignore-revs

0xApotheosis:
This PR sets up a `.git-blame-ignore-revs` so that formatting commits are ignored in git blame on both GibHub, locally, and in IDEs.

We already have this set up in `web`, introduced here: https://github.com/shapeshift/web/pull/1438

- A `.git-blame-ignore-revs` file has been implemented, which allows us to prevent lint commits showing up when using git blame on GitHub. 
- A `.gitconfig` file is added which allows IDEs (tested on WebStorm and VSCode) to respect all skipped commits
- `postinstall.sh` has been updated to automatically use the `.gitconfig` locally once `yarn` / `yarn install` is run

It also works locally by using `git blame --ignore-revs-file .git-blame-ignore-revs`.

For more information on this feature, please see here: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


gomesalexandre (approved):
LGTM, added the trailing commas lint hash to `.git-blame-ignore-revs` in 999a58fdb53d26e335d328b7f1a769c343a437bb:

- Good ol' `git blame` with no flags:

image

- `git blame` with `--ignore-revs-file` flag:

image

Note that vscode gitlens users can automagically have hyper-blame by configuring it like so:

https://github.com/gitkraken/vscode-gitlens/issues/413#issuecomment-523640184

Alternatively, or if not using VSCode, setting the `blame.ignoreRevsFile` option globally will automagically apply the `--ignore-revs-file`:

`git config --global blame.ignoreRevsFile .git-blame-ignore-revs`

PR #980, opened 2022-08-16
feat: parse approve amount

kaladinlight:


gomesalexandre (approved):
Tested the new types while linking lib in the web PR, LGTM
https://github.com/shapeshift/web/pull/2460

PR #985, opened 2022-08-23
feat: asset service overrides

0xdef1cafe:
we want/need a way to be able to override asset service data coming from third party sources

this allows us to override specific fields in a json object, e.g. color/icon/name

additionally this uses the correct icon and color for FOX - the trustwallet data upstream is wrong


packages/asset-service/src/generateAssetData/generateAssetData.ts
```diff
@@ -1,17 +1,24 @@
...
 import { atom, bitcoin, bitcoincash, dogecoin, litecoin } from './baseAssets'
 import * as ethereum from './ethereum'
 import * as osmosis from './osmosis'
+import assetOverrides from './overrides.json'
 import { setColors } from './setColors'
 import { filterOutBlacklistedAssets } from './utils'
 
+export type AssetOverrides = Record<
+  AssetId,
+  Partial<Omit<Asset, 'assetId' | 'chainId' | 'precision'>> // fields that can't be overridden
```
gomesalexandre, line 19: 💜

gomesalexandre (approved):
Tested with `asset-service` linked locally, does what it says on the label

image

PR #989, opened 2022-08-25
feat(swapper): serialize `GetTradeQuoteInput` type args

0xApotheosis:
Serialize `GetTradeQuoteInput` type args so we can use `getTradeQuote` with RTK query, which only supports serializable arguments.

A `HDWallet` object isn't actually needed.

This is a breaking change for `@shapeshiftoss/swapper`.

Follow-up `web` PR to support the breaking change: https://github.com/shapeshift/web/pull/2555

Contributes to https://github.com/shapeshift/web/issues/2556

### DIFF
--- a/packages/swapper/src/api.ts
+++ b/packages/swapper/src/api.ts
@@ -69,7 +69,6 @@ type CommonTradeInput = {
   sellAmount: string
   sendMax: boolean
   sellAssetAccountNumber: number
-  wallet: HDWallet
   receiveAddress: string
 }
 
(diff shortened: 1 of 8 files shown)


0xApotheosis:
Note for @gomesalexandre as you'll probably look at this when you wake up - my local linking isn't working at the moment, so I haven't been able to actually test this yet, though the logic change is pretty straightforward.

gomesalexandre (approved):
Tested while linked in web, didn't notice any regressions - tested with an UTXO (DOGE  -> LTC) swap, and went with the full trade as a sanity check.

PR #990, opened 2022-08-25
fix: osmo: calculate finish tx off second tx in transaction

toshiSat:
Fixes not going to the confirm screen during a swap by sending both sell and buy asset tx info.
Also bumped timeout delays so we can register ibc transactions correctly


elmutt (dismissed):

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
```diff
@@ -309,6 +333,7 @@ export class OsmosisSwapper implements Swapper<ChainId> {
 
     const signed = await osmosisAdapter.signTransaction(signTxInput)
     const tradeId = await osmosisAdapter.broadcastTransaction(signed)
+    const cosmosTxHistory = await cosmosAdapter.getTxHistory({ pubkey: cosmosAddress })
```
kaladinlight, line 336: we can add `pageSize: 1` to limit this request, and we should also probably fetch it just prior to broadcast to be extra safe.

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
```diff
@@ -57,10 +57,31 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
+        | cosmos.ChainAdapter
+        | undefined
+
+      if (!cosmosAdapter)
+        throw new SwapError('OsmosisSwapper: couldnt get cosmos adapter', {
+          code: SwapErrorTypes.GET_TRADE_TXS_FAILED,
+        })
+
+      const cosmosTxHistory = await cosmosAdapter.getTxHistory({
+        pubkey: tradeResult.cosmosAddress,
```
kaladinlight, line 72: `pageSize: 1` here as well

packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
```diff
@@ -57,10 +57,31 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
+      })
+      const currentCosmosTxid = cosmosTxHistory?.transactions[0].txid
+
+      return {
+        sellTxid: tradeResult.tradeId,
+        buyTxid: currentCosmosTxid !== tradeResult.previousCosmosTxid ? currentCosmosTxid : '',
```
kaladinlight, line 78: is there any metadata that exists on the tx that can be use to check that it was created as a result of the sell tx? odds are low, but this could point to the incorrect tx if there was a send/receive before buy tx was sent

elmutt: I agree its hacky and it sucks. But I couldnt find a way to do it otherwise and this seemed like the most reasonable fast solution to work in 99% of cases

kaladinlight: can you drop a comment to state the edge case so we can re-evaluate later if there is a way please

packages/swapper/src/api.ts
```diff
@@ -141,6 +141,11 @@ export type TradeResult = {
   tradeId: string
 }
 
+export interface OsmosisTradeResult extends TradeResult {
```
kaladinlight, line 144: Declare type alongside Osmosis swapper and export from there

kaladinlight (approved):

PR #991, opened 2022-08-25
feat: pass accountId back with websocket txs

0xdef1cafe:
a backwards compatible change that allows us to greatly simplify tx subscription logic in web

by passing back the accountId with a transaction message, web is able to determine which account it
belongs to and index it correctly in the store

this also allows use to delete a quarter of the portfolioSlice which maintained a mapping between
accountIds -> address[], which solely existed to be able to do this lookup


0xdef1cafe:
closing after discussion with @kaladinlight - we have a better approach such that the restful `getTxHistory` and `subscribeTxs` return the same `Transaction` interface

PR #1011, opened 2022-09-02
refactor(caip): remove named capture groups in parseAssetId regular expression

pastaghost:
### Description
Non-breaking changes.
This removes the use of named capture groups in parseAssetIdRegExp to maintain compatibility with the Secure ECMAScript runtime.

### DIFF
--- a/packages/caip/src/assetId/assetId.ts
+++ b/packages/caip/src/assetId/assetId.ts
@@ -10,6 +10,7 @@ import {
 } from '../constants'
 import {
   assertIsAssetNamespace,
+  assertIsAssetReference,
   assertIsChainNamespace,
   assertIsChainReference,
   assertValidChainPartsPair,
@@ -132,23 +133,21 @@ export const fromAssetId: FromAssetId = (assetId) => {
   const matches = parseAssetIdRegExp.exec(assetId)
   if (!matches) throw new Error(`fromAssetId: could not parse AssetId: ${assetId}`)
 
-  // These should never throw because isAssetId() would have already caught it, but they help with type inference
-  assertIsChainNamespace(matches[ASSET_ID_CAPTURE_GROUPS.CHAIN_NAMESPACE_CAPTURE_GROUP])
-  assertIsChainReference(matches[ASSET_ID_CAPTURE_GROUPS.CHAIN_REFERENCE_CAPTURE_GROUP])
-  assertIsAssetNamespace(matches[ASSET_ID_CAPTURE_GROUPS.ASSET_NAMESPACE_CAPTURE_GROUP])
-
   const chainNamespace = matches[ASSET_ID_CAPTURE_GROUPS.CHAIN_NAMESPACE_CAPTURE_GROUP]
   const chainReference = matches[ASSET_ID_CAPTURE_GROUPS.CHAIN_REFERENCE_CAPTURE_GROUP]
   const assetNamespace = matches[ASSET_ID_CAPTURE_GROUPS.ASSET_NAMESPACE_CAPTURE_GROUP]
-
   const shouldLowercaseAssetReference =
     assetNamespace && ['erc20', 'erc721'].includes(assetNamespace)
-
   const assetReference = shouldLowercaseAssetReference
     ? toLower(matches[ASSET_ID_CAPTURE_GROUPS.ASSET_REFERENCE_CAPTURE_GROUP])
     : matches[ASSET_ID_CAPTURE_GROUPS.ASSET_REFERENCE_CAPTURE_GROUP]
-  assertIsChainReference(chainReference)
+
+  // These should never throw because isAssetId() would have already caught it, but they help with type inference
   assertIsChainNamespace(chainNamespace)
+  assertIsChainReference(chainReference)
+  assertIsAssetNamespace(assetNamespace)
+  assertIsAssetReference(assetReference)
+
   const chainId = toChainId({ chainNamespace, chainReference })
 
   if (assetNamespace && assetReference && chainId) {
--- a/packages/caip/src/assetId/assetId.ts
+++ b/packages/caip/src/assetId/assetId.ts
@@ -126,23 +126,26 @@ export const fromAssetId: FromAssetId = (assetId) => {
   const matches = parseAssetIdRegExp.exec(assetId)
   if (!matches) throw new Error(`fromAssetId: could not parse AssetId: ${assetId}`)
 
-  const [, chainNamespace, chainReference, assetNamespace] = matches
-  let [, , , , assetReference] = matches
+  const { 1: chainNamespace, 2: chainReference, 3: assetNamespace, 4: assetReference } = matches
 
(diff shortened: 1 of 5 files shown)


0xdef1cafe:
@pastaghost please fix failing tests

packages/caip/src/assetId/assetId.ts
```diff
@@ -124,26 +130,24 @@ export type FromAssetId = (assetId: AssetId) => FromAssetIdReturn
...
-  // These should never throw because isAssetId() would have already caught it, but they help with type inference
-  assertIsChainNamespace(matches.chainNamespace)
-  assertIsChainReference(matches.chainReference)
-  assertIsAssetNamespace(matches.assetNamespace)
-
-  const chainNamespace = matches.chainNamespace
-  const chainReference = matches.chainReference
-  const assetNamespace = matches.assetNamespace
-
+  const chainNamespace = matches[ASSET_ID_CAPTURE_GROUPS.CHAIN_NAMESPACE_CAPTURE_GROUP]
```
0xdef1cafe, line 136: @pastaghost this might be helpful

named captured groups were just a nice developer experience improvement, and not really essential for this regex to perform its intended purpose

we can simply destructure the same data out as an array 

`const [chainNamespace, chainReference, assetNamespace, assetReference] = matches`

your regex behaves exactly the same

https://regex101.com/r/oGdXUs/1

0xApotheosis:
I suspect we can also revert the changes made to `ASSET_NAMESPACE_STRINGS` since we found the real issue with type narrowing, which was the missing `assertIsAssetReference` type guard assertion.

packages/caip/src/typeGuards.ts
```diff
@@ -21,20 +16,21 @@ export const isChainReference = (
...
 
 export const isAssetReference = (
   maybeAssetReference: AssetReference | string,
 ): maybeAssetReference is AssetReference =>
   Object.values(ASSET_REFERENCE).some((s) => s === maybeAssetReference)
 
 export const isAssetId = (maybeAssetId: AssetId | string): maybeAssetId is AssetReference => {
-  const matches = parseAssetIdRegExp.exec(maybeAssetId)?.groups
+  const matches = parseAssetIdRegExp.exec(maybeAssetId)
+  const [, chainNamespace, chainReference, assetNamespace] = matches || []
```
0xApotheosis, line 28: We also can't do `|| []` because the type guards and associated type guard assertions do not handle `undefined` values, which the destructured values would be if it was indeed an empty array.

TypeScript misses this when using an array constant on the right-hand side of a logical or operator, though you can confirm the downstream effects if you, say, try passing `undefined` to `isChainNamespace` on line 31.

One option is to replace line 28 with the following code that includes an early exit, as `matches ` will either be a valid array, or `null`:

```ts
if (!matches) return false
const [, chainNamespace, chainReference, assetNamespace] = matches
```

Then remove the truthy check for `matches` on line 30, as by this stage we know it's truthy.

We can also tighten up the array destructuring by using object destructuring syntax on the array's numeric keys:

```ts
const { 1: chainNamespace, 2: chainReference, 3: assetNamespace } = matches
```

packages/caip/src/constants.ts
```diff
@@ -79,28 +100,13 @@ type ValidAssetNamespace = {
...
-  [CHAIN_NAMESPACE.CosmosSdk]: ['cw20', 'cw721', 'ibc', 'native', 'slip44'],
+  [CHAIN_NAMESPACE.Utxo]: [ASSET_NAMESPACE.slip44],
+  [CHAIN_NAMESPACE.Evm]: [ASSET_NAMESPACE.slip44, ASSET_NAMESPACE.erc20, ASSET_NAMESPACE.erc721],
+  [CHAIN_NAMESPACE.CosmosSdk]: [
+    ASSET_NAMESPACE.cw20,
+    ASSET_NAMESPACE.cw721,
+    ASSET_NAMESPACE.ibc,
+    ASSET_NAMESPACE.native,
+    ASSET_NAMESPACE.slip44,
+  ],
```
0xApotheosis, line 111: Nice 👌

packages/caip/src/assetId/assetId.ts
```diff
@@ -124,26 +123,22 @@ export type FromAssetId = (assetId: AssetId) => FromAssetIdReturn
...
   const shouldLowercaseAssetReference =
     assetNamespace && ['erc20', 'erc721'].includes(assetNamespace)
 
-  const assetReference = shouldLowercaseAssetReference
-    ? toLower(matches.assetReference)
-    : matches.assetReference
-  assertIsChainReference(chainReference)
-  assertIsChainNamespace(chainNamespace)
+  assetReference = shouldLowercaseAssetReference ? assetReference.toLowerCase() : assetReference
+
```
0xApotheosis, line 141: We can remove this once we use `const` and set in-line.

packages/caip/src/utils.ts
```diff
@@ -4,7 +4,7 @@ import * as constants from './constants'
 
 // https://regex101.com/r/f0xGqP/2
```
0xApotheosis, line 5: Update this, as we've updated the corresponding regex.

0xApotheosis (changes requested):

packages/caip/src/assetId/assetId.ts
```diff
@@ -124,26 +123,22 @@ export type FromAssetId = (assetId: AssetId) => FromAssetIdReturn
...
-  const matches = parseAssetIdRegExp.exec(assetId)?.groups
+  const matches = parseAssetIdRegExp.exec(assetId)
   if (!matches) throw new Error(`fromAssetId: could not parse AssetId: ${assetId}`)
 
-  // These should never throw because isAssetId() would have already caught it, but they help with type inference
-  assertIsChainNamespace(matches.chainNamespace)
-  assertIsChainReference(matches.chainReference)
-  assertIsAssetNamespace(matches.assetNamespace)
+  const [, chainNamespace, chainReference, assetNamespace] = matches
+  let [, , , , assetReference] = matches
```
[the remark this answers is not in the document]
pastaghost: 🔥

packages/caip/src/assetId/assetId.ts
```diff
@@ -124,30 +123,29 @@ export type FromAssetId = (assetId: AssetId) => FromAssetIdReturn
...
 
   if (assetNamespace && assetReference && chainId) {
-    return { chainId, chainReference, chainNamespace, assetNamespace, assetReference }
+    return {
+      chainId,
+      chainReference,
+      chainNamespace,
+      assetNamespace,
+      assetReference:
+        assetNamespace && ['erc20', 'erc721'].includes(assetNamespace)
```
0xApotheosis, line 145: I must have missed this in the previous review, but `assetNamespace` is guaranteed to be defined here as we've already done a type assertion, so we can remove the check.

0xApotheosis (changes requested):
Sorry, one more thing I must have missed in the previous review.

0xApotheosis (approved):
ty for the changes, lgtm!

PR #1013, opened 2022-09-06
feat(caip): add FOX & FOXy assetIds to constants

0xApotheosis:
We hardcode FOX and FOXy as strings throughout `web`, and sometimes use isolated constants for them.

This PR exports them from the CAIP package, giving us a stable source of truth for them moving forward.

Whilst these are the first non-fee EVM asset IDs to be exported, it feels reasonable as these assets in particular are frequently used in client-side code, and are likely to continue to be moving forward.


gomesalexandre (approved):
Conceptually agree with this one but not merging until we further discuss whether or not we want to add such `AssetId`s in caip

> it feels reasonable as these assets in particular are frequently used in client-side code, and are likely to continue to be moving forward.

Indeed, the consensus until now was that the caip package was only responsible for AssetId "primitives" i.e native assets. With that being said, caip is our own flavor of [CASA](https://github.com/ChainAgnostic)'s caip package, and I see no issues in it exporting non-native assets as well, since all the `constants` module really does is give us convenience exports to be consumed in our other packages.

kaladinlight (approved):

PR #1014, opened 2022-09-06
feat(investor-yearn): accept bip44Params in signAndBroadcast input

gomesalexandre:
BREAKING CHANGE: signAndBroadcast now accepts a bip44Params property in its input and will throw if not passed.

### Description

This PR:

- Accepts an optional (for backward-compatibility and not to break the abstracted `InvestorOpportunity` base type) `bip44Params` as input property
- Adds the optional property in the abstracted `InvestorOpportunity` 

### Issue

- Tackles https://github.com/shapeshift/web/issues/2605

### DIFF
--- a/packages/investor-yearn/src/YearnOpportunity.ts
+++ b/packages/investor-yearn/src/YearnOpportunity.ts
@@ -9,6 +9,7 @@ import {
 } from '@shapeshiftoss/investor'
 import { Logger } from '@shapeshiftoss/logger'
 import { KnownChainIds } from '@shapeshiftoss/types'
+import { BIP44Params } from '@shapeshiftoss/types'
 import { type ChainId, type Vault, type VaultMetadata, Yearn } from '@yfi/sdk'
 import type { BigNumber } from 'bignumber.js'
 import isNil from 'lodash/isNil'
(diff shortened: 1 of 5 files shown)


packages/investor-yearn/src/YearnOpportunity.ts
```diff
@@ -176,12 +176,13 @@ export class YearnOpportunity
...
     feePriority?: FeePriority
+    accountNumber?: number
   }): Promise<string> {
-    const { wallet, tx, feePriority } = input
+    const { accountNumber, wallet, tx, feePriority } = input
     const feeSpeed: FeePriority = feePriority ? feePriority : 'fast'
     const chainAdapter = this.#internals.chainAdapter
 
-    const path = toPath(chainAdapter.buildBIP44Params({ accountNumber: 0 }))
+    const path = toPath(chainAdapter.buildBIP44Params({ accountNumber: accountNumber ?? 0 }))
```
0xdef1cafe, line 185: we cannot safely fall back here - we should throw if not provided

0xdef1cafe (changes requested):

packages/investor-yearn/src/YearnOpportunity.ts
gomesalexandre: That's what I was thinking as well, ideally would prefer for the type-system to catch this so we never have to add runtime logic for this, but this would do the trick for now.
Are we happy with this *actually temporary runtime workaround* as a type-safety substitute?

gomesalexandre: Added and rebased interactive to ease landing the `BREAKING CHANGE:` commit body:

https://github.com/shapeshift/lib/blob/474a74421cde7db823dbd0d4b74c9e6affb7e42a/packages/investor-yearn/src/YearnOpportunity.ts#L184

packages/investor-yearn/src/YearnOpportunity.ts
```diff
@@ -176,12 +177,16 @@ export class YearnOpportunity
     wallet: HDWallet
     tx: PreparedTransaction
     feePriority?: FeePriority
+    accountNumber?: number
```
0xdef1cafe, line 180: make this an optional `bip44Params`

0xdef1cafe: and throw if it's not provided - this makes the type sig backwards compatible and means that we know web is up to date

gomesalexandre: Done in b1115cae7c1542776eb72c2d05dfb7853c332f41 - that also solves the problem of `accountNumber` being optional, since we will eventually add it to all investor packages and be able to make it required at `investor` level

PR #1015, opened 2022-09-07
feat: add isAccountTypeRequired() to chain-adapter API

asamere:
adds a new function `isAccountTypeRequired(): boolean` to the chain-adapter api allowing clients (web) to selectively require accountType without devolving to chain(namespace) switching.


asamere:
The test failure is because it needs this HDWallet commit already on master: https://github.com/shapeshift/hdwallet/commit/2efcba42326e0066d68dee5e702064a35d1eab7c#diff-8e302750c51b2b3ee726f599e86c603d1d7882fcaef654533d835251205b8344R143

works on my machine 😁 with `yarn link`

Yarn is pulling the reverted `1.30.0` versions of hdwallet packages: https://github.com/shapeshift/lib/pull/1015/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR2903

0xdef1cafe:
going to close this as the bip44params and accountType live side by side, and we just pass them in under all circumstances - and the compiler correctly complains for UTXO coins so this isn't giving us any better ergonomics than we already have

PR #1017, opened 2022-09-09
chore: add `array-type` lint rule

0xApotheosis:
Adds `array-type` lint rule for codebase consistency.

https://typescript-eslint.io/rules/array-type/


gomesalexandre (approved):
LGTM, checked out locally and ran `yarn lint` which produced no errors.

PR #1019, opened 2022-09-11
fix(swapper): keepkey cow swap problem

markrypt0:
The root node path is a valid path but not the address to sign the order. For some reason the native wallet does not use this path to get the signing address so it's not a problem there.


gomesalexandre (approved):
Tested locally while linked in web, confirmed calling `bip32ToAddressNList` with bip44 path fixes swaps on KK - signature is valid, and Tx is broadcasted populated with the right data.
Tests confirm this and are reflecting the new bip44 `addressNList` as opposed to the previous bip32 list.

image

gomesalexandre (approved):
Re-stamp post branch update

PR #1026, opened 2022-09-14
chore: use fetch for generated typescript client

kaladinlight:


gomesalexandre (approved):
Conceptually looks good - untested but looks sane as this effectively just awaits and unwraps the data field from the fetch API

PR #1032, opened 2022-09-18
feat(investor-foxy): switch from web3 to ethers and JsonRpcBatchProvider

gomesalexandre:
### Description

This switches the current investor-foxy API implementation from web3.js to ethers.js, to be able to use JsonRpcBatchProvider.

web3.js also provides a `BatchRequest` API, unfortunately after spending more than a day on it, it seems like that API is rugged currently and isn't compatible with our programmatic calls with blockNumber specified. 

- [x] Base migration from web3.js to ethers - compiles
- [x] Regression test that values are correct
- [ ] ~~Batch promises together so the underlying JSON RPC calls are automagically batched by `JsonRpcBatchProvider`~~ This will better live in a follow-up PR. Let's get the switch to ethers.js merged first, and then we can improve the business logic

### Issue

N/A

### Risks

Everything FOXy could be borked - test locally against latest `@shapeshiftoss/investor-foxy`.

### DIFF
--- a/packages/investor-foxy/src/api/api.ts
+++ b/packages/investor-foxy/src/api/api.ts
@@ -951,9 +951,11 @@ export class FoxyApi {
     this.verifyAddresses([userAddress, contractAddress])
     const stakingContract = this.getStakingContract(contractAddress)
 
-    let coolDownInfo
+    let coolDownInfo: [amount: string, gons: string, expiry: string]
     try {
-      coolDownInfo = await stakingContract.coolDownInfo(userAddress)
+      coolDownInfo = (await stakingContract.coolDownInfo(userAddress)).map(
+        (info: ethers.BigNumber) => info.toString(),
+      )
     } catch (e) {
       throw new Error(`Failed to get coolDowninfo: ${e}`)
     }
@@ -963,8 +965,12 @@ export class FoxyApi {
     } catch (e) {
       throw new Error(`Failed to getTimeUntilClaimable: ${e}`)
     }
+
+    const [amount, gons, expiry] = coolDownInfo
     return {
-      ...coolDownInfo,
+      amount,
+      gons,
+      expiry,
       releaseTime,
     }
   }
--- a/packages/investor-foxy/src/api/api.ts
+++ b/packages/investor-foxy/src/api/api.ts
@@ -455,21 +455,12 @@ export class FoxyApi {
     }
 
     const stakingContract = this.getStakingContract(contractAddress)
-    const userChecksum = ethers.utils.getAddress(userAddress)
+    // const userChecksum = ethers.utils.getAddress(userAddress)
 
-    const data: string =
-      (
-        await stakingContract.populateTransaction['stake(uint256,address)'](
-          this.normalizeAmount(amountDesired),
-          userAddress,
-          { from: userChecksum, value: 0 },
-        )
-      ).data ?? ''
-
-    // {
-    // value: 0,
-    // from: userChecksum,
-    // })
+    const data = stakingContract.interface.encodeFunctionData('stake(uint256,address)', [
+      this.normalizeAmount(amountDesired),
+      userAddress,
+    ])
 
     const { nonce, gasPrice } = await this.getGasPriceAndNonce(userAddress)
     const chainReferenceAsNumber = Number(this.ethereumChainReference)
--- a/packages/investor-foxy/src/api/api.ts
+++ b/packages/investor-foxy/src/api/api.ts
@@ -964,7 +964,7 @@ export class FoxyApi {
     const contract = new ethers.Contract(tokenContractAddress, foxyAbi, this.provider)
 
     try {
-      const balance = await contract.methods.circulatingSupply().call()
+      const balance = await contract.circulatingSupply()
       return bnOrZero(balance)
     } catch (e) {
       throw new Error(`Failed to get tvl: ${e}`)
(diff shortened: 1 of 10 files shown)


gomesalexandre:
* main
  * **PR #1032** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1032" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1032?utm_source=stack-comment).

packages/investor-foxy/src/api/api.ts
```diff
@@ -85,12 +83,11 @@ const TOKE_IPFS_URL = 'https://ipfs.tokemaklabs.xyz/ipfs'
...
 export class FoxyApi {
   public adapter: ChainAdapter<KnownChainIds.EthereumMainnet>
-  public provider: HttpProvider
+  public provider: JsonRpcBatchProvider
   private providerUrl: string
-  public jsonRpcProvider: JsonRpcProvider
-  public web3: Web3
-  private foxyStakingContracts: Contract[]
-  private liquidityReserveContracts: Contract[]
+  public jsonRpcProvider: JsonRpcBatchProvider
```
gomesalexandre, line 88: Do we still need both?

packages/investor-foxy/src/api/api.ts
```diff
@@ -121,7 +118,7 @@ export class FoxyApi {
    * @param amount
    */
   private normalizeAmount(amount: BigNumber) {
-    return this.web3.utils.toBN(amount.toFixed())
+    return ethers.BigNumber.from(amount.toFixed())
```
gomesalexandre, line 121: This is only used internally so the ethers BigNumber flavor is correct - however the types are now wrong, this is assuming a `BigNumber.js` number. Update return type to reflect this

packages/investor-foxy/src/api/api.ts
```diff
@@ -132,8 +129,8 @@ export class FoxyApi {
       if (dryRun) return signedTx
       try {
         if (this.providerUrl.includes('localhost') || this.providerUrl.includes('127.0.0.1')) {
-          const sendSignedTx = await this.web3.eth.sendSignedTransaction(signedTx)
-          return sendSignedTx?.blockHash
+          const sendSignedTx = await this.provider.sendTransaction(signedTx)
+          return sendSignedTx?.blockHash ?? ''
```
gomesalexandre, line 133: This is confirming to the `sendTransaction` return type - runtime logic didn't change here

packages/investor-foxy/src/api/api.ts
```diff
@@ -163,47 +161,49 @@ export class FoxyApi {
...
 
-  private async getGasPriceAndNonce(userAddress: string) {
-    let nonce: number
+  private async getGasPriceAndNonce(
+    userAddress: string,
+  ): Promise<{ nonce: string; gasPrice: string }> {
+    let nonce
     try {
-      nonce = await this.web3.eth.getTransactionCount(userAddress)
+      nonce = await this.provider.getTransactionCount(userAddress)
```
gomesalexandre, line 185: `toString()` this in place for consistency

packages/investor-foxy/src/api/api.ts
```diff
@@ -455,25 +447,22 @@ export class FoxyApi {
...
     try {
-      estimatedGasBN = await this.estimateDepositGas(input)
+      estimatedGas = await this.estimateDepositGas(input)
     } catch (e) {
       throw new Error(`Estimate Gas Error: ${e}`)
     }
 
     const stakingContract = this.getStakingContract(contractAddress)
-    const userChecksum = this.web3.utils.toChecksumAddress(userAddress)
+    // const userChecksum = ethers.utils.getAddress(userAddress)
```
gomesalexandre, line 458: Remove me  - 0-value and `from` as a checksummed version of the caller address are implied as confirmed by the broadcasted Txs on-chain, calldata is the exact same with or without these `value` and `from` params

packages/investor-foxy/src/api/api.ts
```diff
@@ -981,25 +967,33 @@ export class FoxyApi {
     } catch (e) {
       throw new Error(`Failed to getTimeUntilClaimable: ${e}`)
     }
+
+    const [amount, gons, expiry] = coolDownInfo
     return {
-      ...coolDownInfo,
+      amount,
+      gons,
+      expiry,
```
gomesalexandre, line 975: Explicit vs. implicit, this was previously rugged

gomesalexandre:
Opening for early review, added self-review comments

packages/investor-foxy/src/api/api.ts
gomesalexandre: Disregard me, this is *actually* fine: as was found previously, what ethers call `BigNumber` is in fact ethers own `BigNumber` flavor, which is properly typed as a return type here - still added an explicit return value

packages/investor-foxy/src/api/api.ts
```diff
@@ -237,90 +237,86 @@ export class FoxyApi {
...
-  ): Promise<BigNumber> {
+  async estimateSendWithdrawalRequestsGas(input: TxInputWithoutAmountAndWallet): Promise<string> {
     const { userAddress, contractAddress } = input
     this.verifyAddresses([userAddress, contractAddress])
 
     const stakingContract = this.getStakingContract(contractAddress)
 
     try {
-      const estimatedGas = await stakingContract.methods.sendWithdrawalRequests().estimateGas({
+      const estimatedGas = await stakingContract.estimateGas.sendWithdrawalRequests([], {
```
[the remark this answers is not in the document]
gomesalexandre: Disregard me, this is literal dead chonky code https://github.com/shapeshift/lib/pull/1130

PR #1035, opened 2022-09-19
fix: use json response instead of text response for generated fetch api

kaladinlight:
The default parsing strategy for typescript-fetch is to use `.text()` for primitive response types. We have two unchained endpoints (`/gas/estimate` and `/send`) that currently return plain text responses. This resulted in a value `'"some string"'` being returned by the generated typescript client.

Using `json()` to parse these responses strips the quotes as desired and is the quick fix to prevent blocking lib/releases. Note, I have not tested against any other primitive type responses other than strings, but for our purposes at this time it works. Longer term fix would be to update the response payloads to be json across the board even if only returning a single string response and then we could remove the post processing to change parsing strategy.


gomesalexandre (approved):
Confirmed this fixes the issue with the encoding of the `gasLimit` property:

image

This also properly encodes the txid, as confirmed by clicking the "View on chain" button on pending FOX/ETH stake

PR #1040, opened 2022-09-27
fix(swapper): typo in MidgardCoins

0xApotheosis:


gomesalexandre (approved):
Nice catch ser! 🕵🏽‍♂️

PR #1044, opened 2022-09-29
fix: remove HEGIC from thor pools

0xdef1cafe:
pool is gone upstream https://midgard.ninerealms.com/v2/pools


gomesalexandre (approved):
Confirmed pool is gone upstream

🧁

PR #1046, opened 2022-10-04
fix(swapper): plumb `sendMax` through ThorSwapper

0xApotheosis:
The ThorChain swapper isn't doing anything with the `sendMax` argument being sent from the client.

This PR plumbs it through ThorSwap's `buildTrade` and `getTradeQuote`.

To test locally, link with https://github.com/shapeshift/web/pull/2955 and click "Max" on a UTXO asset - ensure we can get to the confirm trade screen without throwing.

Contributes to https://github.com/shapeshift/web/issues/2949, though requires the `web` PR to fully close: https://github.com/shapeshift/web/pull/2955

### DIFF
--- a/packages/swapper/src/swappers/thorchain/utils/txFeeHelpers/btcTxFees/getBtcTxFees.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/txFeeHelpers/btcTxFees/getBtcTxFees.ts
@@ -11,6 +11,7 @@ export const getBtcTxFees = async ({
   sellAdapter,
   pubkey,
   buyAssetTradeFeeUsd,
+  sendMax,
 }: {
   opReturnData: string
   vault: string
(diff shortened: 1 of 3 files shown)


packages/swapper/src/swappers/thorchain/utils/txFeeHelpers/btcTxFees/getBtcTxFees.ts
```diff
@@ -4,26 +4,31 @@ import { KnownChainIds } from '@shapeshiftoss/types'
...
 import { bn } from '../../../../utils/bignumber'
 
+type GetBtcTxFeesInput = {
+  opReturnData: string
+  vault: string
+  sellAmount: string
+  sellAdapter: UtxoBaseAdapter<UtxoSupportedChainIds>
+  pubkey: string
+  buyAssetTradeFeeUsd: string
+  sendMax: boolean
```
0xApotheosis, line 14: The only thing new here is the addition of the `sendMax` argument.

gomesalexandre (approved):
Tested locally, LGTM

All UTXOs inputs are used when selecting "Max", as the property is now properly plumbed through (confirmed in Tx explorer with the Tx inputs/output totalling to the account's balance)

image

PR #1051, opened 2022-10-05
fix: allow thor trades for wallets that require signAndBroadcast

0xdef1cafe:
we previously missed this and only native/keepkey support thor trades as they support offline
signing

this uses the signAndBroadcast method if it is available, e.g. metamask/xdefi/tally ho/walletconnect


0xdef1cafe:
tested and working with metamask/tally ho

kaladinlight (approved):

PR #1053, opened 2022-10-06
feat(swapper): getAllowanceRequired -> isApprovalRequired

gomesalexandre:
### Description

This PR renames `getAllowanceRequired` to `isApprovalRequired` and changes its signature, making it more ergonomic:

The current logic of `getAllowanceRequired` effectively calculates the remaining allowance needed to interact with x contract, and returns it so we can determine whether or not an approval is needed. However, we never actually use this number, and just use `.gt`/ `.lt` on the returned BN

By renaming this to `isApprovalRequired` and changing the return type to a boolean, we improve the current decoupling, and make it so that consumers don't need to do additional operations on the returned value and can use it as is.


gomesalexandre:
* main
  * **PR #1053** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1053" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1053?utm_source=stack-comment).

0xdef1cafe (approved):

PR #1067, opened 2022-10-18
fix: labelling on the USDT WH assets

crowicked:
Fixes labelling on the USDT WH assets (USDTso and USDTbs), closes issue [#2853](https://github.com/shapeshift/web/issues/2853) in `shapeshift/web`

image


packages/asset-service/src/generateAssetData/overrides.json
```diff
@@ -2,5 +2,29 @@
...
 		"color": "#3761F9",
 		"icon": "https://assets.coincap.io/assets/icons/256/fox.png"
+	},
+	"eip155:1/erc20:0x1cdd2eab61112697626f7b4bb0e23da4febf7b7c": {
+		"explorer": "https://etherscan.io",
+		"explorerAddressLink": "https://etherscan.io/address/",
+		"explorerTxLink": "https://etherscan.io/tx/",
+		"color": "#55AD96",
+		"icon": "https://rawcdn.githack.com/trustwallet/assets/master/blockchains/ethereum/assets/0x1CDD2EaB61112697626F7b4bB0e23Da4FeBF7B7C/logo.png",
+		"name": "USDTso Wormhole",
```
0xdef1cafe, line 12: we really only need to override the name and the symbol for both of these assets, and only do it in this file, not in both.

crowicked: @0xdef1cafe @kaladinlight @0xApotheosis 
Initially did it because I couldn't get it to fix the labelling otherwise. But now I did more testing regarding generation and found some interesting things that are kind of curious (worrisome?):

1. Adding just the `name` and `symbol` overrides in `overrides.json` for `USDTso` and `USDTbs`  - while `USDTso` one properly updated just those fields/props, it would overwrite the entire entry for `USDTbs` in `generatedAssetData.json` with only those two props, resulting in a truncated/broken `generatedAssetData.json` entry:
```
"eip155:1/erc20:0xde60adfddaabaaac3dafa57b26acc91cb63728c4": {
    "name": "USDTbs Wormhole",
    "symbol": "USDTbs",
  }
```
2. Adding the full asset objects with `name` and `symbol` properties updated in `overrides.json` and running `yarn generate` would generate a properly updated `generatedAssetData.json`.
BUT it would still not reflect that update in `shapeshift/web`. 
When I ran `yarn generate` the second time immediately after (meaning `generatedAssetData.json` now already has the properly updated `name` and `symbol` fields), the label changes would be visible in `shapeshift/web` and the issue resolved.
This leads me to believe that even when the `generatedAssetData.json` is populated with proper updated values, the override values are not picked up - unless the updated values are already in `generatedAssetData.json` to begin with. It also works to fix the issue and gets updated properly if only added to `overrides.ts`  under the `ethereum` folder.

If someone can try and reproduce these in their own local dev that would be helpful.
Curious to hear your thoughts.

crowicked: @kaladinlight try with both asset overrides to reproduce, the `USDTso` one gets props updated properly for me as well, the `USDTbs` does not.
Also, try running web locally after - the labels aren't updated in the UI until you run generate the second time (with the `generatedAssetData.json` already populated with the updated values).

kaladinlight:
closing in favor of: https://github.com/shapeshift/lib/pull/1072

PR #1072, opened 2022-10-20
fix: improve asset generation and filter or blacklist all known wormhole assets

kaladinlight:


gomesalexandre (approved):
LGTM, tested:

- `generate` script
- `generate-colors-map` script
- Local symlink, web properly blacklists wormhole assets and overrides are still working

PR #1074, opened 2022-10-21
fix: avax asset ref and upgrade yearn sdk

kaladinlight:


packages/investor-yearn/src/YearnOpportunity.ts
```diff
@@ -104,7 +104,6 @@ export class YearnOpportunity
     this.apy = bnOrZero(vault.metadata.apy?.net_apy)
     this.isNew = vault.metadata.apy?.type === 'new'
     this.name = `${vault.metadata.displayName} ${vault.version}`
-    this.symbol = vault.metadata.symbol
```
kaladinlight, line 107: already setting this on line 103, `vault.symbol` not optional. not sure what `vault.metadata.symbol` is meant to imply, assuming it wouldn't differ from `vault.symbol`?

gomesalexandre: Yeah the double assignment is wrong here, looks like the intent was using the former

gomesalexandre (approved):

0xApotheosis (approved):

PR #1075, opened 2022-10-23
fix(swapper): fix rune as sell asset

0xApotheosis:
Fixes quote when sell asset is native RUNE and no amounts have been entered yet.

Closes https://github.com/shapeshift/web/issues/3066

**Before**

Screen Shot 2022-10-23 at 6 09 00 pm

**After**

Screen Shot 2022-10-23 at 6 05 51 pm

### DIFF
--- a/packages/swapper/src/swappers/thorchain/utils/getPriceRatio/getPriceRatio.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/getPriceRatio/getPriceRatio.ts
@@ -52,6 +52,17 @@ export const getPriceRatio = async (
       return bn(runeBalance).dividedBy(sellAssetBalance).toString()
     }
 
+    if (isRune(sellAssetId) && buyPool) {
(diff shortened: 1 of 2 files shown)


packages/swapper/src/swappers/thorchain/utils/getPriceRatio/getPriceRatio.ts
```diff
@@ -52,6 +52,17 @@ export const getPriceRatio = async (
       return bn(runeBalance).dividedBy(sellAssetBalance).toString()
     }
 
+    if (isRune(sellAssetId) && buyPool) {
```
0xApotheosis, line 55: This whole file is an absolute spaghetti-filled dumpster fire with innumerable conditional returns and rune-specific logic sprinkled throughout - and this PR only adds to it.

gomesalexandre (approved):
Tested with linked lib locally, rate is properly gotten before entering an amount with RUNE as a sell asset

image
image
image

🍦

PR #1077, opened 2022-10-25
fix: yearn SDK rug

0xApotheosis:


0xApotheosis:
This might not be the fix after all - the API response used does not contain enough information to map to the existing classes.

PR #1079, opened 2022-10-25
fix: pass back metadata to onmessage handler

kaladinlight:
- create shared `parsedTxToTransaction` function that both `subscribeTxs.onMessage` and `getTxHistory` can use
- metadata is now passed back in both cases resulting in correct transaction row display in the client
- may have also fixed trade confirmation display in trade modal as I am now seeing that update (although with polling it is not in sync with incoming buy side tx)

### DIFF
--- a/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
@@ -49,6 +49,27 @@ const transformValidator = (validator: unchained.cosmossdk.types.Validator): Val
   apr: validator.apr,
 })
 
+const parsedTxToTransaction = (parsedTx: unchained.cosmossdk.ParsedTx): Transaction => ({
+  address: parsedTx.address,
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -49,6 +49,27 @@ const transformValidator = (validator: unchained.cosmossdk.types.Validator): Val
   apr: validator.apr,
 })
 
+const parsedTxToTransaction = (parsedTx: unchained.cosmossdk.ParsedTx): Transaction => ({
+  address: parsedTx.address,
```
0xdef1cafe, line 53: why no spreadsies?

kaladinlight: good point, there was a time when they were more distinct, but only diff down is `value` vs `totalValue` in transfers. will update

0xdef1cafe (approved):

Issue #1050, opened 2022-10-05
THORChain minimum trade size estimation


0xdef1cafe:
### Overview

We want to ensure the minimum trade amount is at least the `outbound_fee` that the node/midgard reports, otherwise swaps will fail and refund (or even fail to refund).

Note that ETH or ERC20 as a buy asset will require significantly higher fees as THORChain uses aggressive gas pricing to ensure swaps complete in a timely fashion, e.g. at the time of writing a rapid send is ~27 gwei and midgard is using 60.

<img width="667" alt="image" src="https://user-images.githubusercontent.com/88504456/194157521-a25417c5-8e38-44d3-b4b2-6e63ba0a10b7.png">

### References and additional details

* hit https://daemon.thorchain.shapeshift.com/lcd/thorchain/inbound_addresses to when building a quote to get the `outbound_fee`
* return this to web and calculate minimum amounts there

### Acceptance Criteria

* We can reliably execute swaps with the correct minimum amount, and prevent users from executing swaps that will result in a refund, or worse, failure to refund.

### Screenshots/Mockups

<img width="1034" alt="image" src="https://user-images.githubusercontent.com/88504456/194156317-010366b9-468d-4a60-94ce-91932548ef57.png">

PR #1080, opened 2022-10-28
feat(swapper): use refund fee in limit calculation & use `outbound_fee`

0xApotheosis:
This PR:

- Adds logic to factor in a THORSwap trade's potential refund fee when calculating the limit of the output asset. If it's going to cost more to refund the transaction than it is to complete it we'll subtract this amount from the limit amount instead of the fee (i.e. we are ok to receive a little less of the buy asset because if we didn't we'd lose out anyway on the refund cost).
- Uses the `outbound_fee` from the `inbound_addresses` response from the Thornode API to determine the buy asset's trade fee, rather than estimating this manually
- Does a general tidy-up

**Related tickets**
- Closes https://github.com/shapeshift/web/issues/3069 - whilst the slippage remains unchanged, we should no longer get those refunded transactions, as now if it were economical to do so the trade will proceed.
- Closes https://github.com/shapeshift/lib/issues/1050


packages/swapper/src/swappers/thorchain/utils/getLimit/getLimit.ts
```diff
@@ -34,19 +54,56 @@ export const getLimit = async ({
...
 
+  const sellAssetPoolId = adapters.assetIdToPoolAssetId({ assetId: sellAsset.assetId })
+  const sellAssetChainSymbol = sellAssetPoolId?.slice(0, sellAssetPoolId.indexOf('.'))
+  const sellAssetAddressData = await getInboundAddressDataForChain(
+    deps.daemonUrl,
+    sellAssetChainSymbol,
+  )
+
+  // We want this in the buy asset crypto precision
+  const refundFeeBuyAssetCryptoPrecision8 = (() => {
```
0xApotheosis, line 72: This is the crux of the logic with respect to factoring in the refund fee when setting the limit in the TX memo.

0xApotheosis (commented):
Commentary for reviewers.

gomesalexandre (commented):
Spot-on PR, both the logic for the fetched `outbound_fee` usage for refund fee calculation, and a much needed swapper cleanup look good to me - consider this comment a stamp pending runtime testing

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -71,42 +75,35 @@ export const getThorTradeQuote: GetThorTradeQuote = async ({ deps, input }) => {
...
+    })
 
-    const minimumSellAssetAmountCryptoHuman = (() => {
-      // The 1$ minimum doesn't apply for swaps to RUNE, use OutboundTransactionFee in human value instead
-      if (isRune(buyAsset?.assetId)) return RUNE_OUTBOUND_TRANSACTION_FEE_CRYPTO_HUMAN
+    const buyAssetTradeFeeUsd = bn(buyFeeAssetUsdRate)
+      .times(estimatedBuyAssetTradeFeeFeeAssetCryptoHuman)
+      .toString()
+
+    const minimumSellAssetAmountCryptoHuman = bnOrZero(buyAssetTradeFeeUsd).div(sellAssetUsdRate)
```
asamere, line 99: is it possible sellAssetUsdRate is 0 here?

packages/swapper/src/swappers/thorchain/utils/getInboundAddressDataForChain.ts
```diff
@@ -0,0 +1,17 @@
...
+  daemonUrl: string,
+  assetId: AssetId | undefined,
+): Promise<InboundAddressResponse | undefined> => {
+  if (!assetId) return undefined
+  const assetPoolId = adapters.assetIdToPoolAssetId({ assetId })
+  const assetChainSymbol = assetPoolId?.slice(0, assetPoolId.indexOf('.'))
+  const { data: inboundAddresses } = await thorService.get<InboundAddressResponse[]>(
+    `${daemonUrl}/lcd/thorchain/inbound_addresses`,
+  )
+  return inboundAddresses.find((inbound) => inbound.chain === assetChainSymbol)
```
asamere, line 16: we should filter/exclude any address that has `"halted": true` property

0xApotheosis: For sure. Actioned in https://github.com/shapeshift/lib/pull/1080/commits/199015a45ca18acc92337e5b305d8712f264203a.

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
0xApotheosis: A bad server response could certainly make that possible, which would leave us with the string "Infinity" as the `minimumCryptoHuman` - not ideal.

Guarded against this in https://github.com/shapeshift/lib/pull/1080/commits/4a0205c8e9eba6818f3ae6e14543ffb2611549d7.

0xApotheosis:
Ok @gomesalexandre I've tested this and it seems to all be working as expected.

- Quotes / fees roughly match those on THORSwap (so the consumption of `outbound_fee` must be doing its job)
- THORTrades execute as expected

gomesalexandre (approved):
Tested with linked lib, LGTM

#### RUNE - ATOM

- Outbound fee is matching the Ninerealms / Thorswap one

image
image
image

- Inputting an amount higher than the outbound fee allows to proceed with the trade

image

- Inputting an amount lower than the outbound fee doesn't allow to proceed with the trade

image

#### ATOM - RUNE

- Inputting an amount higher than the 0.02 RUNE outbound fee allows to proceed with the trade

image

- Inputting an amount lower than the 0.02 RUNE outbound fee doesn't allow to proceed with the trade

image

#### UTXO - UTXO

- Outbound fee is matching the Ninerealms / Thorswap one

image
image
image

- Inputting an amount higher than the outbound fee allows to proceed with the trade

image

- Inputting an amount lower than the outbound fee doesn't allow to proceed with the trade

image

PR #1082, opened 2022-10-31
chore: generate from source swagger

kaladinlight:
- generate typescript clients from source `swagger.json` checked into repo (prevents failed builds if unchained api goes down)
- reduce environment complexity that has never been used in the generation configs


0xdef1cafe (approved):
ty bigly, does what it says

PR #1089, opened 2022-11-17
feat: idle tranche rewardAssetIds

gomesalexandre:
#### Description

This PR adds support for Idle tranche opportunities reward assets.

Notes: 

- `getClaimableTokens` was left untouched as the actual value of those rewards is always guaranteed to be 0, this data is only needed for completeness / being able to display reward tokens.
- This isn't guaranteed to give us all the reward AssetIds for tranche contracts. This is as programmatic as we can get:

> <cite>@samster91<cite>
unfortunately there are other tokens such as CVX or CRV that are handled directly by the underlying strategy, so we display them in the client but hard-coded

The way we'll be able to achieve full completeness is the same way this is done on Idle.finance dApp, i.e hardcoding it. Web already has support for this in the https://github.com/shapeshift/web/pull/3355 draft, see:

https://github.com/shapeshift/web/blob/fd75f42f584c631f61876d9dd670c8211eb021e0/src/state/slices/opportunitiesSlice/resolvers/idle/constants.ts

Agreed with Idle to provide missing reward AssetIds once the web PR goes in.


gomesalexandre:
* main
  * **PR #1089** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1089" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1089?utm_source=stack-comment).

0xdef1cafe (approved):

PR #1090, opened 2022-11-18
feat: calculate usd price from pools via thornode

kaladinlight:
- midgard is slow replace with thornode
- thornode does not have usd price immediately available, calculate from pool ratios
- get average rune usd price from all available usd pools for most accurate price representation
- set stage for improved function signatures via class variables (eventually want to pull in helper functions to reduce file bloat)
- small cleanup/performance improvement for tracking available pools on `swapper.initialize()`


packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
```diff
@@ -65,12 +68,21 @@ export class ThorchainSwapper implements Swapper<ChainId> {
...
+  midgardUrl: string
+  adapterManager: ChainAdapterManager
+  web3: Web3
 
   constructor(deps: ThorchainSwapperDeps) {
     this.deps = deps
+    this.daemonUrl = deps.daemonUrl
+    this.midgardUrl = deps.midgardUrl
+    this.adapterManager = deps.adapterManager
+    this.web3 = deps.web3
```
gomesalexandre, line 85: Do we need to inject this? Doesn't seem used anywhere

gomesalexandre (commented):
Conceptually looks good to me, though comment-as-a-stamp (CaaS), untested as I'm in a different domain currently

packages/swapper/src/swappers/thorchain/ThorchainSwapper.ts
kaladinlight: not yet, just setting the stage to use class variables and pull in helper functions to improve readability and maintainability of these swappers

kaladinlight: i can remove until used I suppose, just want to start cleaning these up some, but didn't want to add too much to this pr.

0xApotheosis (approved):
Tested whilst linked locally, LGTM - nice tidy up and optimisation.

PR #1092, opened 2022-11-21
fix: return idle raw TVL values

gomesalexandre:
#### Description

TVL parsing in investor-idle is wrong, confirmed the intent was to trim
the values to 6 significant digits (the decimal places of USDC the TVL
is denoted in), but this effectively parsed the numbers as way bigger
ones (e.g trillions instead of millions)

This PR removes the parsing and returns the raw numbers from Idle's API, so that web can
apply the parsing it needs.

### DIFF
--- a/packages/investor-idle/src/IdleOpportunity.ts
+++ b/packages/investor-idle/src/IdleOpportunity.ts
@@ -158,8 +158,8 @@ export class IdleOpportunity
     this.expired = false
     this.apy = bnOrZero(vault.apr).div(100)
     this.tvl = {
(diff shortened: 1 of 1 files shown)


gomesalexandre:
* main
  * **PR #1092** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1092" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1092?utm_source=stack-comment).

PR #1096, opened 2022-11-22
fix: return idle raw TVL values (#1092)

gomesalexandre:
feat(investor-idle): don't 'normalize' amounts

PR #1097, opened 2022-11-22
feat(market-service): idle market service

gomesalexandre:
#### Description

This PR:

- uses the right calculation to get the price of Idle opportunity tokens
- adds an `idle` market-service, getting Idle opportunity tokens' price and market cap

#### Screenshots

image
image

### DIFF
--- a/packages/market-service/src/idle/idle.ts
+++ b/packages/market-service/src/idle/idle.ts
@@ -30,9 +30,6 @@ export class IdleMarketService implements MarketService {
       }),
       providerUrl: this.providerUrls.jsonRpcProviderUrl,
     })
-    ;(async () => {
-      await this.idleInvestor.initialize()
-    })()
   }
 
   async findAll() {
@@ -54,8 +51,8 @@ export class IdleMarketService implements MarketService {
       acc[assetId] = {
         price: opportunity.tvl.balanceUsdc.div(opportunity.tvl.balance).toFixed(),
         marketCap: opportunity.tvl.balanceUsdc.toFixed(), // For Idle, TVL and marketCap are effectively the same
-        volume: '0', // TODO:
-        changePercent24Hr: 0, // TODO
+        volume: '0',
+        changePercent24Hr: 0,
       }
 
       return acc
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* main
  * **PR #1096** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1096" target="_blank">Graphite</a> 
    * **PR #1097** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1097?utm_source=stack-comment).

packages/market-service/src/idle/idle.ts
```diff
@@ -0,0 +1,89 @@
...
+      await this.idleInvestor.initialize()
+    })()
+  }
+
+  async findAll() {
+    const idleOpportunities = await (async () => {
+      const maybeOpportunities = await this.idleInvestor.findAll()
+      if (maybeOpportunities.length) return maybeOpportunities
+
+      await this.idleInvestor.initialize()
```
0xdef1cafe, line 43: prob don't need to reinitialize if we do this in the constructor

packages/market-service/src/idle/idle.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    return marketDataById
+  }
+
+  async findByAssetId({ assetId }: MarketDataArgs): Promise<MarketData | null> {
+    const opportunity = await (async () => {
+      const maybeOpportunities = await this.idleInvestor.findAll()
+      if (maybeOpportunities.length) return await this.idleInvestor.findByOpportunityId(assetId)
+
+      await this.idleInvestor.initialize()
```
0xdef1cafe, line 72: same ere

packages/market-service/src/idle/idle.ts
```diff
@@ -0,0 +1,89 @@
...
+            new unchained.ethereum.Configuration({
+              basePath: this.providerUrls.unchainedEthereumHttpUrl,
+            }),
+          ),
+        },
+        rpcUrl: this.providerUrls.jsonRpcProviderUrl,
+      }),
+      providerUrl: this.providerUrls.jsonRpcProviderUrl,
+    })
+    ;(async () => {
```
gomesalexandre, line 33: Since we do this yeefee in a constructor but constructors are synchronous, it's not going to be blocking the way we think this is, so this is unsafe - hence the re-initialization in other methods

packages/market-service/src/idle/idle.ts
gomesalexandre: Added a comment WRT the unsafety of the constructor initialization https://github.com/shapeshift/lib/pull/1097/files#r1029517425 - happy to remove it in the constructor since it looks like we are too fast to finish initializing before these are run and always end up re-initializing

0xdef1cafe (approved):

PR #1098, opened 2022-11-22
fix: add floor of 0 to limit

0xApotheosis:
If we are sending a small amount to do a Thor trade, and the `outbound_fee` of the sell asset is high, it's possible that the refund fee is higher than the value of the trade.

In this case the limit should be 0 (but not negative), as in "finish this trade no matter what, as the amount is too small to refund it anyway".

The current logic, however, lets the memo limit go negative, which causes an error when executing the TX, e.g. https://live.blockcypher.com/btc/tx/60897031eec3714b079a8d5acb055e84f214bf037863fdcdd11d430c0ca71b5e/

### DIFF
--- a/packages/swapper/src/swappers/thorchain/utils/getLimit/getLimit.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/getLimit/getLimit.ts
@@ -94,15 +94,16 @@ export const getLimit = async ({
   })()
 
   const highestPossibleFeeCryptoPrecision8 = max([
-    buyAssetTradeFeeCryptoPrecision8,
-    refundFeeBuyAssetCryptoPrecision8,
+    // both fees are denominated in buy asset crypto precision 8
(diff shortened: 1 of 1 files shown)


packages/swapper/src/swappers/thorchain/utils/getLimit/getLimit.ts
```diff
@@ -94,13 +94,22 @@ export const getLimit = async ({
   })()
 
   const highestPossibleFeeCryptoPrecision8 = max([
-    buyAssetTradeFeeCryptoPrecision8,
-    refundFeeBuyAssetCryptoPrecision8,
+    // both fees are denominated in buy asset crypto precision 8
+    bnOrZero(buyAssetTradeFeeCryptoPrecision8).toNumber(),
+    bnOrZero(refundFeeBuyAssetCryptoPrecision8).toNumber(),
```
0xApotheosis, line 99: I originally wrote this. I shouldn't be an engineer.

0xdef1cafe (approved):

PR #1101, opened 2022-11-27
chore: use EVM-centric naming

0xApotheosis:
Update some naming to make it clear that the values and functions are used for all EVMs, not just Ethereum.

Note, the hardcoded gas limit value still needs to be replaced (though it does work for Ethereum and Avalanche). 
I've reached out to 9R for more info on how best to do this.


gomesalexandre (approved):
Renaming looks tidy to me 🧹

PR #1103, opened 2022-11-29
chore: improve typing with enums

0xApotheosis:
We use strings in places where we should be using enums, which can, and does lead to errors.
E.g. we are currently passing the string `thorchain` as the THORSwapper name, when it should be `THORChain`.


0xdef1cafe (approved):
insta stamp

0xdef1cafe (approved):

PR #1109, opened 2022-12-01
feat: consolidate Dex enum with SwapperName

0xApotheosis:
We have 2 enum's that serve the same purpose: `Dex` and `SwapperName`.

This PR consolidates the two to use the more commonly used `SwapperName`.
It also allows us to match on the `Trade`'s `dexName` property in `useTxDetails.ts` in `web` to fix the 
transaction link issue described in https://github.com/shapeshift/web/issues/3446.

Note, this is a breaking change. I'll open it once the corresponding `web` PR is ready.


0xApotheosis:
Considering our current dependancy tree, I'm going to think of another approach to this.

PR #1113, opened 2022-12-01
feat(unchained-client): detect failed transaction for cosmossdk chains

kaladinlight:
- use error log to determine if transaction failed


0xApotheosis (approved):
Tidy 🍣

PR #1120, opened 2022-12-05
chore: use bertha for PR and release

0xApotheosis:
This should just work, as bertha is already shared with `lib`.

Screenshot 2022-12-05 at 9 35 50 am


gomesalexandre (approved):
LGTM

bigurl

PR #1121, opened 2022-12-05
chore: fix yarn lock

0xApotheosis:
We forgot to run `yarn.lock` at some stage recently.


gomesalexandre (approved):
LGTM, ran `yarn` locally and diff checks out

PR #1133, opened 2022-12-11
feat(chain-adapters): add osmosis lp chain adapters

pastaghost:
* adds chain adapters for Osmosis LP messages


packages/chain-adapters/src/cosmossdk/osmosis/OsmosisChainAdapter.ts
```diff
@@ -269,6 +271,56 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.OsmosisMain
     }
   }
 
+  async buildLPAddTransaction(
```
elmutt, line 274: adding cli example calling these functions might be helpful during the dev process later when integrating with web. maybe it doesnt matter tho

elmutt (approved):

packages/chain-adapters/src/cosmossdk/types.ts
```diff
@@ -88,4 +88,9 @@ export type FeeData = {
   gasLimit: string
 }
 
+export type Token = {
```
0xdef1cafe, line 91: early feedback, slight nitpick, can we give this a slightly more specific name in this repo? `CosmosSDKToken` perhaps?

toshiSat:
Weird naming conventions, in regards to InMaxs and outMins, but it follows what OSMO is using, so I think it's ok, LGTM

elmutt (approved):

packages/chain-adapters/src/cosmossdk/types.ts
```diff
@@ -88,4 +88,9 @@ export type FeeData = {
   gasLimit: string
 }
 
+export type CosmosSDKToken = {
```
elmutt, line 91: I wonder if this belongs here or even needs its own type at all?

It looks like it's only going to be used in the context of building an osmosis tx.  Sure maybe it could be used for cosmos related code in the future but we don't really know that.

My 2 cents is leaving it out of the cosmos types reduces complexity and spagetification by a very small amount (adding to 1 type file is nicer than 2)

But its so small it really doesnt matter. just a thought

pastaghost: @elmutt I see where you're coming from. I did it this way because personally I think that 
```
export type BuildLPAddTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenInMaxs: [cosmossdk.CosmosSDKToken, cosmossdk.CosmosSDKToken]
}

export type BuildLPRemoveTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenOutMins: [cosmossdk.CosmosSDKToken, cosmossdk.CosmosSDKToken]
}
```

is more readable than 
```
export type BuildLPAddTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenInMaxs: [{ denom: string; amount: string }, { denom: string; amount: string }]
}

export type BuildLPRemoveTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenOutMins: [{ denom: string; amount: string }, { denom: string; amount: string }]
}
```

This is more about making the code self-documenting than it is about future reusability of the type, but that's definitely an option should we add more Cosmos zones in the near future. This is a small thing, so maybe let's just leave it for now.

pastaghost: @elmutt I see where you're coming from. I did it this way because personally I think that 
```
export type BuildLPAddTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenInMaxs: [cosmossdk.CosmosSDKToken, cosmossdk.CosmosSDKToken]
}

export type BuildLPRemoveTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenOutMins: [cosmossdk.CosmosSDKToken, cosmossdk.CosmosSDKToken]
}
```

is more readable than 
```
export type BuildLPAddTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenInMaxs: [{ denom: string; amount: string }, { denom: string; amount: string }]
}

export type BuildLPRemoveTxInput<T extends ChainId> = BuildLPTxInput<T> & {
  tokenOutMins: [{ denom: string; amount: string }, { denom: string; amount: string }]
}
```

This is more about making the code self-documenting than it is about future reusability of the type, but that's definitely an option should we add more Cosmos zones in the near future. This is a small thing, so maybe let's just leave it for now.

packages/chain-adapters/src/cosmossdk/types.ts
kaladinlight: I personally don't like this naming convention when using aliased imports. `cosmossdk.Token` vs the stuttered `cosmossdk.CosmosSDKToken`, just as my 2 cents.

kaladinlight (approved):

packages/chain-adapters/src/cosmossdk/types.ts
kaladinlight: I prefer types for objects as you can also use them when construction the input payload if needed.

0xApotheosis: +1, LGTM @pastaghost.

packages/chain-adapters/src/types.ts
```diff
@@ -173,6 +173,19 @@ export type BuildDepositTxInput<T extends ChainId> = Omit<BuildSendTxInput<T>, '
   memo: string
 }
 
+type BuildLPTxInput<T extends ChainId> = Omit<BuildSendTxInput<T>, 'to' | 'value'> & {
+  poolId: string
+  shareOutAmount: string
+}
+
+export type BuildLPAddTxInput<T extends ChainId> = BuildLPTxInput<T> & {
+  tokenInMaxs: [cosmossdk.CosmosSDKToken, cosmossdk.CosmosSDKToken]
```
0xApotheosis, line 182: This name feels strange, but I'm assuming this is some osmo vernacular I'm not across 👌

0xApotheosis (approved):

PR #1136, opened 2022-12-16
chore(asset-service): add tr translations for asset overrides

firebomb1:
Adding the Turkish translations for asset overrides.


0xApotheosis (approved):
🍰

PR #1122, opened 2022-12-05
feat(swapper): base unit vernacular

gomesalexandre:
<h3 align="center">Description</h3>

<p align="center">
  
</p>

Plumbing PR bringing the base unit vernacular in many places across lib, making sure we:

1. always name things the correct way, fixing some "precision" named vars that were actually base units
2. always use the vernacular, either for our in-house vars, or using destructuring to re-assign API-returned values to the right vernacular

Note that this is only property name changes (no runtime change), but will require web to accommodate for these (https://github.com/shapeshift/web/pull/3441 as a bigger boy, https://github.com/shapeshift/web/pull/3475 for the time being to make the diff less crazy on the former)

PR #1139, opened 2022-12-21
fix(swapper): cowswap quote rate

0xApotheosis:
Fixes a regression in https://github.com/shapeshift/lib/pull/1122: https://github.com/shapeshift/lib/pull/1122/files#diff-19676c94d85c0baf6d1d359ea9be6657fb3a009be5536ea8c01bdba4d78ec2b6L103-L105.

The regression caused CoW Swap rates to be calculated incorrectly in `getCowSwapTradeQuote`.

Note, this regression is not yet in production - we haven't yet bumped swapper to the next major version in `web`.

Testing:

- Get a CoW Swap quote and confirm that the rate is correct.

Before: 

image

Fixed: 

image


pastaghost (approved):

gomesalexandre (approved):
Cross-tested again prod / linked `main` lib swapper:

#### Prod (NOK)

image


#### main lib (NOK)

image


#### This diff (OK)

image


Note, while testing this diff, I found out 0x quotes are broken in a similar fashion, see:

image

PR #1141, opened 2022-12-22
feat(asset-service): add osmosis lp tokens

pastaghost:
* Generate osmosis LP tokens along with normal osmosis assets
* Update generatedAssetData.json

Example, LP token for Osmosis OSMO/ATOM pool:
```json
  "cosmos:osmosis-1/ibc:gamm/pool/1": {
    "assetId": "cosmos:osmosis-1/ibc:gamm/pool/1",
    "chainId": "cosmos:osmosis-1",
    "symbol": "gamm/pool/1",
    "name": "Osmosis OSMO/ATOM LP Token",
    "precision": 6,
    "color": "#750BBB",
    "icon": "https://assets.coincap.io/assets/icons/osmo@2x.png",
    "explorer": "https://www.mintscan.io/osmosis",
    "explorerAddressLink": "https://www.mintscan.io/osmosis/account/",
    "explorerTxLink": "https://www.mintscan.io/osmosis/txs/"
  },
  ```
  
@0xdef1cafe, regarding the question from last week, unchained does know about these and tracks them properly.

cc: @elmutt @toshiSat

### CODE REVIEW

**Review by 0xApotheosis — COMMENTED** [review]
Looks good to me - re-ran asset generation as a sanity check and observed no diff.

Added some comments to tackle at your discretion, nothing blocking.

One thing I noticed though is that whilst the generated assets appear in the searchable assets in `web`, they have no asset pages (clicking on them just reverts to the Dashboard):

Screenshot 2022-12-22 at 3 01 50 pm

Is this expected? It doesn't seem desirable.

**Comment by pastaghost** [comment]
@0xApotheosis Hmm... it's strange that the asset pages for the Osmosis LP tokens are missing - I'll take a look around and see if I can figure out why that's happening. In the meantime, I've updated the PR with your suggestions from above.
I also noticed that the Osmosis icon from `baseAssets.ts` was of semi-potato quality, so I updated that as well.

**Comment by pastaghost** [comment]
While working on the Osmosis LP integration in web, I realized that it's going to be useful to track the underlying assets for all of the liquidity pool tokens. Instead of constructing a separate map and keeping that in state, I think it's convenient to extend the `Asset` type with an optional `underlyingAssets: AssetId[]` property. Since we already have an AssetId->Asset mapping available via `useAppSelector(state => selectAssetById(state, assetId))`, we only need to store the CAIP IDs of the underlying assets. The `underlyingAssets` array is of variable size, so this will work for N-sided liquidity pool tokens. 

Ex:
```json
  "cosmos:osmosis-1/ibc:gamm/pool/812": {
    "assetId": "cosmos:osmosis-1/ibc:gamm/pool/812",
    "chainId": "cosmos:osmosis-1",
    "symbol": "gamm/pool/812",
    "name": "Osmosis AXL/OSMO LP Token",
    "precision": 6,
    "color": "#750BBB",
    "icon": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.png",
    "explorer": "https://www.mintscan.io/osmosis",
    "explorerAddressLink": "https://www.mintscan.io/osmosis/account/",
    "explorerTxLink": "https://www.mintscan.io/osmosis/txs/",
    "underlyingAssets": [
      "cosmos:osmosis-1/ibc:903A61A498756EA560B85A85132D3AEE21B5DEDD41213725D22ABF276EA6945E",
      "cosmos:osmosis-1/slip44:118"
    ]
  }
  ```

**Comment by 0xdef1cafe** [comment]
> @0xApotheosis While working on the Osmosis LP integration in web, I realized that it's going to be useful to track the underlying assets for all of the liquidity pool tokens. Instead of constructing a separate map and keeping that in state, I think it's convenient to extend the `Asset` type with an optional `underlyingAssets: AssetId[]` property. Since we already have an AssetId->Asset mapping available via `useAppSelector(state => selectAssetById(state, assetId))`, we only need to store the CAIP IDs of the underlying assets. The `underlyingAssets` array is of variable size, so this will work for N-sided liquidity pool tokens. We might also consider tracking the fee asset for LP tokens this way.
> 
> Ex:
> 
> ```json
>   "cosmos:osmosis-1/ibc:gamm/pool/812": {
>     "assetId": "cosmos:osmosis-1/ibc:gamm/pool/812",
>     "chainId": "cosmos:osmosis-1",
>     "symbol": "gamm/pool/812",
>     "name": "Osmosis AXL/OSMO LP Token",
>     "precision": 6,
>     "color": "#750BBB",
>     "icon": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.png",
>     "explorer": "https://www.mintscan.io/osmosis",
>     "explorerAddressLink": "https://www.mintscan.io/osmosis/account/",
>     "explorerTxLink": "https://www.mintscan.io/osmosis/txs/",
>     "underlyingAssets": [
>       "cosmos:osmosis-1/ibc:903A61A498756EA560B85A85132D3AEE21B5DEDD41213725D22ABF276EA6945E",
>       "cosmos:osmosis-1/slip44:118"
>     ]
>   }
> ```

let's chat about this next week, i'm pretty strongly against this at first glance. it turns our "flat" asset data into a "join table" that's aware of LP positions.

we can talk about what's missing/required in the front end LP abstraction that would necessitate an approach like this, but there's almost certainly a different approach that doesn't pollute this type.

@gomesalexandre FYI

**Comment by gomesalexandre** [comment]
> let's chat about this next week, i'm pretty strongly against this at first glance. it turns our "flat" asset data into a "join table" that's aware of LP positions.
> 
> we can talk about what's missing/required in the front end LP abstraction that would necessitate an approach like this, but there's almost certainly a different approach that doesn't pollute this type.
> 
> @gomesalexandre FYI

@pastaghost Agree with @0xdef1cafe - IMO this is effectively a concern of the `opportunitiesSlice` and relevant selectors, whatever DeFi related data for a specific asset we should be able to get with opportunities slice selectors. The `Asset` type should really just give us asset metadata e.g what Coingecko "Info" section would give us for this LP token if it was listed there.

Let's discuss this one out and see what we're missing in the abstraction:
- `feeAsset` exists in `investor` but not in the web abstraction, so definitely something that makes sense to add here
- `underlyingAssetIds` could be made a variable-size array instead of a tuple of 0/1/2/3 currently, if this isn't matching Osmosis LP tokens

**Comment by pastaghost** [comment]
> > let's chat about this next week, i'm pretty strongly against this at first glance. it turns our "flat" asset data into a "join table" that's aware of LP positions.
> > we can talk about what's missing/required in the front end LP abstraction that would necessitate an approach like this, but there's almost certainly a different approach that doesn't pollute this type.
> > @gomesalexandre FYI
> 
> @pastaghost Agree with @0xdef1cafe - IMO this is effectively a concern of the `opportunitiesSlice` and relevant selectors, whatever DeFi related data for a specific asset we should be able to get with opportunities slice selectors. The `Asset` type should really just give us asset metadata e.g what Coingecko "Info" section would give us for this LP token if it was listed there.
> 
> Let's discuss this one out and see what we're missing in the abstraction:
> 
> * `feeAsset` exists in `investor` but not in the web abstraction, so definitely something that makes sense to add here
> * `underlyingAssetIds` could be made a variable-size array instead of a tuple of 0/1/2/3 currently, if this isn't matching Osmosis LP tokens

@gomesalexandre Okay, that sounds like a good strategy to me 👍. I'll revert the last four commits and take a look at mobing the data into opportunitiesSlice.

File: packages/asset-service/src/generateAssetData/baseAssets.ts [hunk 1]
```diff
@@ -87,7 +87,7 @@ export const osmosis: Asset = {
   name: 'Osmosis',
   precision: 6,
   color: '#750BBB',
-  icon: 'https://assets.coincap.io/assets/icons/osmo@2x.png',
+  icon: 'https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.png',
```
0xdef1cafe, line 90: nit pick - we can use `githack` here to avoid `429`s from `githubusercontent`- grep for examples

gomesalexandre (dismissed):
LGTM, the diffed `generatedAssetData` is looking good.

Ran `yarn generate` as a sanity check and confirmed we have no diff here

packages/asset-service/src/generateAssetData/baseAssets.ts
pastaghost: 👍

pastaghost: 👍

gomesalexandre (dismissed):

packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -92,8 +98,49 @@ export const getAssets = async (): Promise<Asset[]> => {
...
+
+    const getLPTokenName = (asset1: string, asset2: string): string =>
+      `Osmosis ${asset1}/${asset2} LP Token`
+
+    if (current.pools) {
+      for (const [pairedToken, poolId] of Object.entries(current.pools)) {
+        if (lpTokenAlreadyAdded(poolId)) continue
+
+        const lpAssetDatum: Asset = {
+          assetId: `cosmos:osmosis-1/ibc:gamm/pool/${poolId}`,
```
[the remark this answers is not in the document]
pastaghost: 👍

0xdef1cafe (approved):

PR #1142, opened 2022-12-22
feat(caip): add osmosis lp tokens to adapter

pastaghost:
Enables caip parsing for osmosis lp tokens. 

Note: `parseAssetIdRegExp` was updated to allow for `'/'` characters in the fourth capture group. Eg: `/gamm/pool/1`

cc @elmutt @toshiSat


0xdef1cafe (approved):
unable to test asset generation script with my current connection but conceptually LGTM

PR #1147, opened 2023-01-04
fix: only inject accountNumber, not full bip44params

0xdef1cafe:
**please scrutinize thoroughly, turbo danger here as this relates to address generation/possible loss of funds!**

accompanying web PR incoming before this is merged, as i've removed `buildBIP44Params` in favor of `getBIP44Params` with validation and there's a trivial number of places to change in web.

this PR prevents us from generating "incorrect", or at least non-standard addresses by only passing in `accountNumber`, rather than arbitrary, unvalidated `BIP44Params` into chain adapters.

there are two bulky commits, the first passed type checking of application code, the second fixed type errors and made tests pass. i didn't make any significant changes to test references data, other than swapping bip params for account numbers - once test logic was updated correctly the same addresses were generated (and the tests in this case were actually super useful for once).


packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -130,10 +130,6 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
     return this.coinName
   }
 
-  buildBIP44Params(params: Partial<BIP44Params>): BIP44Params {
```
0xdef1cafe, line 133: remove this, as it doesn't have any validation, and having both `build` and `get` is a bit derp

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -130,10 +130,6 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
...
-    return { ...this.defaultBIP44Params, ...params }
-  }
-
   getBIP44Params({ accountNumber, accountType }: GetBIP44ParamsInput): BIP44Params {
```
0xdef1cafe, line 133: this one is the smarter one and we can use this everywhere

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
0xdef1cafe: this is only used in 3 places in web so it's trivial to swap out

packages/chain-adapters/src/types.ts
```diff
@@ -83,13 +83,15 @@ export type FeeDataEstimate<T extends ChainId> = {
 
 export type SubscribeTxsInput = {
   wallet: HDWallet
-  bip44Params: BIP44Params
+  accountNumber: number
```
0xdef1cafe, line 86: the core of this PR, everything else hangs off this

now that we only *require* `accountNumber`, with optional `index` and `isChange` params (for UTXO chains) below, we use the `getBIP44Params` utility function of the respective base chain adapter to generate the correct `BIP44Params`, derived from the safe defaults defined on the concrete implementation of the adapter, plus whatever we pass in, which is also validated.

this prevents us from passing consuming clients (web) from passing non-standard `BIP44Params`, specifically `purpose` and `coinType` into a chain adapter, and generating non-standard addresses, leading to non-trivially recoverable funds

packages/chain-adapters/src/api.ts
```diff
@@ -50,8 +50,6 @@ export type ChainAdapter<T extends ChainId> = {
    */
   getAccount(pubkey: string): Promise<Account<T>>
 
-  buildBIP44Params(params: Partial<BIP44Params>): BIP44Params
```
0xdef1cafe, line 53: this is deleted in favor of more robust `getBIP44Params`

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
0xdef1cafe: ok 22 but that's basically 3

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -130,14 +130,20 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
...
 
-  buildBIP44Params(params: Partial<BIP44Params>): BIP44Params {
-    return { ...this.defaultBIP44Params, ...params }
-  }
-
-  getBIP44Params({ accountNumber, accountType }: GetBIP44ParamsInput): BIP44Params {
+  getBIP44Params({
+    accountNumber,
+    accountType,
+    index = 0,
```
gomesalexandre, line 136: 🧠

packages/chain-adapters/src/ChainAdapterCLI.ts
```diff
@@ -98,18 +98,12 @@ const getWallet = async (): Promise<NativeHDWallet> => {
 // @ts-ignore:nextLine
 const testBitcoin = async (wallet: NativeHDWallet, broadcast = false) => {
   const chainAdapter = adapters.btc
-  const bip44Params: BIP44Params = {
-    purpose: 84,
-    coinType: 0,
-    accountNumber: 0,
-    isChange: false,
-    index: 10,
```
gomesalexandre, line 106: Not sure why the previous implementation was using the 10th address index, this achieves a similar testing without it (with a different address obviously)

packages/chain-adapters/src/utxo/dogecoin/DogecoinChainAdapter.test.ts
```diff
@@ -459,12 +432,19 @@ describe('DogecoinChainAdapter', () => {
     })
     it('should always be coinType 3', async () => {
       for (const accountType of adapter.getSupportedAccountTypes()) {
-        const r = adapter.getBIP44Params({ accountNumber: 0, accountType })
+        const r = adapter.getBIP44Params({
+          accountNumber: 0,
+          accountType,
+          isChange: false,
+          index: 0,
```
gomesalexandre, line 439: Do we want to be implicit for consistency?

gomesalexandre (commented):
Provisional stamp pending extensive testing with its matey web PR.

- All `buildBip44Params` occurrences have been removed, including in tests' adapter mocks

image

- `adapter.getAddress()` calls now only use `accountNumber`, not full `bip44Params` (confirmed by tsc passing)

-  `account'` / `change` / `address_index` injection is looking consistent with the previous  tests

Also pushed 88c96c086dd3cdb02d73b1d5c8901642a6a8ad0e - noticed this isn't used in lib and only in [web](https://github.com/shapeshift/web/blob/e683d53ff01f07522a03583529b2ee0d4a4e3ad6/src/lib/account/utxo.ts#L28), thinking we could remove it in favor of calling `getBip44Params()` to get the BIP44Params, and `accountTypeToScriptType()` to get the `scriptType` instead?

gomesalexandre (dismissed):
Tested with the matching web PR:
- EVM chains and Cosmos SDK account rows show no regressions vs prod (wasn't able to test Thorchain, see notes), tested:
  - Assets rows
  - Addresses
- EVM chains and Cosmos SDK account pages show no regressions vs. prod, tested:
  - Tx history
  - Balances/charts
- UTXOs account/s pages are showing no regression, tested:
 - Derived 0/1/2 xpubs are looking the same for all account types
 - Tx history
 - Balances/charts
 
Note, I wasn't able to test Thorchain as this code path is never reached:
 
 https://github.com/shapeshift/lib/blob/2b49262c062d5475549917a52d277ff50d1fb3db/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts#L158
 
 However, I'm pretty sure this is lib linking shenanigans, seen this one previously and was automagically solved

packages/chain-adapters/src/utxo/dogecoin/DogecoinChainAdapter.test.ts
0xdef1cafe: done ty

gomesalexandre (dismissed):

gomesalexandre (approved):
Confirmed hiuge conflict fixed

PR #1148, opened 2023-01-04
feat(chain-adapters:UtxoBaseAdapter): allow UTXOs selection from `from` param

gomesalexandre:
#### Description

- [x] Tests
- [x] Partial sends
- [ ] ~~Send max - do we want to handle this for consistency? Given our use-case for savers, we shouldn't need this~~

This PR brings a new `from` input param in `buildSendTransaction` in `getFeeData` and `buildSendTransaction`, being passed down as a new input param in `utxoSelect` and allowing us to filter UTXOs to only the ones for the passed address.

This effectively allows us to ensure the right UTXO inputs are used when building Txs for Thorchain savers.

Tested by monkey-patching a `from` param in `buildSendTransaction`, effectively mocking sending to the address / pubkey (*NOT* an xpub) that was previously used as an output to deposit into savers.

The current philosophy of this added logic is to "let it crash" in case the coinselect algorithm fails to create an output and/or the broadcast fails, so that web can handle the error, do a regular send (without a `from` param) to the original deposit address to create a valid UTXO, and build/broadcast a Tx again.

#### Testing

- View your UTXOs in `https://blockchair.com/litecoin/outputs?q=recipient(<address>),is_spent(false)#` or by setting a breakpoint in https://github.com/shapeshift/lib/blob/ce34273b11c481665b6f28bd5db0171be3eda210/packages/chain-adapters/src/utxo/utxoSelect/index.ts#L23
- In web, call `buildSendTransaction` with a `from` address that will produce valid outputs
- Ensure the Tx is broadcasted

#### Screenshots

image


gomesalexandre:
* main
  * **PR #1148** <a href="https://app.graphite.dev/github/pr/shapeshift/lib/1148" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/lib/1148?utm_source=stack-comment).

packages/chain-adapters/src/types.ts
```diff
@@ -131,6 +131,7 @@ type ChainSignTx = {
 export type SignTx<T extends ChainId> = T extends keyof ChainSignTx ? ChainSignTx[T] : never
 
 export type BuildSendTxInput<T extends ChainId> = {
+  from?: string
```
kaladinlight, line 134: only other thought would be to move from to the `chainSpecific` section since it isn't relevant for evm/cosmossdk chains

gomesalexandre: Ab-solutely, much better umbrella - interestingly `getFeeData` for EVM chains also takes a `chainSpecific.from` due to the way ERC20s send maxes gas estimation work, so that will keep things consistent!

99ce9421753d1fb2d88deced3dceea09c0a62d2e

packages/chain-adapters/src/utxo/utxoSelect/index.ts
```diff
@@ -17,7 +18,11 @@ export type UTXOSelectInput = {
  * _opReturnData is filtered out of the return payload as it is added during transaction signing_
  */
 export const utxoSelect = (input: UTXOSelectInput) => {
-  const utxos = input.utxos.map((x) => ({ ...x, value: Number(x.value) }))
+  let utxos = input.utxos.map((x) => ({ ...x, value: Number(x.value) }))
+
+  if (input.from) {
+    utxos = utxos.filter((utxo) => utxo.address === input.from)
+  }
```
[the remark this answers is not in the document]
gomesalexandre: ser you're getting too functional, have you heard of monoids in the category of endofunctors?

(93097e9a5f7adc9337e2039394b3eeef3b5c8061)

packages/chain-adapters/src/utxo/utxoSelect/index.ts
```diff
@@ -4,20 +4,31 @@ import split from 'coinselect/split'
...
 export const utxoSelect = (input: UTXOSelectInput) => {
-  const utxos = input.utxos.map((x) => ({ ...x, value: Number(x.value) }))
+  const utxos = input.utxos.reduce((acc, utxo) => {
+    const sanitizedUtxo = { ...utxo, value: Number(utxo.value) }
+
+    // If input contains a `from` param, the intent is to only keep the UTXOs from that address
+    // so we can ensure the send address is the one we want
+    // This doesn't do any further checks, so error-handling should be done by the caller e.g `buildSendTransaction` callsites
+    return [...acc, ...(input.from && utxo.address !== input.from ? [] : [sanitizedUtxo])]
+  }, [] as SanitizedUTXO[])
```
0xdef1cafe, line 30:
```suggestion
  const utxos = input.utxos.reduce<SanitizedUTXO[]>((acc, utxo) => {
    const sanitizedUtxo = { ...utxo, value: Number(utxo.value) }

    // If input contains a `from` param, the intent is to only keep the UTXOs from that address
    // so we can ensure the send address is the one we want
    // This doesn't do any further checks, so error-handling should be done by the caller e.g `buildSendTransaction` callsites
    return [...acc, ...(input.from && utxo.address !== input.from ? [] : [sanitizedUtxo])]
  }, [])
```

packages/chain-adapters/src/utxo/utxoSelect/index.test.ts
```diff
@@ -158,4 +178,43 @@ describe('utxoSelect', () => {
     }
     expect(result).toEqual(expectedResult)
   })
+  it('with passed `from` param in input - should return fee and no inputs/outputs in case there is no UTXO for the passed from address', () => {
```
0xdef1cafe, line 181: we probably need one or two additional test cases to ensure that we can send the correct `opReturnData` (i.e. the deposit/withdraw memo) in the outputs too.

packages/chain-adapters/src/utxo/utxoSelect/index.ts
gomesalexandre: 474ea37ad955734325c5452d4d6b53b6be6f5c76

packages/chain-adapters/src/utxo/utxoSelect/index.test.ts
```diff
@@ -1,29 +1,49 @@
...
       confirmations: 2,
-      address: 'bc1qpszctuml70ulzf7f0zy5r4sg9nm65qfpgcw0uy',
+      address: '15uud35JNGQboswqGeGGhTttGcX413VGcw',
       path: "m/84'/0'/0'/0/1",
     },
   ],
   to: 'bc1qppzsgs9pt63cx9x994wf4e3qrpta0nm6htk9v4',
   satoshiPerByte: '1',
-  value: '400',
+  value: '35559',
```
0xdef1cafe, line 46: we want this to be smaller than the values above (`1598` and `31961`) such that we can ensure that this has more valid utxo's to select from and the `from` filter is working correctly, say `1000` should work

packages/chain-adapters/src/utxo/utxoSelect/index.test.ts
0xdef1cafe: disregard this, `opReturnData` is handled by hdwallet signing the tx

packages/chain-adapters/src/utxo/utxoSelect/index.ts
```diff
@@ -4,20 +4,31 @@ import split from 'coinselect/split'
...
  */
 export const utxoSelect = (input: UTXOSelectInput) => {
-  const utxos = input.utxos.map((x) => ({ ...x, value: Number(x.value) }))
+  const utxos = input.utxos.reduce<SanitizedUTXO[]>((acc, utxo) => {
+    const sanitizedUtxo = { ...utxo, value: Number(utxo.value) }
+
+    // If input contains a `from` param, the intent is to only keep the UTXOs from that address
+    // so we can ensure the send address is the one we want
+    // This doesn't do any further checks, so error-handling should be done by the caller e.g `buildSendTransaction` callsites
+    return [...acc, ...(input.from && utxo.address !== input.from ? [] : [sanitizedUtxo])]
```
0xdef1cafe, line 29: this can actually be a decent amount of data and is on the hot path, may want to `acc.push(...` here

0xdef1cafe:
get in

PR #1151, opened 2023-01-05
perf(swapper): add caching to swapper requests

0xApotheosis:
- Adds caching to select swapper-related `GET` requests by including a cache adapter in 0x, THORSwapper, and CoW Swap services.
- Takes an opt-in approach to caching - we want to be explicit about what we cache, as some responses _must_ be current (order status)
- We use a max age of 5 seconds as a minimum bound for effectiveness, but not so large that we get stale data for crucial responses (e.g. THORSwapper inbound addresses)
- Reduces swapper-related requests by about 60%.

Contributes to https://github.com/shapeshift/web/issues/3466

### DIFF
--- a/packages/swapper/package.json
+++ b/packages/swapper/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "axios": "^0.26.1",
+    "axios-cache-adapter": "^2.7.3",
     "bignumber.js": "^9.0.2",
     "lodash": "^4.17.21",
     "retry-axios": "^2.6.0",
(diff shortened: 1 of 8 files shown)


packages/swapper/src/utils.ts
```diff
@@ -1,8 +1,23 @@
...
     throw new AssertionError({ message: `Expected 'val' to be defined, but received ${val}` })
   }
 }
+
+const getRequestFilter = (cachedUrls: string[]) => (request: Request) =>
+  !cachedUrls.some((url) => request.url.includes(url))
+
+export const createCache = (maxAge: number, cachedUrls: string[]): ISetupCache => {
+  const filter = getRequestFilter(cachedUrls)
+  return setupCache({
```
0xApotheosis, line 16: Add `debug: true` to this object to get console logs of what's going on.

packages/swapper/package.json
```diff
@@ -44,9 +45,9 @@
     "@shapeshiftoss/errors": "^1.1.2",
     "@shapeshiftoss/hdwallet-core": "^1.41.0",
     "@shapeshiftoss/types": "^8.1.0",
+    "@shapeshiftoss/unchained-client": "^10.0.1",
```
gomesalexandre, line 48: 💜

gomesalexandre (approved):
Amazing network requests improvement ser.
Linked locally and regression tested 0x, Thor, and CoW swaps, both by waiting less/more than 5 seconds between re-inputting amounts, LGTM!

Confirmed with the `{debug: true}` option the requests are properly cached
image

PR #1155, opened 2023-01-06
fix: utxo base adapter get address

0xdef1cafe:
defaulting to `index = 0` in `getBIP44Params` was a mistake - the `getAddress` function expects undefined as a valid input, which it uses to grab the next receive address. by explicitly passing 0, we were generating a valid address for index 0, but always index 0, because of `getBIP44Params` defaulting to 0 rather than undefined.

this is not an issue for evm or cosmossdk adapters, as while technically we can generate addresses with `index > 0`, this isn't a normal convention.

- fix: don't default index to 0 for utxo base adapter
- test: update tests to mock next address index correctly

### DIFF
--- a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
+++ b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
@@ -140,7 +140,7 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
       throw new Error('accountNumber must be >= 0')
     }
 
-    if (index && index < 0) {
+    if (index !== undefined && index < 0) {
       throw new Error('index must be >= 0')
     }
 
--- a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
+++ b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
@@ -133,14 +133,14 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
   getBIP44Params({
     accountNumber,
     accountType,
-    index = 0,
+    index,
     isChange = false,
   }: GetBIP44ParamsInput): BIP44Params {
     if (accountNumber < 0) {
       throw new Error('accountNumber must be >= 0')
     }
 
-    if (index < 0) {
+    if (index && index < 0) {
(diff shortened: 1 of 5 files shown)


packages/chain-adapters/src/utxo/bitcoin/BitcoinChainAdapter.test.ts
```diff
@@ -339,11 +339,22 @@ describe('BitcoinChainAdapter', () => {
...
+      args.providers.http = {
+        getAccount: jest.fn().mockResolvedValue(getAccountMockResponse),
+      } as any
+    })
     it("should return a p2pkh address for valid derivation root path parameters (m/44'/0'/0'/0/0)", async () => {
       const wallet: HDWallet = await getWallet()
       const adapter = new bitcoin.ChainAdapter(args)
       const accountNumber = 0
 
+      args.providers.http.getAccount = jest.fn().mockResolvedValue({
```
kaladinlight, line 353: shouldn't need this is we pass in explicit `index = 0`

kaladinlight (approved):

gomesalexandre (approved):
Tested locally, confirmed UTXO receive addresses are matching the ones on prod for both account 0 and 1+

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -133,14 +133,14 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
...
-    index = 0,
+    index,
     isChange = false,
   }: GetBIP44ParamsInput): BIP44Params {
     if (accountNumber < 0) {
       throw new Error('accountNumber must be >= 0')
     }
 
-    if (index < 0) {
+    if (index !== undefined && index < 0) {
```
0xdef1cafe, line 143: @kaladinlight this was the only logical change in the last commit with updated tests - `0` is falsy but a valid index obv

kaladinlight: ah yes, better to be explicit. would have still worked coincidentally, but this is the right call

kaladinlight (approved):

gomesalexandre (approved):

PR #1158, opened 2023-01-10
feat(swapper): export `getInboundAddressDataForChain`

0xApotheosis:
- Exports `getInboundAddressDataForChain` for consumption in `web`, whereby a hook is used to check for pool halts if THORSwapper is used
- Adds a new flag, `excludeHalted`, to `getInboundAddressDataForChain`, allowing `web` to receive all pools and explicitly check if a pool is halted (enables a fail-closed approach in `web`)

Contributes to https://github.com/shapeshift/web/issues/2925


gomesalexandre (approved):
LGTM ✅

PR #1159, opened 2023-01-11
fix: use cosmos sdk account number in osmo buildTradeTx

0xdef1cafe:
this was my mistake in a recent refactor - the Osmosis `account_number` that is returned from the `getAccount` call is *not* the same as the `accountNumber` that is used to construct `BIP44Params`

this PR reverts to using the correct `account_number` when building the trade tx, as per @elmutt @toshisat


elmutt (approved):

0xApotheosis (approved):
This fixes _a_ bug, but not the "Invalid trade pair" bug.

PR #1160, opened 2023-01-11
fix: cosmos account number

elmutt:
use correct ibcAccount number


0xdef1cafe (approved):
Untested but looks sane

PR #1161, opened 2023-01-11
fix(swapper): cache osmo pools endpoint

elmutt:
Cache osmo pools data because its called a LOT


0xApotheosis (approved):
Nice one, the limit of 10 seems like a good call.

PR #1165, opened 2023-01-14
fix(swapper): osmo swapper cached url, small refactor

pastaghost:
* moves the osmosis swapper axios cache configuration to a separate file to match the pattern used in other swappers
* removes the trailing '/' from the osmoService cached URL so that calls to the `/gamm/v1beta1/pools` endpoint are cached properly


0xApotheosis (approved):
Nice refactor, cheers señor.

Tested locally, the removal of the `/` has a noticeable impact on the number of calls made 👌

PR #1166, opened 2023-01-15
fix: having / on end was negating cache banefits

elmutt:
Fixes a small mixtake with osmo pools caching. 

Nothing was "broken" but we werent getting cache benefits


0xApotheosis:
@pastaghost already fixed this in https://github.com/shapeshift/lib/pull/1165.

PR #1171, opened 2023-01-22
fix(asset-service): generate osmosis lp asset data

pastaghost:
* fixes osmosis LP token generation
* regenerated `generatedAssetData.json`
* use `rawcdn.githack.com` urls instead of `raw.githubusercontent.com` urls for osmosis icons
* update osmosis icon urls every time assetdata is regenerated to avoid using stale image data cached by githack

* verified that the number of osmosis lp tokens previously generated (110) is similar to the number of pool tokens generated after this fix (118)

### DIFF
--- a/packages/asset-service/src/generateAssetData/osmosis/index.ts
+++ b/packages/asset-service/src/generateAssetData/osmosis/index.ts
@@ -56,7 +56,7 @@ export const getAssets = async (): Promise<Asset[]> => {
   */
   const lpAssetsAdded = new Set()
 
-  return await assetData.assets.reduce<Promise<Asset[]>>(async (accPrevious, current) => {
+  return assetData.assets.reduce<Promise<Asset[]>>(async (accPrevious, current) => {
     const acc = await accPrevious
     if (!current) return acc
 
--- a/packages/asset-service/src/generateAssetData/osmosis/index.ts
+++ b/packages/asset-service/src/generateAssetData/osmosis/index.ts
@@ -1,6 +1,5 @@
 import { ASSET_REFERENCE, osmosisChainId, toAssetId } from '@shapeshiftoss/caip'
 import axios from 'axios'
-import memoize from 'lodash/memoize'
 
 import { Asset } from '../../service/AssetService'
 import { getRenderedIdenticonBase64, IdenticonOptions } from '../../service/GenerateAssetIcon'
@@ -43,33 +42,10 @@ type OsmosisAssetList = {
 }
 
 export const getAssets = async (): Promise<Asset[]> => {
-  /** Helper function to get latest commit hash for default branch of given repo.
-   * This is used to construct githack URLs */
-  const getCurrentHash = memoize(async (repository: string): Promise<string> => {
-    /* Trim any trailing slashes */
-    while (repository.charAt(repository.length - 1) === '/') {
-      repository = repository.slice(0, -1)
-    }
-
-    const {
-      data: { default_branch: defaultBranch },
-    } = await axios.get(`https://api.github.com/repos/${repository}`)
-
-    const {
-      data: {
-        object: { sha: currentHash },
-      },
-    } = await axios.get(
-      `https://api.github.com/repos/${repository}/git/refs/heads/${defaultBranch}`,
-    )
-
-    return currentHash
-  })
-
   /* Fetch asset list */
-  const currentAssetListsHash = await getCurrentHash('osmosis-labs/assetlists')
+
   const { data: assetData } = await axios.get<OsmosisAssetList>(
-    `https://raw.githack.com/osmosis-labs/assetlists/${currentAssetListsHash}/osmosis-1/osmosis-1.assetlist.json`,
+    `https://rawcdn.githack.com/osmosis-labs/assetlists/main/osmosis-1/osmosis-1.assetlist.json`,
   )
 
   if (!assetData) throw new Error('Could not get Osmosis asset data!')
@@ -105,16 +81,6 @@ export const getAssets = async (): Promise<Asset[]> => {
 
     const assetId = `cosmos:osmosis-1/${assetNamespace}:${assetReference}`
 
-    const logoURI = await (async (): Promise<string> => {
-      let uri = current.logo_URIs.png ?? current.logo_URIs.svg ?? ''
-      if (uri) {
-        uri = uri
-          .replace('raw.githubusercontent.com', 'rawcdn.githack.com')
-          .replace('master', await getCurrentHash('cosmos/chain-registry'))
-      }
-      return uri
-    })()
-
     const assetDatum: Asset = {
       assetId,
       chainId: osmosisChainId,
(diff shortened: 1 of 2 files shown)


packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -17,33 +18,72 @@ type OsmoAsset = {
...
   coingecko_id: string
-  pools?: {
-    [key: string]: number
-  }
+  keywords: string[]
 }
 
 type OsmosisAssetList = {
-  chain_id: string
+  chain_name: string
```
cjthompson, line 41: I may be out of the loop but why this name change?

pastaghost: @cjthompson Osmosis changed the schema upstream. See [here](https://github.com/osmosis-labs/assetlists)

packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -17,33 +18,72 @@ type OsmoAsset = {
...
   assets: OsmoAsset[]
 }
 
 export const getAssets = async (): Promise<Asset[]> => {
+  /** Helper function to get latest commit hash for default branch of given repo.
+   * This is used to construct githack URLs */
+  const getCurrentHash = memoize(async (repository: string): Promise<string> => {
```
gomesalexandre, line 48: Curious why do we need this?
https://raw.githack.com/osmosis-labs/assetlists/main/osmosis-1/osmosis-1.assetlist.json works similarly, unless the default branch ever changes out of main upstream, which seems like a very unlikely scenario to me, we are good without a commit hash

packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -17,33 +18,72 @@ type OsmoAsset = {
...
+  console.info('assetData', assetData)
+
   /* Osmosis pool IDs are guaranteed to be unique integers, so we use a set to keep track of 
     which pools we've already seen. A lookup is necessary because the Osmosis asset list 
     contains duplicate entries for each pool, eg. ATOM/OSMO == OSMO/ATOM. 
   */
   const lpAssetsAdded = new Set()
 
-  return assetData.assets.reduce<Asset[]>((acc, current) => {
+  return await assetData.assets.reduce<Promise<Asset[]>>(async (accPrevious, current) => {
```
gomesalexandre, line 85: `return await somePromise` is effectively `return somePromise`, since you get a promise out of `getAssets` either way
```suggestion
  return assetData.assets.reduce<Promise<Asset[]>>(async (accPrevious, current) => {
```

packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -67,14 +107,24 @@ export const getAssets = async (): Promise<Asset[]> => {
 
     const assetId = `cosmos:osmosis-1/${assetNamespace}:${assetReference}`
 
+    const logoURI = await (async (): Promise<string> => {
+      let uri = current.logo_URIs.png ?? current.logo_URIs.svg ?? ''
+      if (uri) {
+        uri = uri
+          .replace('raw.githubusercontent.com', 'rawcdn.githack.com')
+          .replace('master', await getCurrentHash('cosmos/chain-registry'))
```
gomesalexandre, line 115: Ditto rationale?

packages/asset-service/src/generateAssetData/osmosis/index.ts
```diff
@@ -17,33 +18,72 @@ type OsmoAsset = {
...
+   * This is used to construct githack URLs */
+  const getCurrentHash = memoize(async (repository: string): Promise<string> => {
+    /* Trim any trailing slashes */
+    while (repository.charAt(repository.length - 1) === '/') {
+      repository = repository.slice(0, -1)
+    }
+
+    const {
+      data: { default_branch: defaultBranch },
+    } = await axios.get(`https://api.github.com/repos/${repository}`)
```
gomesalexandre, line 56: Also note in case we keep this function that this will fire XHRs to GitHub API unauthenticated. Not an issue here given the fact that we only run `generateAssetData` so few times, but generally a better idea to assume `gh` is installed and spawn a gh process instead

gomesalexandre (commented):
`generatedAssetData` looks good - added some comments WRT the commit hash fetching rationale, conceptual stamp until these are answered

packages/asset-service/src/generateAssetData/osmosis/index.ts
pastaghost: This serves two purposes. First, [GitHack FAQ](https://raw.githack.com/faq) instructs that we should be using CDN URLs (repo/hash/file) here, and we're violating the GitHack TOS by using development URLs (repo/branch/file) in a production application. Second, fetching and generating a CDN URL from the latest default branch commit hash ensures that the asset list is always up-to-date.

See more [here](https://raw.githack.com/).

packages/asset-service/src/generateAssetData/osmosis/index.ts
pastaghost: Same rationale as above.

packages/asset-service/src/generateAssetData/osmosis/index.ts
gomesalexandre: Still unsure about this.

According to the docs: ` To ensure that the CDN always serves the version of the file you want, use a git tag or commit ref in the file's path instead of a branch name, and update the URL if you push a new version of the file.`

Let's say you run the generation asset twice in a day and the latest upstream commit hash didn't change, the `master` blob for said file would still be the same.
Realistically, you'd only need to run this when the file actually changed, so using a commit hash or a branch name would produce the same effect.

Now the comment WRT using an actual CDN URL *is* valid:

image

However, we're still using `raw.githack.com` (dev) vs . `rawcdn.githack.com` (prod) here?

https://github.com/shapeshift/lib/blob/bb16508cbb7d05595d3df771f1707d9084aafdb5/packages/asset-service/src/generateAssetData/osmosis/index.ts#L73

Either way, given this is a run once, generate, forget and repeat a few weeks later kind of script, caching/rate-limiting is a non-issue for us.

If we want to add a safety factor, happy with changing to the prod URL, but I stand with the fact using a commithash is over-engineering for our use-case, let's not clutter the PR with it.

pastaghost: @gomesalexandre Yeah, that's right. Using the branch name or the commit hash is functionally the same in that case. The reason I'm fetching the latest hash from Github instead of hardcoding a fixed commit hash is to be able to use CDN URLs and simultaneously make sure that we have the latest file contents.

Nice catch with the `raw.githack` vs r`awcdn.githack` typo! I'll fix that now.

I'm not worried about caching/rate limiting on our end with the asset list - it's the icon URLs that end up in `generatedAssetList.json` and baked into the production web app that I'm worried about. I did this to avoid any issues with clients being unable to fetch asset icons because of those URLs being rate-limited. Since I had that in place for the icons, I thought I might as well just slap it onto the asset list URL as well.

gomesalexandre (approved):
LGTM, tested `yarn generate` in asset-service and there was no Osmo assets diff.
There was a diff on a few Ethereum assets, but absolutely fine to leave this for a follow-up the next time we regenerat assetData.

PR #1172, opened 2023-01-22
fix(asset-service): use recommended githack url format for osmosis urls

pastaghost:
* use `rawcdn.githack.com` + latest commit hash in github static data requests to avoid rate limiting and ensure data freshness

### DIFF
--- a/packages/asset-service/src/generateAssetData/osmosis/index.ts
+++ b/packages/asset-service/src/generateAssetData/osmosis/index.ts
@@ -1,5 +1,6 @@
 import { ASSET_REFERENCE, osmosisChainId, toAssetId } from '@shapeshiftoss/caip'
 import axios from 'axios'
+import memoize from 'lodash/memoize'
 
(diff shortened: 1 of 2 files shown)


gomesalexandre (commented):
Conceptually happy with this, @shapeshift/owners FYI I've discussed this one out with @pastaghost and the rationale is:
- using the pro URL vs. the dev URL, which isn't rate-limited and will "always" (with caveats, see below) serve the freshest URL to clients. As discussed, the rate-limiting is a non-problem we're solving given images are cached on the client, but can't argue using the pro URL is a better practice
- when using the pro URL, responses are cached for a year, thus using `main` branch as a commit-ish effectively means we wouldn't get any fresh data (e.g images updated)

If we're all happy with this one, happy to merge it and propagating the practice of using the pro URL with an actual revision across asset-service

kaladinlight (approved):

gomesalexandre (approved):

PR #1173, opened 2023-01-23
feat(market-service): fix osmosis market-service bug, remove market-service cli

pastaghost:
This is necessary to provide market data for Osmosis network assets.

* adds an optional `fetchFromAllProviders` parameter to `marketService.fetchAll()`. If this is set to `true`, the union of the market data sets from all supported providers (only CoinGecko and Osmosis are used) is returned. If `fetchAllProviders` is not set to `true`, the original behavior is maintained.
* adds tests for `marketService.fetchAll()` to `marketDataCLI.ts`
* fixes a table lookup bug in the Osmosis market data provider
* fixes a test broken by the same bug above


packages/market-service/src/market-service-manager.ts
```diff
@@ -59,16 +59,59 @@ export class MarketServiceManager {
...
+        CoinCapMarketService, // Response is sometimes slow (~500ms), (128 results returned, 0 unique)
+        YearnTokenMarketCapService, // API errors, response is very slow (~4000ms) (208 results returned, 107 unique)
+        YearnVaultMarketCapService, // API errors, response is slow (~5500ms) (107 results returned, 107 unique)
+        IdleMarketService, // (0 results returned)
+        FoxyMarketService, // (0 results returned)
+      ]
+
+      const marketProviders = this.marketProviders.filter((provider) =>
+        blacklistedProviders.every((blacklisted) => !(provider instanceof blacklisted)),
+      )
```
gomesalexandre, line 76: Can be reduced to a single iteration?

packages/market-service/src/market-service-manager.ts
```diff
@@ -59,16 +59,59 @@ export class MarketServiceManager {
...
+       * from all resolved promises, Results from more reliable providers will be prioritized over
+       * identical results from less reliable providers.
+       */
+      result = {}
+      const promises = []
+      for (let i = 0; i < marketProviders.length; i++) {
+        promises.push(marketProviders[i].findAll(args))
+      }
+      const allResults = (await Promise.allSettled(promises))
+        .filter((res) => res.status === 'fulfilled' && res.value)
```
gomesalexandre, line 88: 🧠

packages/market-service/sample.env
```diff
@@ -0,0 +1,4 @@
+COINGECKO_API_KEY='CG-we5Z4KbdYMCgMUVAqDjQMWfc'
+ETHEREUM_NODE_URL='https://dev-daemon.ethereum.shapeshift.com'
+UNCHAINED_ETHEREUM_HTTP_URL='https://dev-api.ethereum.shapeshift.com'
+UNCHAINED_ETHEREUM_WS_URL='wss://dev-api.ethereum.shapeshift.com'
```
gomesalexandre, line 4: Do we need these for this PR? These are stemming from old implementations, if you look at e.g types of `FoxyMarketService` instantiation, you will notice `providerUrls` which we pass from web, which is now how we pass env vars:

https://github.com/shapeshift/lib/blob/505516c23b582ff1cd2d684b6c84a117157e3423/packages/market-service/src/foxy/foxy.ts#L21

packages/market-service/src/marketDataCLI.ts
```diff
@@ -1,13 +1,16 @@
...
 // import { HistoryTimeframe } from '@shapeshiftoss/types'
 // import { Yearn } from '@yfi/sdk'
 // import { CoinGeckoMarketService } from './coingecko/coingecko'
 // import { YearnVaultMarketCapService } from './yearn/yearn-vaults'
 // import { YearnTokenMarketCapService } from './yearn/yearn-tokens'
 // import { OsmosisMarketService } from './osmosis/osmosis'
-import { FOXY_ASSET_ID, FoxyMarketService } from './foxy/foxy'
+// import { FOXY_ASSET_ID, FoxyMarketService } from './foxy/foxy'
+// import { MarketServiceManager } from './market-service-manager'
+dotenv.config()
```
gomesalexandre, line 13: Ditto we shouldn't need this

packages/market-service/src/market-service-manager.ts
```diff
@@ -59,16 +59,59 @@ export class MarketServiceManager {
...
+      for (let i = 0; i < allResults.length; i++) {
+        for (const [k, v] of Object.entries(allResults[i])) {
+          result[k] = result[k] ?? v
+        }
+      }
+      /* Make sure results are sorted by market cap */
+      result = Object.fromEntries(
+        Object.entries(result).sort(([, a], [, b]) => (a.marketCap < b.marketCap ? 1 : -1)),
+      )
+    } else {
```
gomesalexandre, line 101: Nitpick, I'd probably keep this branch first and early return vs. an if/else but purely stylistic

gomesalexandre (commented):
From a maintenance and regressions perspective, LGTM. Tested for regressions in market data and didn't find any. Added some comments WRT env vars in lib

packages/market-service/src/marketDataCLI.ts
kaladinlight: also, how useful is this file in general? would just remove unless we add a reasonable working default case or something...

gomesalexandre: Yeah, looks like we're not really using the CLIs anymore at this point, maybe it be the market-data one or any of the CLI. The only time I remember using the CLI was during ATOM dev, when I used it for depositing / withdrawing, all of which could be done by e.g importing your seed in Keplr

pastaghost: FWIW, I found the CLI to be extremely useful for quick testing in both this PR and #1176. 
At the same time, anything here can be done with Jest, and not having the CLI script available to use during development motivates writing and maintaining proper tests.

packages/market-service/src/market-service-manager.ts
```diff
@@ -58,16 +58,62 @@ export class MarketServiceManager {
...
   async findAll(args: FindAllMarketArgs): Promise<MarketCapResult> {
     let result: MarketCapResult | null = null
-    // Go through market providers listed above and look for market data for all assets.
-    // Once data is found, exit the loop and return result. If no data is found for any
-    // provider, throw an error.
+    if (args.fetchFromAllProviders) {
+      /** Some of the providers above have performance issues. Instead of modifying the default behavior,
+       * we'll just use a blacklist exclude the questionable providers with the fetchFromAllProviders option is enabled.
+       */
+      const blacklistedProviders = [
```
0xdef1cafe, line 65: lemme provide a little context and potentially simpler strategy here

iterating the `marketProviders` array in this `findAll` method never actually made any sense.

from the PR description i'm gathering the intent of this is to get market data for osmo tokens.

if that's the case, this *feels* like the responsibility of the defi resolvers in the front end, as this market data is only required for that domain of the app, and we could just call the `findByAssetId` method with each required asset id. if that's too noisy, or we agree we kind of always need this data, we could do this here with just a whitelist of coingecko + osmo data providers.

what do you reckon?

0xdef1cafe: after discussing - root cause/motivation of this PR was market data unable for long tail ibc tokens at runtime

IMO, the respective domain (osmo LPing) in the client should be responsible for fetching this lazily as required using the `findByAssetId` method, which can be called with an automatically generated RTK query hook (`useFindByAssetIdQuery`) or imperatively, like so https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#example-1

0xdef1cafe (approved):

PR #1176, opened 2023-01-23
feat(chain-adapters): return token balances from cosmossdk getaccount method

pastaghost:
* CosmosSdkBaseAdapter::getAccount() returns asset balance data
* this works for both cosmos and osmosis accounts and return data has been verified with the chain-adapter CLI.


packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -191,7 +192,12 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
           })),
         }))
 
-        return { ...data, delegations, redelegations, undelegations, rewards }
+        const assets = data.assets.map<CosmosSDKToken>((asset) => ({
+          amount: asset.amount,
+          denom: asset.denom,
```
gomesalexandre, line 197: Can we redeclare `generateAssetIdFromOsmosisDenom` in lib and make this `assetId` for parity with `EVMBaseAdapter`? Lib shouldn't speak token addresses / denoms but rather all CAIP

gomesalexandre (commented):
Looks good to me, added one comment WRT exposing the raw denom vs. the massaged `AssetId`. Web should now have all it needs to understand osmosis CAIPs, so we should expose the account tokens as such

packages/chain-adapters/src/cosmossdk/types.ts
```diff
@@ -89,7 +90,7 @@ export type FeeData = {
 }
 
 export type CosmosSDKToken = {
-  denom: string
+  assetId: AssetId | string
```
gomesalexandre, line 93:
```suggestion
  assetId: AssetId
```

gomesalexandre (commented):
Comment

0xdef1cafe (approved):

PR #1182, opened 2023-01-30
feat(swapper): add getSwappersWithQuoteMetadata method

0xApotheosis:
Add a new instance method, `getSwappersWithQuoteMetadata`, to the `SwapperManager` class.

This method returns all of the information we'll need to allow a user to pick from an ordered list of available swappers when consumed by `web`.

Removes `getBestSwapper`, so this is a breaking change PR.
Please do not clear the commit history when merging.


packages/swapper/src/manager/SwapperManager.ts
```diff
@@ -89,44 +91,29 @@ export class SwapperManager {
...
-      },
-      undefined,
-    )
-
-    const bestSwapper = bestQuoteTuple?.[0]
-    return bestSwapper
+    const settledSwapperDetailRequests: PromiseSettledResult<SwapperWithQuoteMetadata>[] =
+      await Promise.allSettled(
+        supportedSwappers.map(async (swapper) => {
+          const quote = await swapper.getTradeQuote(args)
```
gomesalexandre, line 97: As a safety net, do we want to catch errors here and `console.error()` instead of throwing? I.e no quote but still a valid swapper

gomesalexandre (commented):
Conceptually LGTM, will stamp after testing the consumption in a matching web PR

gomesalexandre (approved):
Tested with the matching web PR and confirmed this brings no regressions.

Managed to get best swappers / quotes for 0x and THOR trades, as well as proceed all the way to completed swaps.

packages/swapper/src/manager/SwapperManager.ts
0xApotheosis: As discussed out of band - our current assumption is that if we can't get a quote, the swapper is not valid for that pair/input args.

PR #1183, opened 2023-01-30
feat: parse osmosis lp transactions

pastaghost:
* adds parser support for osmosis `join_pool` and `exit_pool` messages

### DIFF
--- a/packages/swapper/CHANGELOG.md
+++ b/packages/swapper/CHANGELOG.md
@@ -1,3 +1,16 @@
+# [@shapeshiftoss/swapper-v16.0.0](https://github.com/shapeshift/lib/compare/@shapeshiftoss/swapper-v15.8.0...@shapeshiftoss/swapper-v16.0.0) (2023-01-30)
+
+
+### Features
+
+* **swapper:** add getSwappersWithQuoteMetadata method ([#1182](https://github.com/shapeshift/lib/issues/1182)) ([ee35823](https://github.com/shapeshift/lib/commit/ee358230a62b75788d3ff24f1923c511f166b708))
+
+
+### BREAKING CHANGES
+
+* **swapper:** removes getBestSwapper method
+BREAKING CHANGES: removes getBestSwapperMethod
+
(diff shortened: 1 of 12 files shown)


packages/swapper/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# [@shapeshiftoss/swapper-v16.0.0](https://github.com/shapeshift/lib/compare/@shapeshiftoss/swapper-v15.8.0...@shapeshiftoss/swapper-v16.0.0) (2023-01-30)
```
0xdef1cafe, line 1: no need to do this manually - semantic release generates this automatically from the commit messages into `main`

0xdef1cafe: ah this is just a rebase issue nvm

kaladinlight:
think this needs a rebase to get back in sync with main before review

packages/unchained-client/src/cosmossdk/parser/utils.ts
```diff
@@ -24,6 +25,7 @@ const assetIdByDenom = new Map<string, AssetId>([
 
 export const getAssetIdByDenom = (denom: string, assetId: string): AssetId | undefined => {
   if (assetIdByDenom.has(denom)) return assetIdByDenom.get(denom) as AssetId
+  if (assetId === osmosisAssetId) return adapters.osmosisDenomToAssetId(denom)
```
kaladinlight, line 28: this logic essentially lives right below this at L:32. could create the asset id there if we wanted to.

```
if (assetNamespace === 'gamm') {
  return toAssetId({ chainId, assetNamespace: 'ibc', assetReference: denom })
}
```

pastaghost: Yeah, at first I thought it might be nice to keep any osmosis-specific parsing stuff out of the cosmossdk base utils, but the doing it here saves several lines and a follow-up PR, so I agree.

kaladinlight (approved):

PR #1189, opened 2023-02-02
feat(caip): generate asset ids from osmosis lp token denoms

pastaghost:
* adds handler to generatAssetIdFromOsmosisDenom for LP tokens
* adds tests for generateAssetIdFromOsmosisDenom


gomesalexandre (approved):
LGTM

PR #1193, opened 2023-02-07
feat(market-service): safe URL construction with URL module

pastaghost:
* uses the URL module to safely construct osmosis market service URLs. This avoids potential errors with the base URL is passed in from web with/without a trailing slash.


0xdef1cafe (approved):
No need to wrap in IIFE but this is better anyway

PR #1201, opened 2023-02-17
feat(chain-adapters): add osmosis token send support

pastaghost:
* adds support for sending osmosis network tokens to OsmosisChainAdapter


gomesalexandre (approved):
Tested locally, LGTM. Web implementation isn't handling precision properly, but as far as the IBC denom logic is concerned, this does what it says on the box

PR #1202, opened 2023-02-22
fix: check memo destination matches chain

0xApotheosis:


0xApotheosis:
Closing, taking a slightly different approach.

PR #1206, opened 2023-02-22
feat: asset value

pastaghost:
* adds the asset-value package

### DIFF
--- a/packages/asset-value/package.json
+++ b/packages/asset-value/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "@shapeshiftoss/asset-value",
+  "version": "1.0.0",
+  "description": "Math and formatting utility class for asset data",
+  "homepage": "https://github.com/shapeshift/lib/tree/main/packages/asset-value",
+  "license": "MIT",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist/*.js",
+    "dist/*.d.ts",
+    "dist/service"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/pastaghost/asset-value"
+  },
+  "scripts": {
+    "build": "yarn clean && yarn compile",
+    "clean": "rm -rf dist && rm -rf tsconfig.build.tsbuildinfo",
+    "compile": "tsc -p tsconfig.build.json",
+    "dev": "tsc --watch",
+    "prepare": "yarn build",
+    "test": "jest test",
+    "type-check": "tsc --project ./tsconfig.build.json --noEmit"
+  },
+  "dependencies": {
+    "@shapeshiftoss/asset-service": "^8.8.1",
+    "@shapeshiftoss/caip": "^8.13.0",
+    "bignumber.js": "^9.1.1",
+    "ts-md5": "^1.3.1"
+  }
+}
--- a/packages/asset-value/src/assetValue.ts
+++ b/packages/asset-value/src/assetValue.ts
@@ -1,0 +1,212 @@
+import { AssetId } from '@shapeshiftoss/caip'
+import { BigNumber } from 'bignumber.js'
+import { Md5 } from 'ts-md5'
+
+import { CHECKSUM_LENGTH, DEFAULT_FORMAT_DECIMALS } from './constants'
+import { AssetValueParams, isAssetValueParams, SerializedAssetValue } from './types'
+import { AssetValueFormat } from './types'
+import { bn, bnOrZero } from './utils'
+
+type AssetValueState = {
+  assetId: AssetId
+  precision: number
+  value: string
+}
+
+//TODO(pastaghost): Add JSDoc documentation
+export class AssetValue {
+  state: AssetValueState
+
+  public constructor(params: AssetValueParams | SerializedAssetValue) {
+    if (isAssetValueParams(params)) {
+      if (params.asset && !(params.asset.assetId?.length && params.asset.precision > -1)) {
+        throw new Error('Cannot initialize AssetValue with invalid asset')
+      }
+
+      if (!params.asset && !(params.assetId?.length && params.precision > -1)) {
+        throw new Error('Cannot initialize AssetValue with invalid asset')
+      }
+
+      let _value: string
+      switch (params.format) {
+        case AssetValueFormat.BASE_UNIT:
+          _value = bnOrZero(params.value).toFixed(0, BigNumber.ROUND_DOWN)
+          break
+        case AssetValueFormat.PRECISION:
+          _value = bnOrZero(params.value)
+            .multipliedBy(bn(10).pow(params.asset ? params.asset.precision : params.precision))
+            .toString()
+          break
+      }
+      this.state = {
+        assetId: params.asset ? params.asset.assetId : params.assetId,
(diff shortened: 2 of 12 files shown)


packages/asset-value/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "dist/*.js",
+    "dist/*.d.ts",
+    "dist/service"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/pastaghost/asset-value"
```
kaladinlight, line 19: update with shapeshift repo

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+  precision: number
+  value: string
+}
+
+//TODO(pastaghost): Add JSDoc documentation
+export class AssetValue {
+  state: AssetValueState
+
+  public constructor(params: AssetValueParams | SerializedAssetValue) {
+    if (isAssetValueParams(params)) {
```
kaladinlight, line 21: wouldn't an `if (typeof params === 'string') { doSerializedStuff }` be more straight forward?

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+    const serializedState = JSON.stringify({
+      a: this.state.assetId,
+      p: this.state.precision,
+      v: this.state.value,
+    })
+    const hash = Md5.hashStr(serializedState).slice(0, CHECKSUM_LENGTH)
+    return `${serializedState}|${hash}` as SerializedAssetValue
+  }
+
+  public mult = this.multipliedBy.bind(this)
```
kaladinlight, line 174: to follow bignumberjs shorthand
```suggestion
  public times = this.multipliedBy.bind(this)
```

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+  public div = this.dividedBy.bind(this)
+  public gte = this.isGreaterThanOrEqualTo.bind(this)
```
kaladinlight, line 176: missing `gt`

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+
+type AssetValueState = {
+  assetId: AssetId
+  precision: number
+  value: string
+}
+
+//TODO(pastaghost): Add JSDoc documentation
+export class AssetValue {
+  state: AssetValueState
```
kaladinlight, line 18: reason for object over standard class vars? also should probably make private to avoid mutation outside of class

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+          _value = bnOrZero(params.value).toFixed(0, BigNumber.ROUND_DOWN)
+          break
+        case AssetValueFormat.PRECISION:
+          _value = bnOrZero(params.value)
+            .multipliedBy(bn(10).pow(params.asset ? params.asset.precision : params.precision))
+            .toString()
+          break
+      }
+      this.state = {
+        assetId: params.asset ? params.asset.assetId : params.assetId,
```
kaladinlight, line 42: shorthand option available
```suggestion
        assetId: params.assetId ?? params.asset.assetId,
```

packages/asset-value/src/types.ts
```diff
@@ -0,0 +1,38 @@
+import { Asset } from '@shapeshiftoss/asset-service'
+import { AssetId } from '@shapeshiftoss/caip'
+
+export enum AssetValueFormat {
+  BASE_UNIT = 1,
+  PRECISION = 2,
+}
+
+export type AssetValueParams = {
```
kaladinlight, line 9: personally would just add these types along side asset value class for easier readability in such a small package (could just reduce to single index.ts really imo)

packages/asset-value/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const CHECKSUM_LENGTH = 8
+export const DEFAULT_FORMAT_DECIMALS = 6
```
kaladinlight, line 2: could be useful to support configurable `DEFAULT_FORMAT_DECIMALS` possibly?

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+  public eq = this.isEqualTo.bind(this)
+  public base = this.toBaseUnit.bind(this)
+  public prec = this.toPrecision.bind(this)
+
+  private loadSerializedAssetValue(value: SerializedAssetValue) {
+    const [serializedState, hash] = value.split('|')
+    if (!(serializedState && hash)) {
+      throw new Error('Cannot initialize AssetValue from improperly-formatted SerializedAssetValue')
+    }
+    const parsedState = JSON.parse(serializedState)
```
kaladinlight, line 188: I don't recall the thrown error from `JSON.parse`, but it can throw if you want to throw a custom error

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!(parsedState && parsedState.a && parsedState.p && parsedState.v)) {
+      throw new Error('Cannot initialize AssetValue from underspecified SerializedAssetValue')
```
kaladinlight, line 190: could be extra nice and keep track of missing keys and provide as detail for the error

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  public toSerialized(): SerializedAssetValue {
+    const serializedState = JSON.stringify({
+      a: this.state.assetId,
+      p: this.state.precision,
+      v: this.state.value,
+    })
+    const hash = Md5.hashStr(serializedState).slice(0, CHECKSUM_LENGTH)
+    return `${serializedState}|${hash}` as SerializedAssetValue
```
kaladinlight, line 171: don't think as cast should be necessary

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+import { bn, bnOrZero } from './utils'
+
+type AssetValueState = {
+  assetId: AssetId
+  precision: number
+  value: string
+}
+
+//TODO(pastaghost): Add JSDoc documentation
+export class AssetValue {
```
kaladinlight, line 17: my only other thought is around the use of `bnOrZero` throughout in relation to being a big number wrapped asset. kind of makes it unpure in a sense, though it is for our use case and we use that for all of our bignumber math (I still have mixed feeling about this in general but it is what it is haha)

packages/asset-value/src/assetValue.ts
```diff
@@ -0,0 +1,212 @@
...
+    return new AssetValue({
+      value,
+      assetId: this.state.assetId,
+      precision: this.state.precision,
+      format: AssetValueFormat.BASE_UNIT,
+    })
+  }
+
+  public dividedBy(factor: string | number): AssetValue {
+    const value = bnOrZero(this.state.value).dividedBy(bnOrZero(factor)).toFixed()
```
kaladinlight, line 94: I have mentioned staying away from `bnOrZero` specifically for `dividedBy`. Result is `Infinity` vs `NaN` with `toFixed`. Honestly not sure how we handle either value in the client throughout though...

PR #1210, opened 2023-02-23
fix(swapper): osmosis multi-account swaps

pastaghost:
* fixes ATOM(account 0) -> OSMO(account 1) trades as reported by [TShifty](https://discord.com/channels/554694662431178782/1066088208905031681/1078117897060225124).
* note: there is probably a way to accomplish the same thing making better use of the type system than sticking an optional field onto the base trade types. I played around with this for a while, but ended up reverting to the naiive case for expediency. If you've got any suggestions, please let me know and I'll be happy to fix.

### DIFF
--- a/packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
+++ b/packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
@@ -163,6 +163,7 @@ export class OsmosisSwapper implements Swapper<ChainId> {
       sellAmountBeforeFeesCryptoBaseUnit: sellAmountCryptoBaseUnit,
       receiveAddress,
       accountNumber,
+      receiveAccountNumber,
     } = args
 
     if (!sellAmountCryptoBaseUnit) {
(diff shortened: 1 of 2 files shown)


packages/swapper/src/swappers/osmosis/OsmosisSwapper.ts
```diff
@@ -357,7 +365,7 @@ export class OsmosisSwapper implements Swapper<ChainId> {
     const cosmosAddress = isFromOsmo ? receiveAddress : sellAddress
     const signTxInput = await buildTradeTx({
       osmoAddress,
-      accountNumber,
+      accountNumber: receiveAccountNumber,
```
gomesalexandre, line 368: LGTM, this was super confusing to read haha:

https://github.com/shapeshift/lib/blob/d76db7a521277d7d2cb6950cfe86494edc7948f1/packages/swapper/src/swappers/osmosis/utils/helpers.ts#L297-L298

gomesalexandre (approved):
Not the biggest fan of `receiveAccountNumber` given we already have `receiveAddress` (should be one or the other, i.e we should be able to derive the receive address with the account number?), but won't nitpick this one, happy to merge in this state and maybe revisit later so we can unblock Osmo

PR #1213, opened 2023-02-24
fix(swapper): unrug osmo->atom cross-chain swaps

pastaghost:
* uses correct account number for swaps from osmosis to cosmoshub


gomesalexandre (approved):
Tested locally in https://github.com/shapeshift/web/pull/3901, confirmed this is the goat-er

PR #1214, opened 2023-02-24
fix(chain-adapters): field name in osmosis lp remove transaction

pastaghost:
* fixes incorrectly-named field in osmosis lp remove transaction


gomesalexandre (approved):
✅

PR #1215, opened 2023-02-24
chore: upgrade bignumber.js to match web

kaladinlight:
- due to mismatched versions, there were bignumber warning being thrown in web when locally linked


gomesalexandre (approved):
Tested locally with all packages linked, confirmed the warnings are now gone 🐐

PR #1216, opened 2023-02-24
chore: add identicon generation to avalanche and optimism

kaladinlight:
- `icon` field is required and in the case of missing icon, avalanche and optimism weren't generating identicons as placeholder

### DIFF
--- a/packages/asset-service/src/generateAssetData/avalanche/index.ts
+++ b/packages/asset-service/src/generateAssetData/avalanche/index.ts
@@ -1,19 +1,19 @@
 import { avalancheChainId } from '@shapeshiftoss/caip'
 
+import { Asset } from '../../service/AssetService'
 import { getRenderedIdenticonBase64 } from '../../service/GenerateAssetIcon'
 import { avax } from '../baseAssets'
 import * as coingecko from '../coingecko'
 
-export const getAssets = async () => {
+export const getAssets = async (): Promise<Asset[]> => {
   const assets = await coingecko.getAssets(avalancheChainId)
-  assets.push(avax)
-  return assets.map((asset) => {
-    if (!asset.icon) {
-      asset.icon = getRenderedIdenticonBase64(asset.assetId, asset.symbol, {
+  return [...assets, avax].map((asset) => ({
+    ...asset,
+    icon:
+      asset.icon ||
+      getRenderedIdenticonBase64(asset.assetId, asset.symbol, {
         identiconImage: { size: 128, background: [45, 55, 72, 255] },
         identiconText: { symbolScale: 7, enableShadow: true },
-      })
-    }
-    return asset
-  })
+      }),
+  }))
 }
--- a/packages/asset-service/src/generateAssetData/avalanche/index.ts
+++ b/packages/asset-service/src/generateAssetData/avalanche/index.ts
@@ -1,10 +1,19 @@
 import { avalancheChainId } from '@shapeshiftoss/caip'
 
+import { getRenderedIdenticonBase64 } from '../../service/GenerateAssetIcon'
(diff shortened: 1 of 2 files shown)


packages/asset-service/src/generateAssetData/avalanche/index.ts
```diff
@@ -1,10 +1,19 @@
...
 import * as coingecko from '../coingecko'
 
 export const getAssets = async () => {
   const assets = await coingecko.getAssets(avalancheChainId)
   assets.push(avax)
-  return assets
+  return assets.map((asset) => {
+    if (!asset.icon) {
+      asset.icon = getRenderedIdenticonBase64(asset.assetId, asset.symbol, {
+        identiconImage: { size: 128, background: [45, 55, 72, 255] },
```
kaladinlight, line 13: @reallybeard not sure if there is a specific identicon background color meant to be used for a specific chain? ethereum and osmosis seem to use the same background color, so leaving as this unless I hear otherwise

packages/asset-service/src/generateAssetData/avalanche/index.ts
```diff
@@ -1,10 +1,19 @@
...
+        identiconText: { symbolScale: 7, enableShadow: true },
+      })
+    }
+    return asset
+  })
```
0xApotheosis, line 18: I'd prefer a typed non-mutatey version of this:

```ts
export const getAssets = async (): Promise<Asset[]> => {
  const assets = await coingecko.getAssets(avalancheChainId)
  return [...assets, avax].map((asset) => ({
    ...asset,
    icon: asset.icon
      ? asset.icon
      : getRenderedIdenticonBase64(asset.assetId, asset.symbol, {
          identiconImage: { size: 128, background: [45, 55, 72, 255] },
          identiconText: { symbolScale: 7, enableShadow: true },
        }),
  }))
}
```

packages/asset-service/src/generateAssetData/optimism/index.ts
```diff
@@ -1,10 +1,19 @@
...
 
+import { getRenderedIdenticonBase64 } from '../../service/GenerateAssetIcon'
 import { optimism } from '../baseAssets'
 import * as coingecko from '../coingecko'
 
 export const getAssets = async () => {
   const assets = await coingecko.getAssets(optimismChainId)
   assets.push(optimism)
-  return assets
+  return assets.map((asset) => {
```
0xApotheosis, line 10: And this guy.

0xApotheosis (approved):
LGTM, I'm happy if @reallybeard is happy with Kevin's assumption.

gomesalexandre (dismissed):
LGTM! Pretty much what @0xApotheosis said about not mutating, and bearded man input, but realistically the former is a nitpick so:

🥜

packages/asset-service/src/generateAssetData/avalanche/index.ts
kaladinlight: https://github.com/shapeshift/lib/pull/1216/commits/405ed66ca234589240d65344fd4d05d0eeed6994

packages/asset-service/src/generateAssetData/optimism/index.ts
kaladinlight: https://github.com/shapeshift/lib/pull/1216/commits/405ed66ca234589240d65344fd4d05d0eeed6994

packages/asset-service/src/generateAssetData/avalanche/index.ts
reallybeard: We can just roll with what we have for the current assets. We have no specific colors for each chain. You can see the accounts page to see the reference to chain colors.

PR #1230, opened 2023-04-03
fix: cowswap buyAmountCryptoBaseUnit logic

0xApotheosis:
There is a logic error in the `getCowSwapTradeQuote` function. 

If we attempt to sell an amount that is less than the CoW Swap minimum, we get a quote using the minimum amount to ensure we still get a quote back.

The issue was that we are _always_ passing the buy asset amount that this minimum sell amount would buy (from the quote response) in the `buyAmountCryptoBaseUnit` field, regardless of if we made the minimum or not.

This is leading to inflated CoW Swap ratios which were rugging swapper selection UI, among a host of other bugs relating to minimum amounts.

To test this you'll need to be running https://github.com/shapeshift/web/pull/4177 on `web`, then:

1. Attempt to replicate the bug in https://github.com/shapeshift/web/issues/4168
2. With `REACT_APP_FEATURE_TRADE_RATES` enabled, we should no longer see negative amounts for CoW Swap, and it should be in it's correct place in the list (i.e. not "best" when it's worse than 0x) - trading a very small amount of FOX for ETH is a good way to get rates from CoW Swap, THORSwap and 0x and test this.

Closes https://github.com/shapeshift/web/issues/4168

Note for you @gomesalexandre, as I know you'll find it(!) - this does not yet fix the minimum amounts issue, which is separate, and fixed in https://github.com/shapeshift/web/pull/4178.

### DIFF
--- a/packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
+++ b/packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
@@ -65,12 +65,13 @@ export async function getCowSwapTradeQuote(
     const { minimum, maximum } = await getCowSwapMinMax(deps, sellAsset, buyAsset)
 
     const minQuoteSellAmount = bnOrZero(minimum).times(bn(10).exponentiatedBy(sellAsset.precision))
+    const isSellAmountBelowMinimum = bnOrZero(sellAmountBeforeFeesCryptoBaseUnit).lt(
+      minQuoteSellAmount,
+    )
 
     // making sure we do not have decimals for cowswap api (can happen at least from minQuoteSellAmount)
     const normalizedSellAmountCryptoBaseUnit = normalizeIntegerAmount(
-      bnOrZero(sellAmountBeforeFeesCryptoBaseUnit).lt(minQuoteSellAmount)
-        ? minQuoteSellAmount
-        : sellAmountBeforeFeesCryptoBaseUnit,
+      isSellAmountBelowMinimum ? minQuoteSellAmount : sellAmountBeforeFeesCryptoBaseUnit,
     )
 
     const apiInput: CowSwapSellQuoteApiInput = {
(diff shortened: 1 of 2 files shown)


packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
```diff
@@ -142,11 +143,18 @@ export async function getCowSwapTradeQuote(
...
+    // If isQuoteSellAmountBelowMinimum we don't want to replace it with normalizedSellAmount
     // The purpose of this was to get a quote from CowSwap even with small amounts
-    const quoteSellAmount = bnOrZero(sellAmountCryptoBaseUnit).lt(minQuoteSellAmount)
+    const quoteSellAmountCryptoBaseUnit = isQuoteSellAmountBelowMinimum
       ? sellAmountBeforeFeesCryptoBaseUnit
-      : normalizedSellAmount
+      : normalizedSellAmountCryptoBaseUnit
+
+    // Similarly, if isQuoteSellAmountBelowMinimum we can't use the buy amount from the quote
+    // because we aren't actually selling the minimum amount (we are attempting to sell an amount less than it)
```
gomesalexandre, line 154: 🧠

gomesalexandre (commented):
Tested in conjunction with https://github.com/shapeshift/web/pull/4177:

1. With an amount lower than the minimum, CoW swapper is in its correct position (i.e not best, since a trade with this amount would fail so the receive amount is effectively 0)

image

2. Miner fees are correctly displayed, see above

3. When selecting CoW, the receive amount (both in the input and min. expected after fees) shown doesn't account for the fee (i.e should be zero), though I believe this is a web concern and fixed in a follow-up PRs in the stack?

image
image

4. As noted in the description, minimum amounts are still incorrect, i.e 1400 FOX shows "Insufficient sell amount" currently but 1500 doesn't

image

According to CoWSwap's UI, the minimum should be ~126 FOX (though with very unfavourable receive amount where fees make up for ~87% of the sell amount, I believe with our heuristics the actual amount we deem to be minimum should be around 350 FOX e.g fees being ~40% of the sell amount)

image
image

If we confirm 3. isn't a matter of this PR, happy to get it and its matching web PR in

PR #1231, opened 2023-04-03
Fix rune deposits

BitHighlander:
use coinAsset as "THOR.RUNE" as defined in the thorchain swap protocol

### DIFF
--- a/packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
@@ -158,7 +158,7 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
       const msg: Message = {
         type: 'thorchain/MsgDeposit',
         value: {
-          coins: [{ asset: 'rune', amount: bnOrZero(value).toString() }],
+          coins: [{ asset: 'THOR.RUNE', amount: bnOrZero(value).toString() }],
           memo,
           signer: from,
         },


packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
```diff
@@ -158,7 +158,7 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
       const msg: Message = {
         type: 'thorchain/MsgDeposit',
         value: {
-          coins: [{ asset: 'rune', amount: bnOrZero(value).toString() }],
+          coins: [{ asset: 'THOR.RUNE', amount: bnOrZero(value).toString() }],
```
0xApotheosis, line 161: Does that mean trading into RUNE never worked for KeepKey?

The asset has been set to `rune` since Adam's original implementation in September 2022.

gomesalexandre: @BitHighlander pls ^^
Also relevant: https://discord.com/channels/554694662431178782/886041582422356018/1092531960200581240

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
```diff
@@ -154,11 +154,12 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
 
       const from = await this.getAddress({ accountNumber, wallet })
       const account = await this.getAccount(from)
-
+
```
gomesalexandre, line 157: This fails CI, please run `yarn local-ci` and `yarn lint:fix`

gomesalexandre (commented):
Regression tested ~~while linked in web~~ while manually applying this diff in the `chain-adapters` dist. This can't be tested by linking, nor installing built module as this check fails (this is an issue I've definitely seen previously):

https://github.com/shapeshift/lib/blob/62bf446e4451882df82ad02a86fdfe321b21acbe/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts#L161

### Develop

#### Swap to RUNE

image

#### Swap from RUNE

image

- Doesn't go past the "Thorchain Account" step on the KK

### This diff's `chain-adapters` "linked"

#### Swap to RUNE

image

#### Swap from RUNE

image

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
gomesalexandre: @0xApotheosis yeah looks like swaps *from* RUNE never worked in the first place and we never spotted it, KK has stronger validation on the `thorchain/MsgDeposit` MsgType:

https://github.com/shapeshift/hdwallet/commit/91d8d40a0c53d26db90189b2ab28e1b1a5d4cfb3#diff-d27caa3e5179cbadb8b52b36dd988e4cc50f0affa4b65e8f6af23a2dd6eae88dR77

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
gomesalexandre, line 157:
```suggestion
```

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
```diff
@@ -154,11 +154,10 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
 
       const from = await this.getAddress({ accountNumber, wallet })
       const account = await this.getAccount(from)
-
       const msg: Message = {
```
gomesalexandre, line 157:
```suggestion
      // https://dev.thorchain.org/thorchain-dev/concepts/memos#asset-notation
      const msg: Message = {
```

PR #1234, opened 2023-04-09
fix: cowswap minimum flag

0xApotheosis:
Tip for reviewers: the only actual logic change in this PR is in https://github.com/shapeshift/lib/commit/65beae0c4ad9fff1cd2c94b50fbf5f276e27b289, the reset is just renaming.

Existing logic incorrectly assumes that the `sellAmountCryptoBaseUnit` is the amount not including fees.

In reality, CoW Swap subtracts any trading fees from the `sellAmountCryptoBaseUnit` value in the returned response, so we need to add them back on to get the true "sell amount" of the sell asset.

This is causing the UI to show that the "sell amount is lower than the fee" when the sell amount is above the minimum by less than the fee amount.

To test, attempt to use CoW swap to trade, say, $21 - this should now work, whereas previously is would not until you sold $20 plus the fee amount.

Also refactors the naming of `minimum` and `maximum` to show their actual units.

### DIFF
--- a/packages/swapper/src/api.ts
+++ b/packages/swapper/src/api.ts
@@ -180,8 +180,8 @@ export type SwapSource = {
 }
 
 export interface MinMaxOutput {
-  minimum: string
-  maximum: string
+  minimumAmountCryptoHuman: string
+  maximumAmountCryptoHuman: string
 }
 
 export type ApprovalNeededOutput = {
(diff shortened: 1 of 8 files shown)


0xApotheosis:
Will fix tests, then open.

0xApotheosis:
For expediency of fixing a prod issue I'm going to open a separate PR that excludes the name fixes.

I'll do this once the monorail service is up and running 🚄

0xApotheosis:
Replaced by https://github.com/shapeshift/lib/pull/1235.

PR #1235, opened 2023-04-10
fix: isQuoteSellAmountBelowMinimum

0xApotheosis:
Existing logic incorrectly assumes that the `sellAmountCryptoBaseUnit` is the amount not including fees.

In reality, CoW Swap subtracts any trading fees from the `sellAmountCryptoBaseUnit` value in the returned response, so we need to add them back on to get the true "sell amount" of the sell asset.

This is causing the UI to show that the "sell amount is lower than the fee" when the sell amount is above the minimum by less than the fee amount.

To test, attempt to use CoW swap to trade, say, $21 - this should now work, whereas previously is would not until you sold $20 plus the fee amount.

### DIFF
--- a/packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
+++ b/packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
@@ -113,6 +113,10 @@ export async function getCowSwapTradeQuote(
       },
     } = quoteResponse
 
+    const quoteSellAmountPlusFeesCryptoBaseUnit = bnOrZero(sellAmountCryptoBaseUnit).plus(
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Tested against develop web with current lib version, both with the `TradeRates` flag on and off

### With `TradeRates` on and manual CoW selection:

#### This lib diff

image
image

#### `develop`'s lib, note the "Insufficient sell amount" which is indeed incorrect

image
image

### With `TradeRates` off and monkey patched `ZrxSwapperEthereum` off in web

#### This lib diff

image
image

#### `develop`s lib

image
image

I wasn't able to reproduce the sad behavior with the `TradeRates` flag off, though turning the `TradeRates` flag on confirmed adding back the CoW fees (i.e the fees taken from the sell asset and effectively occurring on the buy asset) for the minimum is now correct.

